PR #8, opened 2019-11-27
Add feature deploy

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

I think the `deploy` subcommand can be optional, because developers can send a RPC request to deploy the contract by using other tools.


koushiro:
I have updated the change about the `deploy` feature in README, PTAL again :)

PR #17, opened 2020-01-08
Save built artifact, it will be available in gitlab job

TriplEight:
I will play with this branch for a while


TriplEight:
@ascjones ready to be merged, CI fail is not related
if you don't mind I took `--release` away from `test` job because AFAIU it doesn't really make sense. This combination (`cargo test` and `--release`) does not check your "release" state of the code, instead it results in `bench` cargo profile which is just slower than `test` one. Feel free to return it, but tell me why then, I want to know.

PR #29, opened 2020-02-04
Add wasm-opt optimization build step

### LINKED ISSUE #23: Add Binaryen's `wasm-opt` to `cargo contract build`
After merging https://github.com/paritytech/ink/pull/312 ink! smart contracts will suffer from bloated Wasm binaries due to missing optimizations (or bugs?) between Rust and Wasm translation.

These cause a `0x00` 16kB buffer to not be removed from the Wasm binary.
However, Binaryen's `wasm-opt` with `-O3` (or even `-O2`) can easily do this among other neat optimization routines.

Some contracts have been shown to have a reduced Wasm binary size from 27kB down to 9kB just through using `wasm-opt`.

The `cargo contract build` process should include `wasm-opt` by default, however, if `wasm-opt` does not exist on the user machine it should simply avoid using it (as it is been done today) and print the user a warning that the resulting Wasm binary might not be optimal and refer the user to install the `wasm-opt` tool.

ascjones:
Closes #23.

- Requires `wasm-opt` to be installed on user machine
- If not present, then the build still "succeeds" but with a message encouraging the user to install the tool.
- Displays the size of the file before and after optimization
- Adds some pretty colours

image

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -23,6 +23,7 @@ use std::{
 use anyhow::{Context, Result};
 use cargo_metadata::MetadataCommand;
 use parity_wasm::elements::{External, MemoryType, Module, Section};
+use colored::Colorize;
 
 /// This is the maximum number of pages available for a contract to allocate.
 const MAX_MEMORY_PAGES: u32 = 16;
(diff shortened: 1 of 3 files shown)


src/cmd/build.rs
```diff
@@ -153,20 +161,87 @@ fn post_process_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
     Ok(())
 }
 
+/// Attempts to perform optional wasm optimization using `wasm-opt`.
+///
+/// The intention is to reduce the size of bloated wasm binaries as a result of missing
+/// optimizations (or bugs?) between Rust and wasm.
```
Robbepop:
```suggestion
/// optimizations (or bugs?) between Rust and Wasm.
```

src/cmd/build.rs
```diff
@@ -153,20 +161,87 @@ fn post_process_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
...
+///
+/// This step depends on the `wasm-opt` tool being installed. If it is not the build will still
+/// succeed, and the user will be encouraged to install it for further optimizations.
+fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
+    // check `wasm-opt` installed
+    if which::which("wasm-opt").is_err() {
+        println!(
+            "{}",
+            "wasm-opt is not installed. Install this tool on your system in order to \n\
+             reduce the size of your contract's wasm binary. \n\
```
Robbepop:
```suggestion
             reduce the size of your contract's Wasm binary. \n\
```

PR #33, opened 2020-02-06
Reduce wasm binary size with `cargo-xbuild` & removing `rlib` crate-type

### LINKED ISSUE #32: Use Xargo for building Wasm contracts
This is a contender issue for: https://github.com/paritytech/cargo-contract/issues/22

It solves the same problems but with a different tool. For more details visit the linked issue.
Instead of invoking Cargo in order to build ink! smart contracts we should instead build them using Xargo (https://crates.io/crates/xargo).

This requires another `Xargo.toml` in the contract's root directory with the following contents:

```toml
[target.wasm32-unknown-unknown.dependencies]
core = {default-features=false, features=["panic_immediate_abort"]}
std = {default-features=false, features=["panic_immediate_abort"]}
alloc = {}
```

Also we have to remove the contents of the `.cargo/config` file to not override Xargo settings: (We demonstrate this here by simply commenting the line out.)

```
[target.wasm32-unknown-unknown]
rustflags = [
#	"-C", "link-args=-z stack-size=65536 --import-memory"
]
```

Another obstacle is that currently Cargo (or `rustc`) overrides `profile` settings given in `Cargo.toml` when `rlib` is specified as in:

```
[lib]
name = "erc20"
crate-type = [
	# Used for normal contract Wasm blobs.
	"cdylib",
	# Used for ABI generation.
	"rlib",
]
```

Unfortunately in ink! we use `rlib` for generating ink! metadata.
For automating this process we have to somehow get rid of this `rlib` field during compilation in order to NOT to override the following `profile` section in `Cargo.toml`:

```
[profile.release]
panic = "abort"
lto = true
opt-level = "z"
codegen-units = 1
```

All these settings are very important to get all the important optimizations in the final Wasm binary.

After taking all these steps building a highly optimized ink! Wasm smart contract is just:

```
xargo build --no-default-features --target wasm32-unknown-unknown --release
```

So simply exchange `cargo` with `xargo`.
Also this still requires a pass by `wasm-opt` afterwards so this pass should be run before applying `wasm-opt` for final optimizations.

ascjones:
Closes #32.

This PR significantly reduces the final wasm binary size by custom building rust `libcore` with the `panic_immediate_abort` feature which [prevents panic strings and formatting code being included in the final executable](https://github.com/johnthagen/min-sized-rust#remove-panic-string-formatting-with-panic_immediate_abort).

This is achieved using [`cargo-xbuild`](https://github.com/rust-osdev/cargo-xbuild).

## `rlib` crate-type

`rlib` is required to generate the metadata and also for `ink-as-dependency`. However it also causes bloat in the final Wasm blob so it will be automatically removed if enabled when running `cargo contract build`. 

Likewise if the `Cargo.toml` doesn't have the `rlib` in `crate-types` and the user runs `generate-metadata`, then that will also be added temporarily in order for build to succeed.

A copy of the users `Cargo.toml` is made, amended as necessary and written to a temporary directory. Relative paths are rewritten to absolute paths.

## todo
- [x] For generate-metadata modify Cargo.toml to add "rlib" crate-type
- [x] ~~User option to restore original Cargo.toml if process terminated~~ (superseded)
- [x] Use tmp dir for amended manifests
- [x] Convert to `xbuild`
- [x] Test all happy and unhappy paths
- [x] Bug with building after build then `cargo clean` (file not found)
- [x] Figure out discrepancy of reported opt file size vs actual
- [x] Test all ink! examples work
- [x] Fix Delegator build
- [x] Get `generate-metadata` to work
- [x] Remove `.cargo/config` linkargs
~~Add `[profile.release]` section, use provided values by default?~~ See #40 
- [x] Add `--verbose` and `--quiet` args, `quiet` by default


Robbepop:
Very nice! All in all the code looks clean.
Worked for Flipper and ERC20 but not so well for Delegator.
(File not found error.)

ascjones:
Using `xargo` doesn't work with `ink/examples/delegator`, there is a [known issue](https://github.com/japaric/xargo/issues/189) which causes the top level `delegator` contract to fail with "conflicting versions" of the custom build `libcore`.  

Since `xargo` is in maintenance mode I am switching over to `cargo-xbuild`, which just today added the feature to [enable `panic_immediate_abort`](https://github.com/rust-osdev/cargo-xbuild/pull/56) which was previously missing.

I've also opened a PR to make it work with workspace configs (as our contract configs are): https://github.com/rust-osdev/cargo-xbuild/pull/57

README.md
```diff
@@ -35,6 +35,19 @@ SUBCOMMANDS:
     help                 Prints this message or the help of the given subcommand(s)
 ```
 
+## Contract build config
+
+Building the contract uses `cargo-xbuild` under the hood for optimum Wasm binary size. This requires the following
+configuration section to be added to your contract's `Cargo.toml`:
```
Robbepop: Cannot this be automated with the temporary `Cargo.toml` file?

ascjones: Yes, but it should be there permanently in the config.

Robbepop:
I have had one bug with the temporary `Cargo.toml`. My `cargo contract build` somehow failed (for a random reason) but left the temporary `Cargo.toml` and the original `Cargo.toml` in their intermediary state. I had to manually remove the temporary and rename the original in order to make everything work again. This should never happen. Also we should try to make this temporary time frame as short as possible.

ascjones:
Yes that's the thing I'm least happy about in this PR. If you run it again it will clean up after itself.

Can you think of a better alternative? Perhaps copy to a temp dir, modify the `Cargo.toml` there set the target dir to our local target dir?

Robbepop:
> Yes that's the thing I'm least happy about in this PR. If you run it again it will clean up after itself.
> 
> Can you think of a better alternative? Perhaps copy to a temp dir, modify the `Cargo.toml` there set the target dir to our local target dir?

I am not sure. In theory I would try to minimize copy overhead and also try to not invalidate the Cargo build cache everytime I build my contract. How do other (similar) build helpers (programs) handle this case?

ascjones:
> In theory I would try to minimize copy overhead and also try to not invalidate the Cargo build cache everytime I build my contract

These were similarly my concerns with this approach, which is why I opted for modifying inline.

> How do other (similar) build helpers (programs) handle this case?

I'm not sure - I'll have a look and find out.

src/cmd/build.rs
```diff
@@ -72,27 +79,69 @@ pub fn collect_crate_metadata(working_dir: Option<&PathBuf>) -> Result<CrateMeta
...
+/// Uses [`cargo-xbuild`](https://github.com/rust-osdev/cargo-xbuild) for maximum optimization of
+/// the resulting Wasm binary.
+fn build_cargo_project(crate_metadata: &CrateMetadata, verbosity: Option<Verbosity>) -> Result<()> {
+    util::assert_channel()?;
+
+    // set RUSTFLAGS, read from environment var by cargo-xbuild
+    std::env::set_var(
+        "RUSTFLAGS",
+        "-C link-arg=-z -C link-arg=stack-size=65536 -C link-arg=--import-memory",
+    );
```
Robbepop: So this is why we no longer need `.cargo/config` file, right?

src/cmd/build.rs
```diff
@@ -72,27 +79,69 @@ pub fn collect_crate_metadata(working_dir: Option<&PathBuf>) -> Result<CrateMeta
...
+
+    let verbosity = verbosity.map(|v| match v {
+        Verbosity::Verbose => xargo_lib::Verbosity::Verbose,
+        Verbosity::Quiet => xargo_lib::Verbosity::Quiet,
+    });
```
Robbepop: Nice, so `cargo contract build --verbose` works now! :)

src/cmd/build.rs
```diff
@@ -72,27 +79,69 @@ pub fn collect_crate_metadata(working_dir: Option<&PathBuf>) -> Result<CrateMeta
...
+
+    let xbuild = |manifest_path: &ManifestPath| {
```
Robbepop: So `xbuild` is always used and not only when it is installed as it was before. But I think that this is actually the better decision since it is now used as library. So, good! :)

src/cmd/metadata.rs
```diff
@@ -14,57 +14,89 @@
...
+        let target_dir_arg = format!("--target-dir={}", target_dir.to_string_lossy());
+        util::invoke_cargo(
+            "run",
+            &[
+                "--package",
+                "abi-gen",
+                &tmp_manifest_path.cargo_arg(),
+                &target_dir_arg,
+                "--release",
+                // "--no-default-features", // Breaks builds for MacOS (linker errors), we should investigate this issue asap!
```
Robbepop: I think `--no-default-features` is wrong anyway because this tool performs IO, so for example cannot be `no_std`.

src/workspace.rs
```diff
@@ -0,0 +1,393 @@
...
+        format!("--manifest-path={}", self.path.to_string_lossy())
+    }
+
+    /// The directory path of the manifest path.
+    ///
+    /// Returns `None` if the path is just the plain file name `Cargo.toml`
+    pub fn directory(&self) -> Option<&Path> {
+        let just_a_file_name =
+            self.path.iter().collect::<Vec<_>>() == vec![Path::new(MANIFEST_FILE)];
+        if !just_a_file_name {
```
Robbepop: lol

src/workspace.rs
```diff
@@ -0,0 +1,393 @@
...
+    ///
+    /// Enables the use of a temporary amended copy of the manifest.
+    ///
+    /// # Rewrites
+    ///
+    /// - `[lib]/path`
+    /// - `[dependencies]`
+    ///
+    /// Dependencies with package names specified in `exclude_deps` will not be rewritten.
+    fn rewrite_relative_paths<I, S>(&mut self, exclude_deps: I) -> Result<&mut Self>
```
Robbepop: Wow this was way more work than I expected. O.O

src/cmd/build.rs
ascjones: Precisely!

src/workspace.rs
ascjones: Me too...

PR #46, opened 2020-03-09
Fix saving artifact

TriplEight:


TriplEight:
@ascjones you can find build artifacts [here](https://gitlab.parity.io/parity/cargo-contract/-/jobs/411273/artifacts/browse/artifacts/) please let me know if you need to upload binaries to more persistent storage or even release them.

ascjones:
Awesome thanks @TriplEight. 

I have started doing releases, though currently we assume users will install from source. It might be good to get the binaries too for release - I will let you know.

TriplEight:
@ascjones thing is last week we've bee discussing "pallet-contracts-waterfall" CI and it wants to use the latest "cargo-contract" version. Which it currently does, by installing from the source. But also guys wanted to have an ability to roll back and regress test with the previous versions. So maybe if it's early to release it to cratesio, you can think of storing binaries on s3, or in github releases.

PR #47, opened 2020-03-18
Update template to ink! 2.0 dependencies

Robbepop:
Follow-up for https://github.com/paritytech/ink/pull/353.

### DIFF
--- a/template/Cargo.toml
+++ b/template/Cargo.toml
@@ -5,13 +5,19 @@ authors = ["[your_name] <[your_email]>"]
 edition = "2018"
 
 [dependencies]
-ink_abi = { git = "https://github.com/paritytech/ink", package = "ink_abi", default-features = false, features = ["derive"], optional = true }
-ink_primitives = { git = "https://github.com/paritytech/ink", package = "ink_primitives", default-features = false }
-ink_core = { git = "https://github.com/paritytech/ink", package = "ink_core", default-features = false }
-ink_lang = { git = "https://github.com/paritytech/ink", package = "ink_lang", default-features = false }
+ink_abi = { version = "2.0.0", git = "https://github.com/paritytech/ink", tag = "latest-v2.0", package = "ink_abi", default-features = false, features = ["derive"], optional = true }
+ink_primitives = { version = "2.0.0", git = "https://github.com/paritytech/ink", tag = "latest-v2.0", package = "ink_primitives", default-features = false }
+ink_core = { version = "2.0.0", git = "https://github.com/paritytech/ink", tag = "latest-v2.0", package = "ink_core", default-features = false }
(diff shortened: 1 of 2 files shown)


template/Cargo.toml
```diff
@@ -5,13 +5,19 @@ authors = ["[your_name] <[your_email]>"]
 edition = "2018"
 
 [dependencies]
-ink_abi = { git = "https://github.com/paritytech/ink", package = "ink_abi", default-features = false, features = ["derive"], optional = true }
-ink_primitives = { git = "https://github.com/paritytech/ink", package = "ink_primitives", default-features = false }
-ink_core = { git = "https://github.com/paritytech/ink", package = "ink_core", default-features = false }
-ink_lang = { git = "https://github.com/paritytech/ink", package = "ink_lang", default-features = false }
+ink_abi = { version = "2.0.0", git = "https://github.com/paritytech/ink", tag = "latest-v2.0", package = "ink_abi", default-features = false, features = ["derive"], optional = true }
+ink_primitives = { version = "2.0.0", git = "https://github.com/paritytech/ink", tag = "latest-v2.0", package = "ink_primitives", default-features = false }
+ink_core = { version = "2.0.0", git = "https://github.com/paritytech/ink", tag = "latest-v2.0", package = "ink_core", default-features = false }
```
athei: Shouldn't it be `branch=latest-v2.0`?. I would prefer 2 without the .0 though. Even both is valid in semver I think it is more obvious that we are incrementing minors.

Robbepop: The `latest-v2.0` is a moving that that will always move to the latest `v2.0` version. Once we are at `v2.1` there will be two tags for those purposes. But tbh I hope that we will make it to crates.io before that happens. So this can be seen as more of a hack towards the goal of just being at crates.io.

athei: This will be equivalent to putting semver `2.0.0` instead of `2.0` there. Is that really what you want?

template/Cargo.toml
```diff
@@ -5,13 +5,19 @@ authors = ["[your_name] <[your_email]>"]
 edition = "2018"
 
 [dependencies]
-ink_abi = { git = "https://github.com/paritytech/ink", package = "ink_abi", default-features = false, features = ["derive"], optional = true }
-ink_primitives = { git = "https://github.com/paritytech/ink", package = "ink_primitives", default-features = false }
-ink_core = { git = "https://github.com/paritytech/ink", package = "ink_core", default-features = false }
-ink_lang = { git = "https://github.com/paritytech/ink", package = "ink_lang", default-features = false }
+ink_abi = { version = "2.0.0", git = "https://github.com/paritytech/ink", tag = "latest-v2", package = "ink_abi", default-features = false, features = ["derive"], optional = true }
```
athei: Version should be just "2" then. Otherwise it does not match the expected semver of the moving tag.

PR #51, opened 2020-03-27
Add option to build with unmodified original manifest

ascjones:
By default a contract's `Cargo.toml` will be amended to optimize the final Wasm binary. At the moment that means just removing the `rlib` crate type if present, though in the future we may make further modifications.

This PR provides the option `-Z original-manifest` (feedback on the name appreciated) to disable all modifications to the `Cargo.toml`: it will just build using the original file. Note that this is introduced as an `unstable-option` (short `-Z`)

This will result in a larger final contract binary size if `rlib` is enabled, but gives full control to the user for options from the `Cargo.toml`. This is useful for troubleshooting e.g https://github.com/paritytech/ink/issues/364.

## Options

In general we should try to minimize the number of options/flags we support in order to keep it as simple as possible, so this and other options should be considered through that lens.

**Update**: with this in mind I have added this as an `unstable-option` (short `-Z`),


ascjones:
Suggestion from @Robbepop: uglify this option to discourage use `e.g. put debug- before command`

ascjones:
@Robbepop I've made this an unstable option `-Z`, and also print a warning during build

src/main.rs
```diff
@@ -112,6 +112,37 @@ impl TryFrom<&VerbosityFlags> for Option<Verbosity> {
     }
 }
 
+#[derive(Debug, StructOpt)]
+struct UnstableOptions {
+    /// Use the original manifest (Cargo.toml), do not modify for build optimizations
+    #[structopt(long = "unstable-options", short = "Z", number_of_values = 1)]
+    options: Vec<String>,
+}
```
Robbepop, line 120: How would this look like if we specify multiple unstable options? Each having their own `-Z` or one `-Z` for all? I guess we should mimic `rustc` here that requires one `-Z` per unstable feature usage.

ascjones: Indeed that should work i.e. 1 `-Z` per option, I took this pattern from https://github.com/TeXitoi/structopt/blob/master/examples/keyvalue.rs#L22

Issue #40, opened 2020-02-25
Add default `[profile.release]` values to Cargo.toml if not present


ascjones:
As a follow up to #33, which enables amending the `Cargo.toml`, we could add the following section if it is not present:

```
[profile.release]
panic = "abort"
lto = true
opt-level = "z"
overflow-checks = true
```

These are the default optimal settings for contract binary size. However if any of these settings are already defined in the Cargo.toml, then those values should be used instead of these defaults. Potentially the user could be warned if using a non-optimal setting.

Robbepop:
Note that
- `overflow-checks = true` add overhead to the final Wasm binary size but is something like safe-math in Ethereum. However, it is safe-math overtuned because it applies safe-math everywhere instead of locally in user provided code. So more overhead than needed which is sub optimal.
- `codegen-units = 1` could be useful since it greatly further optimized the final Wasm, however, takes a lot more time to actually compile.
- `lto = true` is similar to `codegen-units = 1` in that it further optimizes at the cost of compilation.
- `panic = abort` is always needed.
- `opt-level = "z"` has shown in the past to be inferior to `opt-level = 3` when it comes to resulting Wasm file size. Has to be benchmarked and tested.

So we might end up with 2 main configurations:
1. `debug` configuration that tries to build quickly
2. `release` configuration that tries to apply most optimizations

ascjones:
Closed by #55

PR #55, opened 2020-06-15
Add preferred default release profile settings

ascjones:
Rel #40, https://github.com/paritytech/ink/pull/448

If any of the following `[profile.release]` settings are missing, then they are added with the following preferred defaults:

```
[profile.release]
panic = "abort"
lto = true
opt-level = "z"
overflow-checks = true
codegen-units = 1
```
Any of the above that are manually specified are **not** overwritten. This allows manual control over e.g. `opt-level` where in some cases `opt-level = 3` might result in a smaller binary size.  

**Template changes:**
  - Remove `[profile.release]` section
  - ~~Use `scale-info` from crates.io instead of `type-metadata`~~ (reverted: requires ink! 3 release)

### DIFF
--- a/src/workspace.rs
+++ b/src/workspace.rs
@@ -432,7 +432,7 @@ impl Workspace {
 
 /// Subset of cargo profile settings to configure defaults for building contracts
 pub struct Profile {
-    opt_level: OptLevel,
+    opt_level: Option<OptLevel>,
     lto: Lto,
     // `None` means use rustc default.
     codegen_units: Option<u32>,
@@ -444,7 +444,7 @@ impl Profile {
     /// The preferred set of defaults for compiling a release build of a contract
     pub fn default_contract_release() -> Profile {
         Profile {
-            opt_level: OptLevel::Z,
+            opt_level: Some(OptLevel::Z),
             lto: Lto::Bool(true),
             codegen_units: Some(1),
             overflow_checks: true,
@@ -463,7 +463,9 @@ impl Profile {
                 profile.insert(key.into(), value);
             }
         };
-        set_value_if_vacant("opt-level", self.opt_level.to_toml_value());
+        if let Some(opt_level) = self.opt_level {
+            set_value_if_vacant("opt-level", opt_level.to_toml_value());
+        }
         set_value_if_vacant("lto", self.lto.to_toml_value());
         if let Some(codegen_units) = self.codegen_units {
             set_value_if_vacant("codegen-units", codegen_units.into());
@@ -474,13 +476,12 @@ impl Profile {
 }
 
 /// The [`opt-level`](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level) setting
-#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash, PartialOrd, Ord)]
 #[allow(unused)]
+#[derive(Clone, Copy)]
 pub enum OptLevel {
-    Zero,
-    One,
-    Two,
-    Three,
+    O1,
+    O2,
+    O3,
     S,
     Z,
 }
@@ -488,10 +489,9 @@ pub enum OptLevel {
 impl OptLevel {
     fn to_toml_value(&self) -> value::Value {
         match self {
-            OptLevel::Zero => 0.into(),
-            OptLevel::One => 1.into(),
-            OptLevel::Two => 2.into(),
-            OptLevel::Three => 3.into(),
+            OptLevel::O1 => 1.into(),
+            OptLevel::O2 => 2.into(),
+            OptLevel::O3 => 3.into(),
             OptLevel::S => "s".into(),
             OptLevel::Z => "z".into(),
         }
--- a/src/workspace.rs
+++ b/src/workspace.rs
@@ -432,7 +432,7 @@ impl Workspace {
 
 /// Subset of cargo profile settings to configure defaults for building contracts
 pub struct Profile {
-    opt_level: Option<OptLevel>,
+    opt_level: OptLevel,
     lto: Lto,
     // `None` means use rustc default.
     codegen_units: Option<u32>,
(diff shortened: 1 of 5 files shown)


src/workspace.rs
```diff
@@ -137,6 +137,28 @@ impl Manifest {
 
     /// Set `[profile.release]` lto flag
     pub fn with_profile_release_lto(&mut self, enabled: bool) -> Result<&mut Self> {
+        let lto = self
+            .get_profile_release_table_mut()?
+            .entry("lto")
+            .or_insert(enabled.into());
+        *lto = enabled.into();
+        Ok(self)
+    }
```
Robbepop, line 146: Isn't this superseeded by `with_profile_release_defaults`?

src/workspace.rs
```diff
@@ -411,3 +429,170 @@ impl Workspace {
...
+}
+
+impl Default for Profile {
+    fn default() -> Profile {
+        Profile {
+            opt_level: OptLevel::One,
+            lto: Lto::Bool(false),
+            codegen_units: None,
+            overflow_checks: false,
+            panic: PanicStrategy::Unwind,
```
Robbepop, line 450: Not sure in whichever context unwinding is applicable to Wasm smart contracts. But I still like it being here as reference. Obviously those defaults are not applicable anyways, right?

src/workspace.rs
```diff
@@ -411,3 +429,170 @@ impl Workspace {
...
+        set_value_if_vacant("overflow-checks", self.overflow_checks.into());
+        set_value_if_vacant("panic", self.panic.to_toml_value());
+    }
+}
+
+/// The [`opt-level`](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level) setting
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash, PartialOrd, Ord)]
+#[allow(unused)]
+pub enum OptLevel {
+    Zero,
```
Robbepop, line 492: rename to `None` or `NoOptimizations`?

src/workspace.rs
```diff
@@ -411,3 +429,170 @@ impl Workspace {
...
+    One,
+    Two,
+    Three,
```
Robbepop, line 495: Rename to `O1`, `O2` and `O3`?

src/workspace.rs
Robbepop: We could also wrap `OptLevel` into an `Option` where `None` is simply that no optimizations are performed.

src/workspace.rs
```diff
@@ -411,3 +429,170 @@ impl Workspace {
...
+/// The [`link-time-optimization`](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) setting.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash, PartialOrd, Ord)]
+pub enum Lto {
+    /// False = no LTO
+    /// True = "Fat" LTO
+    Bool(bool),
+    /// Named LTO settings like "thin".
+    #[allow(unused)]
+    Named(&'static str),
+}
```
Robbepop, line 522: As with `OptLevel`, either we pack both in `Option` to simulate the `None` or `Off` case or we add an additional variant both but not different tactics for both.

src/workspace.rs
```diff
@@ -411,3 +429,170 @@ impl Workspace {
...
+        }
+    }
+}
+
+/// The [`link-time-optimization`](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) setting.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash, PartialOrd, Ord)]
+pub enum Lto {
+    /// False = no LTO
+    /// True = "Fat" LTO
+    Bool(bool),
```
Robbepop, line 518: Would it be much worse to have `FatLto`, `ThinLto` and `Off` as variants? Kind of dislike the `Bool` here.

src/workspace.rs
```diff
@@ -411,3 +429,170 @@ impl Workspace {
...
+            Lto::Bool(b) => value::Value::Boolean(*b),
+            Lto::Named(n) => value::Value::String(n.to_string()),
+        }
+    }
+}
+
+/// The `panic` setting.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash, PartialOrd, Ord)]
+pub enum PanicStrategy {
+    Unwind,
```
Robbepop, line 536: Not sure how this is every going to be applicable for Wasm smart contracts to be honest.

Robbepop: Or is it?

src/workspace.rs
```diff
@@ -411,3 +429,170 @@ impl Workspace {
...
+    #[test]
+    fn merge_profile_preserves_user_defined_settings() {
+        let profile = Profile::default_contract_release();
+
+        let manifest_toml = r#"
+panic = "unwind"
+lto = false
+opt-level = 3
+overflow-checks = false
+codegen-units = 256
```
Robbepop, line 583: Would indentation kill this?

src/workspace.rs
ascjones: No because this will overwrite the user-defined `lto` value if present. This is required for `generate-metadata` when user has `[profile.release] lto = true`.

src/workspace.rs
ascjones: I copied the `Profile` struct originally from [`cargo`](https://github.com/rust-lang/cargo/blob/b6a4b074ed924878fefb631c7d559ca47331e5eb/src/cargo/core/profiles.rs#L584), and brought the `Default` impl with it. It might be useful for reference but I think I will remove it since it is unused.

As far as having `Unwind` as a possible value of `PanicStrategy`, although it is not applicable I think it is useful to keep it since it is a valid value for that field.

Robbepop: Maybe emit a warning if a user tries to compile an ink! contract for the Wasm target (`target=wasm32-unknown-unknown`) and with `panic=unwind`.

src/workspace.rs
ascjones: See above, gonna keep it for reference because it is a valid value for `PanicStrategy`, even though unused/not applicable for the wasm contract build itself

src/workspace.rs
ascjones: Copied this struct from `cargo`, but could refactor here to make it better

ascjones: Opted for explicit variants rather than options

src/workspace.rs
ascjones: See what you think of the explicit variants

PR #62, opened 2020-07-03
Generate extended contract metadata

### LINKED ISSUE #57: Extend contract metadata with cargo metadata
Rel https://github.com/paritytech/ink/issues/299

Cargo metadata should be used to extract the following and pass it to the `ink_abi`:

e.g. 
- contract version
- ink! version 
- rust version, 
- author,
- license, etc.

If we are able to generate the `main.rs` (PR incoming) of the `.ink/abi_gen` then we can set those values directly e.g.

```
ink_abi::Metadata::build()
    .layout(contract_layout) // generated by ink!
    .general(contract_info) // generated by ink!
    .meta(cargo_toml_info) // injected by cargo-contract
    .finalize()
```

ascjones:
Closes #57. Part of https://github.com/paritytech/ink/issues/299.

Generates extended metadata described here: https://github.com/paritytech/ink/issues/299#issuecomment-651132343 (reproduced below)

- In root:
    - `metadata_version`: Required semantic version for the smart contract metadata.
- In `source` section:
    - `hash`: Required of the associated Wasm file using BLAKE2 256-bit
    - `language`: e.g. Solidity, AssemblyScript or Rust (including version X.Y.Z (semantic versioning))
    - `compiler`: e.g. Solang, ink!/Rust (including version X.Y.Z (semantic versioning))
- In `contract` section:
    - `name`: Required identifier for the smart contract name. (https://crates.io/crates/unicode-xid)
    - `version`: Required semantic version field for the smart contract.
    - `authors`: Required non-empty sequence of strings.
    - `description:` Optional short description about the smart contract.
    - `documentation`: Optional URL to link to API-level documentation of the smart contract.
    - `repository`: Optional URL to link to the repository.
    - `homepage`: Optional URL to link to the smart contract's associated homepage.
    - `license`: Optional field to identify the license of the smart contract given in SPDX license list 3.6 format OR a URL to a user provided non-standard license.
- Add `user` section for custom metadata fields that are generally ignored by standard tools.

Types for the extended metadata are defined in `cargo-contract` and combined with metadata generated by `ink_lang`. This maintains minimal coupling with `ink!` allowing for forwards/backwards compatibility. See the reasoning here https://github.com/paritytech/ink/issues/299#issuecomment-656750372.


src/cmd/metadata/contract.rs
```diff
@@ -0,0 +1,386 @@
...
+pub struct ContractMetadata {
+    metadata_version: semver::Version,
+    source: InkProjectSource,
+    contract: InkProjectContract,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    user: Option<InkProjectUser>,
+    /// Raw JSON of the metadata generated by the ink! contract itself
+    #[serde(flatten)]
+    ink: Map<String, Value>,
+}
```
Robbepop, line 39: Maybe you have misunderstood me in the call what I expect to live where or maybe there is generally a confusion on my side about this topic, however, what I meant was the following:

We want to have the full definitions (data structures such as `ContractMetadata`) to live in the `ink_metadata` crate. However, we do not want any of the code generation feature in that crate (e.g. what you did with `ToTokens` impls).
The `ink_metadata` should contain all the data structures and definitions for the whole ink! metadata definitions and should be able to serialize (and optionally deserialize) those to JSON and soon binary. That's it for `ink_metadata`.

What I expect to exist in `cargo-contract` is the code generation for what was formerly known by the `ink_abi` tool. So generation of code that is parameterized over what `cargo-contract` was able to query about the contract state. The generated code must be parameterized over the extended ink! metadata but should other than that be fairly straight forward. All the other definitions and algorithms about how ink! metadata is defined and how serialization is done should be in `ink_metadata` crate.

tl;dr: `cargo-contract` queries ink!'s `Cargo.toml` for extended metadata and generates the static code that itself generates the ink! metadata code with definitions from `ink_metadata`.

ascjones: > tl;dr: cargo-contract queries ink!'s Cargo.toml for extended metadata and generates the static code that itself generates the ink! metadata code with definitions from ink_metadata.

Indeed that's how I had it originally.

The main benefit of this approach is that the metadata generation is decoupled from the version of `ink!` for metadata generation (as long as the contract defines an `__ink_generate_metadata` extern). This means the tool could run metadata generation against different versions of `ink!` as it currently does. Also it solves the problem of brittle codegen for combining the extended metadata with the contract metadata.

However if this is not a requirement, and we want to release new `cargo-contract` major versions with `ink!` major versions, then I can move it all back and just do the codegen,

ascjones: See comment here also: https://github.com/paritytech/ink/issues/299#issuecomment-656750372

src/cmd/metadata/contract.rs
```diff
@@ -0,0 +1,241 @@
...
+#[derive(Debug, Serialize)]
+pub struct ContractMetadata {
+    metadata_version: semver::Version,
+    source: Source,
+    contract: Contract,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    user: Option<User>,
+    /// Raw JSON of the metadata generated by the ink! contract itself
+    #[serde(flatten)]
+    ink: Map<String, Value>,
```
Robbepop, line 35: Not optimal this this is not really checked or intransparent to `cargo-contract` but I guess this is to not create some dependency on to `ink_metadata` crate?

src/cmd/metadata/contract.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub fn new(
+        source: Source,
+        contract: Contract,
+        user: Option<User>,
+        ink: Map<String, Value>,
+    ) -> Self {
+        let metadata_version = semver::Version::parse(METADATA_VERSION)
+            .expect("METADATA_VERSION is a valid semver string");
+
+        ContractMetadata {
```
Robbepop, line 49:
```suggestion
        Self {
```

src/cmd/metadata/contract.rs
```diff
@@ -0,0 +1,241 @@
...
+    contract: Contract,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    user: Option<User>,
+    /// Raw JSON of the metadata generated by the ink! contract itself
+    #[serde(flatten)]
+    ink: Map<String, Value>,
+}
+
+impl ContractMetadata {
+    /// Construct a new ContractMetadata
```
Robbepop, line 39: Btw.: For doc comments when you specify a type name you either specify it as you did but wrapped in "`" so that it is correctly highlighted as a type or you do not write it as a type at all. In this example you either write:

- Construct a new `ContractMetadata`
or (more preferably)
- Construct new contract metadata.

Robbepop: At least the above is what I learned when looking at Rust std lib docs.

src/cmd/metadata/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+            let target_dir_arg = format!("--target-dir={}", target_dir.to_string_lossy());
+            let stdout = util::invoke_cargo(
+                "run",
+                &[
+                    "--package",
+                    "metadata-gen",
+                    &manifest_path.cargo_arg(),
+                    &target_dir_arg,
+                    "--release",
+                    // "--no-default-features", // Breaks builds for MacOS (linker errors), we should investigate this issue asap!
```
Robbepop, line 64: Have we investigated or turned this into an issue, yet?

Robbepop: But also I think it is fine that we do not compile the `metadata-gen` with `--no-default-features`. Why did we in the first place?

src/cmd/metadata/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+            self.unstable_options.clone(),
+        )?;
+
+        let wasm = fs::read(&self.crate_metadata.dest_wasm)?;
+
+        use ::blake2::digest::{Update as _, VariableOutput as _};
+        let mut output = [0u8; 32];
+        let mut blake2 = blake2::VarBlake2b::new_keyed(&[], 32);
+        blake2.update(wasm);
+        blake2.finalize_variable(|result| output.copy_from_slice(result));
```
Robbepop, line 153: I think these 3 lines could be replaced with a single call to: https://docs.rs/digest/0.9.0/digest/trait.Digest.html#tymethod.digest

src/cmd/metadata/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+            let license = contract.get("license").expect("contract.license not found");
+
+            let user = metadata_json.get("user").expect("user section not found");
+
+            // calculate wasm hash
+            let wasm = fs::read(&crate_metadata.dest_wasm)?;
+            let mut output = [0u8; 32];
+            let mut blake2 = blake2::VarBlake2b::new_keyed(&[], 32);
+            blake2.update(wasm);
+            blake2.finalize_variable(|result| output.copy_from_slice(result));
```
Robbepop, line 325: Same as above.

Robbepop: Also not sure why you test a functionality by some other functionality with the exact same implementation. Normally when testing hashes I do so with some concrete hashes that I compute once and then take for granted.

src/workspace/mod.rs
```diff
@@ -162,7 +162,7 @@ impl Workspace {
         F: FnOnce(&ManifestPath) -> Result<()>,
     {
         let tmp_dir = tempfile::Builder::new()
-            .prefix(".cargo-contract_")
+            .prefix("cargo-contract_")
             .tempdir()?;
```
Robbepop, line 166: How important is it that this name does not conflict with any other name? Because if it is important we should maybe uglify it a bit more.

templates/tools/generate-metadata/main.rs
```diff
@@ -5,9 +5,8 @@ extern "Rust" {
 }
 
 fn main() -> Result<(), std::io::Error> {
-	let ink_project = unsafe { __ink_generate_metadata() };
-	let contents = serde_json::to_string_pretty(&ink_project)?;
-	std::fs::create_dir("target").ok();
-	std::fs::write("target/metadata.json", contents)?;
+	let metadata = unsafe { __ink_generate_metadata() };
+	let contents = serde_json::to_string_pretty(&metadata)?;
+	print!("{}", contents);
```
Robbepop, line 10: Interesting trick to go through the `stdout` bytes here. Why was this design chosen?

Cargo.toml
```diff
@@ -54,5 +57,5 @@ wabt = "0.9.2"
 
 [features]
 default = []
-extrinsics = ["sp-core", "subxt", "async-std", "futures", "hex", "url"]
+extrinsics = ["sp-core", "subxt", "async-std", "futures", "hex"]
 test-ci-only = []
```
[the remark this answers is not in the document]
ascjones: Added

src/cmd/metadata/contract.rs
ascjones: Yes exactly that. It's an evolution of the previous approach, which was just writing the serialized metadata produced by `ink!` metadata to a file. It maintains the behaviour of forwards/backwards compatibility with different versions of `ink!`

src/cmd/metadata/mod.rs
ascjones: I've no idea this comment is leftover from Sergei's initial work. I can remove this since it does not seem necessary (we now control the generation of the `metadata-gen` package after all)

src/workspace/mod.rs
ascjones: It's a prefix to a unique generated name e.g. `/tmp/cargo_contract_1afbe3` (I made this up but something like that)

templates/tools/generate-metadata/main.rs
ascjones: Previously when we only had `ink!` metadata, the generated `metadata-gen` package wrote directly to  `target/metadata.json`. 

Now we require combining the `ink!` metadata with the extended contract metadata. We need some way of marshalling the data returned by invoking `__ink_generate_metadata` back to the calling `cargo-contract` process where the metadata can be combined.

Note that in the first iteration of this, the extended metadata generated by `cargo-contract` and the `ink!` contract metadata were combined inside the `ink_metadata` crate, so the data was marshalled in the other direction (via code generation). However as discussed previously this is considerably less flexible in terms of compatibility with different versions of `ink!`.

Using json serialized to `stdout` is a more idiomatic way of inter process communication, and provides us with that property of backwards/forwards compatibility with `ink!` versions.

src/cmd/metadata/mod.rs
ascjones: It is essentially just testing that we are using that same hash function to generate the hash. Also a hardcoded hash could cause this test to break a lot with different versions of `ink!` and `rustc`.

PR #69, opened 2020-08-05
Extract contract metadata to its own crate

### LINKED ISSUE #65: Extract extended contract metadata types to separate crate
Follow up to #57, as dicussed here https://github.com/paritytech/ink/issues/299#issuecomment-658710757.

Types for extended contract metadata can be shared with https://github.com/hyperledger-labs/solang by extracting the [types](https://github.com/paritytech/cargo-contract/blob/4148a33c1393c9bf483216d139dba47d92c4659c/src/cmd/metadata/contract.rs) to their own crate.

To start with, we could add the crate as a workspace package in the `cargo-contract` repo.

ascjones:
Closes #65, rel https://github.com/paritytech/ink/issues/299.

Not sure about the name: have called it `contract-metadata` for now.


metadata/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+//! let contract = Contract::new(
+//!     "incrementer".to_string(),
+//!     Version::new(2, 1, 0),
+//!     vec!["Parity Technologies <admin@parity.io>".to_string()],
+//!     Some("increment a value".to_string()),
+//!     Some(Url::parse("http://!docs.rs/").unwrap()),
+//!     Some(Url::parse("http://!github.com/paritytech/ink/").unwrap()),
+//!     Some(Url::parse("http://!example.com/").unwrap()),
+//!     Some("Apache-2.0".to_string()),
+//! );
```
Robbepop, line 40: Looks like a builder pattern would be appropriate here.

metadata/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+    pub fn new(
+        name: String,
+        version: Version,
+        authors: Vec<String>,
+        description: Option<String>,
+        documentation: Option<Url>,
+        repository: Option<Url>,
+        homepage: Option<Url>,
+        license: Option<String>,
+    ) -> Self {
```
Robbepop, line 234: Personally I'd prefer a builder pattern. This would be especially handy for all these optional parameters. Doesn't need to be a fancy build pattern with compile-time checks in this case.

ascjones:
@Robbepop ready for another look now I've added the builder

metadata/lib.rs
```diff
@@ -28,16 +28,17 @@
...
-//! );
+//! let contract = Contract::builder()
+//!     .name("incrementer".to_string())
+//!     .version(Version::new(2, 1, 0))
+//!     .authors(vec!["Parity Technologies <admin@parity.io>".to_string()])
+//!     .description(Some("increment a value".to_string()))
+//!     .documentation(Some(Url::parse("http:docs.rs/").unwrap()))
+//!     .repository(Some(Url::parse("http:github.com/paritytech/ink/").unwrap()))
+//!     .homepage(Some(Url::parse("http:example.com/").unwrap()))
+//!     .license(Some("Apache-2.0".to_string()))
```
Robbepop, line 39: I would expect from a builder pattern to receive the arguments as `T` and not `Option<T>` so if you just do not use `.description(whatever)` the `description` of a contract would implicitly be `None`.

metadata/lib.rs
```diff
@@ -259,6 +241,120 @@ impl User {
     }
 }
 
+/// Builder for contract metadata
+#[derive(Default)]
+pub struct ContractBuilder {
+    name: Option<String>,
+    version: Option<Version>,
+    authors: Option<Vec<String>>,
```
Robbepop, line 250: I think it is better to have just `Vec<String>` here since an empty vector is basically similar to no authors. The `authors` setter should check that the given `authors` parameter is non-empty though and that the already set `authors` is still empty.

metadata/lib.rs
```diff
@@ -259,6 +241,120 @@ impl User {
...
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        self.authors = Some(authors.into_iter().map(|s| s.as_ref().to_string()).collect());
+        self
+    }
+
+    /// Set the contract description (optional)
+    pub fn description<S>(&mut self, description: Option<S>) -> &mut Self
```
Robbepop, line 290: Should be
```
pub fn description<S>(&mut self, description: S) -> &mut Self
where
    S: AsRef<str>,
```
and also should check if `description` is still empty. (assert)

metadata/lib.rs
```diff
@@ -259,6 +241,120 @@ impl User {
...
+    {
+        self.description = description.map(|s| s.as_ref().to_string());
+        self
+    }
+
+    /// Set the contract documentation url (optional)
+    pub fn documentation(&mut self, documentation: Option<Url>) -> &mut Self {
+        self.documentation = documentation;
+        self
+    }
```
Robbepop, line 302: Same as in `description`

metadata/lib.rs
```diff
@@ -259,6 +241,120 @@ impl User {
...
+
+    /// Set the contract repository url (optional)
+    pub fn repository(&mut self, repository: Option<Url>) -> &mut Self {
+        self.repository = repository;
+        self
+    }
```
Robbepop, line 308: Same as in `description`

metadata/lib.rs
```diff
@@ -259,6 +241,120 @@ impl User {
...
+
+    /// Set the contract homepage url (optional)
+    pub fn homepage(&mut self, homepage: Option<Url>) -> &mut Self
+    {
+        self.homepage = homepage;
+        self
+    }
```
Robbepop: Same as in `description`

metadata/lib.rs
```diff
@@ -259,6 +241,120 @@ impl User {
...
+
+    /// Set the contract license (optional)
+    pub fn license<S>(&mut self, license: Option<S>) -> &mut Self
+        where
+            S: AsRef<str>
+    {
+        self.license = license.map(|s| s.as_ref().to_string());
+        self
+    }
```
Robbepop: Same as in `description`

metadata/lib.rs
```diff
@@ -259,6 +241,120 @@ impl User {
...
+            if self.name.is_none() {
+                required.push("name");
+            }
+            if self.version.is_none() {
+                required.push("version")
+            }
+            if self.authors.is_none() {
+                required.push("authors")
+            }
+            Err(format!("Missing required non-default fields: {:?}", required))
```
Robbepop: Nice!

metadata/lib.rs
ascjones: Yes I had it this way originally, however in the place where it is used https://github.com/paritytech/cargo-contract/blob/ea01a8a540faa3c718e84d1bf3003db09315b5fb/src/cmd/metadata.rs#L119-L129 the values I am passing are already `Option`s, so it required lots of `if let Some(opt) = opt { builder.opt(opt) }`.

Could have both by adding an extra method for each field e.g. `description_opt`?

metadata/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        self.authors = Some(
+            authors
+                .into_iter()
+                .map(|s| s.as_ref().to_string())
+                .collect(),
+        );
+        self
```
Robbepop, line 286: misses asserts to check if old `authors` is empty and new `authors` is non-empty

metadata/lib.rs
ascjones: I have this as an option to make this pattern possible: https://github.com/paritytech/cargo-contract/blob/50b585b8da8856fe291d1936316729d64a99329b/metadata/lib.rs#L346

metadata/lib.rs
ascjones: Done

Robbepop (approved):
LGTM now
let's wait for the checks to pass and then you are free to merge

PR #86, opened 2020-10-22
Pin nightly version to fix build

ascjones:
See if this fixes the CI


ascjones:
Hmm no joy because we need to install default toole e.g. `fmt` etc. 

`error: 'cargo-fmt' is not installed for the toolchain 'nightly-2020-10-01-x86_64-unknown-linux-gnu'`

We could add the install of those to the script, or we vould update the base image to be pinned to this nightly? @TriplEight

TriplEight:
@ascjones here https://github.com/paritytech/scripts/pull/231, also I'm stoppint the nightly rebuild for your CI image, please ping me in the scripts repo when we will be able to return to the normal flow.

TriplEight:
This way (`$RUSTUP_TOOLCHAIN`) of overriding a nightly version does not take into account everything that is installed, as we as this one https://github.com/substrate-developer-hub/substrate-node-template/pull/98. 
AFAIK there's no good way to "override and substitute", or, say, downgrade the toolchain with Rust built-in tools.

PR #90, opened 2020-10-26
Some modern CI features

TriplEight:
- cache prepopulation
- `rules` instead of `only`
- rm some unnecessary vars


TriplEight:
Will work after the image from https://github.com/paritytech/scripts/pull/232 is built.

ascjones (approved):
LGTM

PR #93, opened 2020-10-29
Specify optional --manifest-path for build and generate-metadata

ascjones:
Allows building contracts/generating metadata from a location outside the working directory.

Usage:

```
cargo run -- contract build --manifest-path ../ink/examples/erc2/Cargo.toml
cargo run -- contract generate-metadata --manifest-path ../ink/examples/erc20/Cargo.toml    
```

The `--manifest-path` flag is so named from the equivalent option in `cargo` which has the same effect.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -31,6 +31,7 @@ use subxt::PairSigner;
 use anyhow::{Error, Result};
 use colored::Colorize;
 use structopt::{clap, StructOpt};
+use crate::workspace::ManifestPath;
 
 #[derive(Debug, StructOpt)]
 #[structopt(bin_name = "cargo")]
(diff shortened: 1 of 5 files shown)


cmichi (approved):
Great, thx!

src/main.rs
```diff
@@ -163,6 +165,9 @@ enum Command {
     /// Compiles the smart contract
     #[structopt(name = "build")]
     Build {
+        /// Path to the Cargo.toml of the contract to build.
```
cmichi, line 168: Nitpick: You don't end the other comments with `.` and since these will show up on the cli interface I think they should be consistent.

```suggestion
        /// Path to the Cargo.toml of the contract to build
```

PR #95, opened 2020-10-30
Use binaryen-rs as dep instead of requiring manual wasm-opt installation

### LINKED ISSUE #91: Fail the build if wasm-opt not installed
Currently a warning is displayed if `wasm-opt` is not installed. However some users are not reading the warning.

https://github.com/paritytech/cargo-contract/blob/d28a743b32a77d2ab3ddb463d101e1fe91344b15/src/cmd/build.rs#L201-L208

Since contract authors would definitely want their contracts optimized (storage rent!), we should fail the build if it is not installed. They won't be able to ignore that.

If for whatever reason they want to still compile without installing `wasm-opt`, we should provide an option e.g. `--no-wasm-opt` or `--skip-wasm-optimizations`

## Alternative

Investigate using https://github.com/pepyakin/binaryen-rs as a dependency so as not to require manual installation of `wasm-opt`

cmichi:
Closes #91.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,6 +16,8 @@ variables:
   CARGO_HOME:                      "/ci-cache/${CI_PROJECT_NAME}/cargo/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   RUST_LIB_BACKTRACE:              "0"
+
+  # Necessary for building binaryen-sys, which is part of the binaryen dependency
   CXX:                             "/usr/bin/clang++-8"
 
 workflow:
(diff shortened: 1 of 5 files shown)


cmichi:
CI fails with
```
Job's log exceeded limit of 4194304 bytes.
```
That's because the compilation output of `binaryen` is huge. I've already contacted Denis about increasing the output limit.

.gitlab-ci.yml
```diff
@@ -17,6 +17,9 @@ variables:
   CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   RUST_LIB_BACKTRACE:              "0"
 
+  # Necessary for building binaryen-sys, which is part of the binaryen dependency
+  CXX:                             "/usr/bin/clang++-8"
```
TriplEight, line 21: are you sure you want to put it on the top level? Means everything will be linked anew now. Similarly, we could put it in the CI image.

cmichi: I'm not sure about possible side-effects it would have in the CI image for other projects using the image.

Regarding top-level: It's needed for all the stages in the file (`test`, `build`, `fmt`), so yeah linking anew is needed for all of them.

TriplEight: It's just ink that uses `ink-ci-linux` image currently. So yeah, if yoou need it everywhere, and if you'd use the CI image as a build local env (as it's possible with vscode), and maybe if someone compiles ink! externally, that'd make sense to introduce this variable to an image.

cmichi: Good point, I created https://github.com/paritytech/scripts/pull/235.

ascjones (approved):
:heart: this

TriplEight:
@ascjones don't you need the same change in your CI env?

cmichi:
I updated the `scripts` PR to include it in both the ink! base image and the `cargo-contracts` one. I had thought there was only one image for both projects.

TriplEight:
cargo-contract is installing fine to the CI image, it's this what needs to be fixed: https://github.com/paritytech/canvas-node/issues/21

PR #79, opened 2020-09-22
Interact with contracts: upload, instantiate and call commands

ascjones:
Adds `upload`, `instantiate` and `call` commands which allows interacting with contracts on live chains with a compatible `pallet-contracts`.

## Upload

```bash
cargo contract upload --suri //Alice --dry-run
```

image

```bash
cargo contract upload --suri //Alice
```

image

## Instantiate

```bash
cargo contract instantiate \
       --constructor new \
       --args false \
       --suri //Alice \
       --code-hash 0xbc1b42256696c8a4187ec3ed79fc602789fc11287c4c30926f5e31ed8169574e \
       --dry-run

```

image

```bash
cargo contract instantiate \
       --constructor new \
       --args false \
       --suri //Alice \
       --code-hash 0xbc1b42256696c8a4187ec3ed79fc602789fc11287c4c30926f5e31ed8169574e
```

image

### Instantiate with code

Omitting `--code-hash` will call `instantiate_with_code`:

image

## Call

```bash
cargo contract call \
       --contract 5FQNUFjuSxjB1o67CneHXieB1Hy6xAj2A4h4dUVLCvNkgRmD \
       --message flip \
       --suri //Alice \
       --dry-run
```

image

```bash
cargo contract call \
       --contract 5FQNUFjuSxjB1o67CneHXieB1Hy6xAj2A4h4dUVLCvNkgRmD \
       --message flip \
       --suri //Alice
```

image

```sh
cargo contract call \
       --contract 5FQNUFjuSxjB1o67CneHXieB1Hy6xAj2A4h4dUVLCvNkgRmD \
       --message get \
       --suri //Alice \
       --dry-run
```

### Debug message

If the contract is build in debug mode then 

image

Will result in...

image

# Todo

- [x] Display all events resulting from `Call`
- [x] Optional `0x` prefix for hex encoded code hash
- [x] SS58 AccountIds
- [x] Release and use `pallet-contracts-primitives` and `sp-rpc` from `crates.io`
- [x] Documentation
  - [x] CLI usage help
  - [x] extrinsics.md
  - [x] transcoding/scon code docs 

# Follow ups

These will be promoted to their own issue, or into separate issues for following up once this is merged.

- Signing: safer alternatives for value bearing chains than providing secret key on the command line. Possibly allow using encrypted file with private key, or integrating with external signers somehow.
- Dynamic `DispatchError::Module` errors, so that error details are displayed correctly across different chains https://github.com/paritytech/subxt/issues/443
[...]

PR #97, opened 2020-11-02
Implement new commands `build` and `check` + introduce bundles (.contract files)

### LINKED ISSUE #96: Add cargo contract pack command
The proposed `cargo contract pack` is going to do the following:
- Builds the contract: `cargo contract build`
- Builds the metadata of the contract: `cargo contract metadata`
- After both steps with their respective outputs in the `target/ink` directory the `cargo contract pack` command additionally packs the generated `contract.wasm` into the `source.wasm` field of the `metadata.json` file and stores it as `contract.pack`.

As follow-up the Canvas UI (and eventually Polkadot JS Apps) should build in support for the `.pack` extension and might even be able to drop support for non packed contracts completely.

cmichi:
Closes #96 by implementing

* a new command `cargo contract build`, which results in a `target/metadata.json`, a `target/<name>.wasm`, and a `target/<name>.contract` (the .pack file mentioned in the issue).
* a new command `cargo contract check`, which just tries to build the contract without generating metadata or optimizing the Wasm.
* deprecates `cargo contract generate-metadata` for `cargo contract build` (which implicitly does this or via `--metadata-only` skips unnecessary operations not strictly needes for generating metadata).

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,8 @@ OPTIONS:
 
 SUBCOMMANDS:
     new                  Setup and create a new smart contract project
-    build                Compiles the smart contract
-    generate-metadata    Generate contract metadata artifacts
+    build                Compiles the contract, generates metadata, bundles both together in a '.contract' file
+    check                Check that the Wasm builds; does not optimize, generate metadata, or bundle
     test                 Test the smart contract off-chain
     deploy               Upload the smart contract code to the chain
     instantiate          Instantiate a deployed smart contract
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ OPTIONS:
 SUBCOMMANDS:
     new                  Setup and create a new smart contract project
     build                Compiles the contract, generates metadata, bundles both together in a '.contract' file
-    check                Check that the Wasm builds; does not optimize, generate metadata, or bundle
+    check                Check that the code builds as Wasm; does not output any build artifact to the top level `target/` directory
     test                 Test the smart contract off-chain
     deploy               Upload the smart contract code to the chain
     instantiate          Instantiate a deployed smart contract
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -28,7 +28,7 @@
 //! let language = SourceLanguage::new(Language::Ink, Version::new(2, 1, 0));
 //! let compiler = SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
 //! let wasm = SourceWasm::new(vec![0u8]);
-//! let source = Source::new(Some(wasm), Some([0u8; 32]), language, compiler);
+//! let source = Source::new(Some(wasm), Some(CodeHash([0u8; 32])), language, compiler);
 //! let contract = Contract::builder()
 //!     .name("incrementer".to_string())
 //!     .version(Version::new(2, 1, 0))
@@ -522,7 +522,7 @@ mod tests {
         let compiler =
             SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
         let wasm = SourceWasm::new(vec![0u8, 1u8, 2u8]);
-        let source = Source::new(Some(wasm), Some([0u8; 32]), language, compiler);
+        let source = Source::new(Some(wasm), Some(CodeHash([0u8; 32])), language, compiler);
         let contract = Contract::builder()
             .name("incrementer".to_string())
             .version(Version::new(2, 1, 0))
@@ -604,7 +604,7 @@ mod tests {
         let language = SourceLanguage::new(Language::Ink, Version::new(2, 1, 0));
         let compiler =
             SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
-        let source = Source::new(None, Some([0u8; 32]), language, compiler);
+        let source = Source::new(None, Some(CodeHash([0u8; 32])), language, compiler);
         let contract = Contract::builder()
             .name("incrementer".to_string())
             .version(Version::new(2, 1, 0))
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -33,9 +33,9 @@
 //!     .version(Version::new(2, 1, 0))
 //!     .authors(vec!["Parity Technologies <admin@parity.io>".to_string()])
 //!     .description("increment a value".to_string())
-//!     .documentation(Url::parse("http:docs.rs/").unwrap())
-//!     .repository(Url::parse("http:github.com/paritytech/ink/").unwrap())
-//!     .homepage(Url::parse("http:example.com/").unwrap())
+//!     .documentation(Url::parse("http://docs.rs/").unwrap())
+//!     .repository(Url::parse("http://github.com/paritytech/ink/").unwrap())
+//!     .homepage(Url::parse("http://example.com/").unwrap())
 //!     .license("Apache-2.0".to_string())
 //!     .build()
 //!     .unwrap();
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -99,13 +99,23 @@ impl ContractMetadata {
     }
 }
 
+/// Representation of the Wasm code hash.
+#[derive(Debug, Eq, PartialEq)]
+pub struct CodeHash(pub [u8; 32]);
+
+impl Serialize for CodeHash {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serialize_as_byte_str(&self.0[..], serializer)
+    }
+}
+
 #[derive(Debug, Serialize)]
 pub struct Source {
-    #[serde(
-        serialize_with = "serialize_as_byte_str_foo",
-        skip_serializing_if = "Option::is_none"
-    )]
-    hash: Option<[u8; 32]>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    hash: Option<CodeHash>,
     language: SourceLanguage,
     compiler: SourceCompiler,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -116,7 +126,7 @@ impl Source {
     /// Constructs a new InkProjectSource.
     pub fn new(
         wasm: Option<SourceWasm>,
-        hash: Option<[u8; 32]>,
+        hash: Option<CodeHash>,
         language: SourceLanguage,
         compiler: SourceCompiler,
     ) -> Self {
@@ -434,14 +444,6 @@ impl ContractBuilder {
             ))
         }
     }
-}
-
-fn serialize_as_byte_str_foo<S>(bytes: &Option<[u8; 32]>, serializer: S) -> Result<S::Ok, S::Error>
-where
-    S: serde::Serializer,
-{
-    let bytes = bytes.expect("must_exist");
-    serialize_as_byte_str(&bytes[..], serializer)
 }
 
 /// Serializes the given bytes as byte string.
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -28,7 +28,7 @@
 //! let language = SourceLanguage::new(Language::Ink, Version::new(2, 1, 0));
 //! let compiler = SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
 //! let wasm = SourceWasm::new(vec![0u8]);
-//! let source = Source::new(Some(wasm), [0u8; 32], language, compiler);
+//! let source = Source::new(Some(wasm), Some([0u8; 32]), language, compiler);
 //! let contract = Contract::builder()
 //!     .name("incrementer".to_string())
 //!     .version(Version::new(2, 1, 0))
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -520,7 +520,7 @@ mod tests {
         let compiler =
             SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
         let wasm = SourceWasm::new(vec![0u8, 1u8, 2u8]);
-        let source = Source::new(Some(wasm), [0u8; 32], language, compiler);
+        let source = Source::new(Some(wasm), Some([0u8; 32]), language, compiler);
         let contract = Contract::builder()
             .name("incrementer".to_string())
             .version(Version::new(2, 1, 0))
@@ -602,7 +602,7 @@ mod tests {
         let language = SourceLanguage::new(Language::Ink, Version::new(2, 1, 0));
         let compiler =
             SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
-        let source = Source::new(None, [0u8; 32], language, compiler);
+        let source = Source::new(None, Some([0u8; 32]), language, compiler);
         let contract = Contract::builder()
             .name("incrementer".to_string())
             .version(Version::new(2, 1, 0))
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -101,8 +101,11 @@ impl ContractMetadata {
 
 #[derive(Debug, Serialize)]
 pub struct Source {
-    #[serde(serialize_with = "serialize_as_byte_str")]
-    hash: [u8; 32],
+    #[serde(
+        serialize_with = "serialize_as_byte_str_foo",
+        skip_serializing_if = "Option::is_none"
+    )]
+    hash: Option<[u8; 32]>,
     language: SourceLanguage,
     compiler: SourceCompiler,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -113,7 +116,7 @@ impl Source {
     /// Constructs a new InkProjectSource.
     pub fn new(
         wasm: Option<SourceWasm>,
-        hash: [u8; 32],
+        hash: Option<[u8; 32]>,
         language: SourceLanguage,
         compiler: SourceCompiler,
     ) -> Self {
@@ -431,6 +434,14 @@ impl ContractBuilder {
             ))
         }
     }
+}
+
+fn serialize_as_byte_str_foo<S>(bytes: &Option<[u8; 32]>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    let bytes = bytes.expect("must_exist");
+    serialize_as_byte_str(&bytes[..], serializer)
 }
 
 /// Serializes the given bytes as byte string.
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -61,7 +61,7 @@ use url::Url;
 const METADATA_VERSION: &str = "0.1.0";
 
 /// Smart contract metadata.
-#[derive(Debug, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 pub struct ContractMetadata {
     #[serde(rename = "metadataVersion")]
     metadata_version: semver::Version,
@@ -100,7 +100,7 @@ impl ContractMetadata {
 }
 
 /// Representation of the Wasm code hash.
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct CodeHash(pub [u8; 32]);
 
 impl Serialize for CodeHash {
@@ -112,7 +112,7 @@ impl Serialize for CodeHash {
     }
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 pub struct Source {
     #[serde(skip_serializing_if = "Option::is_none")]
     hash: Option<CodeHash>,
@@ -140,7 +140,7 @@ impl Source {
 }
 
 /// The bytes of the compiled Wasm smart contract.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct SourceWasm {
     wasm: Vec<u8>,
 }
@@ -172,7 +172,7 @@ impl Display for SourceWasm {
 }
 
 /// The language and version in which a smart contract is written.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct SourceLanguage {
     language: Language,
     version: Version,
@@ -201,7 +201,7 @@ impl Display for SourceLanguage {
 }
 
 /// The language in which the smart contract is written.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum Language {
     Ink,
     Solidity,
@@ -219,7 +219,7 @@ impl Display for Language {
 }
 
 /// A compiler used to compile a smart contract.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct SourceCompiler {
     compiler: Compiler,
     version: Version,
@@ -247,7 +247,7 @@ impl SourceCompiler {
 }
 
 /// Compilers used to compile a smart contract.
-#[derive(Debug, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 pub enum Compiler {
     RustC,
     Solang,
@@ -263,7 +263,7 @@ impl Display for Compiler {
 }
 
 /// Metadata about a smart contract.
-#[derive(Debug, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 pub struct Contract {
     name: String,
     version: Version,
@@ -287,7 +287,7 @@ impl Contract {
 }
 
 /// Additional user defined metadata, can be any valid json.
-#[derive(Debug, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 pub struct User {
     #[serde(flatten)]
     json: Map<String, Value>,
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -94,7 +94,7 @@ impl ContractMetadata {
         }
     }
 
-    pub fn remove_wasm(&mut self) {
+    pub fn remove_source_wasm_attribute(&mut self) {
         self.source.wasm = None;
     }
 }
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -27,7 +27,8 @@
 //!
 //! let language = SourceLanguage::new(Language::Ink, Version::new(2, 1, 0));
 //! let compiler = SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
-//! let source = Source::new([0u8; 32], language, compiler);
+//! let wasm = SourceWasm::new(vec![0u8]);
+//! let source = Source::new(Some(wasm), [0u8; 32], language, compiler);
 //! let contract = Contract::builder()
 //!     .name("incrementer".to_string())
 //!     .version(Version::new(2, 1, 0))
@@ -100,16 +101,56 @@ pub struct Source {
     hash: [u8; 32],
     language: SourceLanguage,
     compiler: SourceCompiler,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    wasm: Option<SourceWasm>,
 }
 
 impl Source {
     /// Constructs a new InkProjectSource.
-    pub fn new(hash: [u8; 32], language: SourceLanguage, compiler: SourceCompiler) -> Self {
+    pub fn new(
+        wasm: Option<SourceWasm>,
+        hash: [u8; 32],
+        language: SourceLanguage,
+        compiler: SourceCompiler,
+    ) -> Self {
         Source {
             hash,
             language,
             compiler,
+            wasm,
         }
+    }
+}
+
+/// The bytes of the compiled Wasm smart contract.
+#[derive(Debug)]
+pub struct SourceWasm {
+    wasm: Vec<u8>,
+}
+
+impl SourceWasm {
+    /// Constructs a new `SourceWasm`.
+    pub fn new(wasm: Vec<u8>) -> Self {
+        SourceWasm { wasm }
+    }
+}
+
+impl Serialize for SourceWasm {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serialize_as_byte_str(&self.wasm[..], serializer)
+    }
+}
+
+impl Display for SourceWasm {
+    fn fmt(&self, f: &mut Formatter<'_>) -> DisplayResult {
+        write!(f, "0x").expect("failed writing to string");
+        for byte in &self.wasm {
+            write!(f, "{:02x}", byte).expect("failed writing to string");
+        }
+        write!(f, "")
     }
 }
 
@@ -463,7 +504,8 @@ mod tests {
         let language = SourceLanguage::new(Language::Ink, Version::new(2, 1, 0));
         let compiler =
             SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
-        let source = Source::new([0u8; 32], language, compiler);
+        let wasm = SourceWasm::new(vec![0u8, 1u8, 2u8]);
+        let source = Source::new(Some(wasm), [0u8; 32], language, compiler);
         let contract = Contract::builder()
             .name("incrementer".to_string())
             .version(Version::new(2, 1, 0))
@@ -507,7 +549,8 @@ mod tests {
                 "source": {
                     "hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
                     "language": "ink! 2.1.0",
-                    "compiler": "rustc 1.46.0-nightly"
+                    "compiler": "rustc 1.46.0-nightly",
+                    "wasm": "0x000102"
                 },
                 "contract": {
                     "name": "incrementer",
@@ -544,7 +587,7 @@ mod tests {
         let language = SourceLanguage::new(Language::Ink, Version::new(2, 1, 0));
         let compiler =
             SourceCompiler::new(Compiler::RustC, Version::parse("1.46.0-nightly").unwrap());
-        let source = Source::new([0u8; 32], language, compiler);
+        let source = Source::new(None, [0u8; 32], language, compiler);
         let contract = Contract::builder()
             .name("incrementer".to_string())
             .version(Version::new(2, 1, 0))
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -93,6 +93,10 @@ impl ContractMetadata {
             abi,
         }
     }
+
+    pub fn remove_wasm(&mut self) {
+        self.source.wasm = None;
+    }
 }
 
 #[derive(Debug, Serialize)]
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -33,6 +33,16 @@ use parity_wasm::elements::{External, MemoryType, Module, Section};
 /// This is the maximum number of pages available for a contract to allocate.
 const MAX_MEMORY_PAGES: u32 = 16;
 
+/// Result of the metadata generation process.
+pub struct BuildResult {
+    /// Path to the resulting metadata file.
+    pub dest_metadata: Option<PathBuf>,
+    /// Path to the resulting Wasm file.
+    pub dest_wasm: Option<PathBuf>,
+    /// Path to the bundled file.
+    pub dest_bundle: Option<PathBuf>,
+}
+
 /// Builds the project in the specified directory, defaults to the current directory.
 ///
 /// Uses [`cargo-xbuild`](https://github.com/rust-osdev/cargo-xbuild) for maximum optimization of
@@ -240,16 +250,40 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
 pub(crate) fn execute(
     manifest_path: &ManifestPath,
     verbosity: Option<Verbosity>,
-    unstable_options: UnstableFlags,
     optimize_contract: bool,
-) -> Result<Option<PathBuf>> {
+    skip_bundle: bool,
+    skip_metadata: bool,
+    unstable_options: UnstableFlags,
+) -> Result<BuildResult> {
     let crate_metadata = CrateMetadata::collect(manifest_path)?;
-    execute_with_metadata(
-        &crate_metadata,
+    if skip_metadata {
+        let dest_wasm = execute_with_metadata(
+            &crate_metadata,
+            verbosity,
+            optimize_contract,
+            unstable_options,
+        )?;
+        let res = BuildResult {
+            dest_wasm: dest_wasm,
+            dest_metadata: None,
+            dest_bundle: None,
+        };
+        return Ok(res);
+    }
+
+    let metadata_result = super::metadata::execute(
+        &manifest_path,
         verbosity,
-        unstable_options,
+        !skip_bundle,
         optimize_contract,
-    )
+        unstable_options,
+    )?;
+    let res = BuildResult {
+        dest_wasm: Some(metadata_result.wasm_file),
+        dest_metadata: Some(metadata_result.metadata_file),
+        dest_bundle: metadata_result.bundle_file,
+    };
+    Ok(res)
 }
 
 /// Executes build of the smart-contract which produces a wasm binary that is ready for deploying.
@@ -262,8 +296,8 @@ pub(crate) fn execute(
 pub(crate) fn execute_with_metadata(
     crate_metadata: &CrateMetadata,
     verbosity: Option<Verbosity>,
-    unstable_options: UnstableFlags,
     optimize_contract: bool,
+    unstable_options: UnstableFlags,
 ) -> Result<Option<PathBuf>> {
     println!(
         " {} {}",
@@ -300,7 +334,7 @@ mod tests {
             cmd::new::execute("new_project", Some(path)).expect("new project creation failed");
             let manifest_path =
                 ManifestPath::new(&path.join("new_project").join("Cargo.toml")).unwrap();
-            super::execute(&manifest_path, None, UnstableFlags::default(), true)
+            super::execute(&manifest_path, None, UnstableFlags::default(), true, true)
                 .expect("build failed");
             Ok(())
         })
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -339,8 +339,15 @@ mod tests {
             cmd::new::execute("new_project", Some(path)).expect("new project creation failed");
             let manifest_path =
                 ManifestPath::new(&path.join("new_project").join("Cargo.toml")).unwrap();
-            super::execute(&manifest_path, None, true, false, false, UnstableFlags::default())
-                .expect("build failed");
+            super::execute(
+                &manifest_path,
+                None,
+                true,
+                false,
+                false,
+                UnstableFlags::default(),
+            )
+            .expect("build failed");
             Ok(())
         })
     }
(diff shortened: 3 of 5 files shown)


src/cmd/metadata.rs
```diff
@@ -287,9 +304,13 @@ mod tests {
...
-            let metadata_file =
-                cmd::metadata::execute(test_manifest.manifest_path, None, UnstableFlags::default())
-                    .expect("generate metadata failed");
+            let metadata_file = cmd::metadata::execute(
+                test_manifest.manifest_path,
+                None,
+                true,
+                UnstableFlags::default(),
+            )
+            .expect("generate metadata failed");
```
ascjones, line 313: I know this `expect` was here before but maybe should be replaced by a `?`

src/cmd/metadata.rs
```diff
@@ -331,17 +353,14 @@ mod tests {
             let user = metadata_json.get("user").expect("user section not found");
 
             // calculate wasm hash
-            let wasm = fs::read(&crate_metadata.dest_wasm)?;
+            let fs_wasm = fs::read(&crate_metadata.dest_wasm)?;
```
ascjones, line 356: Could reuse the `blake2_hash` function you added

src/cmd/metadata.rs
```diff
@@ -112,7 +115,16 @@ impl GenerateMetadataCommand {
         let source = {
             let lang = SourceLanguage::new(Language::Ink, ink_version.clone());
             let compiler = SourceCompiler::new(Compiler::RustC, rust_version);
-            Source::new(hash, lang, compiler)
+            let maybe_wasm = match self.include_wasm {
```
ascjones, line 118: An `if` expression is more idiomatic and readable I think

src/cmd/metadata.rs
```diff
@@ -34,6 +36,7 @@ const METADATA_FILE: &str = "metadata.json";
 struct GenerateMetadataCommand {
     crate_metadata: CrateMetadata,
     verbosity: Option<Verbosity>,
+    include_wasm: bool,
```
ascjones, line 39: We should consider producing a different file name, perhaps `<contract_name>.pack` or such as @Robbepop  suggested in #96.

ascjones (commented):
What do you think about giving the resulting file a different name/extension than just `metadata.json`, so that we know we are dealing with a packed file? Might be useful for the `canvas-ui` to know to expect a packed file. @Robbepop?

src/cmd/metadata.rs
Robbepop: After a run with `cargo contract pack` we expect as output 3 files: the `.wasm`, the `metadata.json` and the `.pack` file.

src/cmd/metadata.rs
```diff
@@ -21,33 +21,49 @@ use crate::{
...
 impl GenerateMetadataCommand {
-    pub fn exec(&self) -> Result<PathBuf> {
+    pub fn exec(&self) -> Result<GenerateMetadataResult> {
         util::assert_channel()?;
         println!("  Generating metadata");
 
         let cargo_meta = &self.crate_metadata.cargo_meta;
-        let out_path = cargo_meta.target_directory.join(METADATA_FILE);
+        let out_path = if self.include_wasm {
+            let fname = format!("{}.pack", self.crate_metadata.package_name);
```
athei, line 58: Just a little bike shedding: I would prefer `.contract` as a file extensions.

src/main.rs
```diff
@@ -262,20 +273,48 @@ fn exec(cmd: Command) -> Result<String> {
...
+            let pack_result = cmd::metadata::execute(
+                &manifest_path,
+                verbosity.try_into()?,
+                true,
+                unstable_options.try_into()?,
+            )?;
+            Ok(format!(
+                "\nYour contract is ready. You can find it here:\n{}
+                \nYour metadata file is ready. You can find it here:\n{}
+                \nYour packed contract is ready. You can find it here:\n{}",
```
athei, line 297:
```suggestion
                "\nYour contract's code is ready. You can find it here:\n{}
                \nYour contract's metadata is ready. You can find it here:\n{}
                \nYour contract bundle (code + metadata) is ready. You can find it here:\n{}",
```

Strictly speaking the wasm file is only the contract's code. A contract is the instantiation of that code. I am not suggesting that for splitting hairs. I think the message should make more clear what every artefact is.

src/main.rs
```diff
@@ -262,20 +273,48 @@ fn exec(cmd: Command) -> Result<String> {
...
+                &manifest_path,
                 verbosity.try_into()?,
+                false,
                 unstable_options.try_into()?,
             )?;
             Ok(format!(
-                "Your metadata file is ready.\nYou can find it here:\n{}",
-                metadata_file.display()
+                "\nYour metadata file is ready.\nYou can find it here:\n{}",
+                res.metadata_file.display().to_string().bold(),
```
athei, line 317:
```suggestion
                "\nYour metadata is ready.\nYou can find it here:\n{}",
                res.metadata_file.display().to_string().bold(),
```

Calling one thing file and the other don't is inconsistent.

athei (commented):
I think everyone wants to generate that pack. We should make that as easy as possible. Imho the cli should look like this:

There is only the `build` command which does what `pack` does today. It has two options `skip-code` and `skip-metadata`. If either of those is specified we also skip the generation of the pack. Specifying both is an error (obviously).

ascjones:
> I think generating that pack is the default case. We should make that as easy as possible. Imho the cli should look like this:
> 
> There is only the `build` command which does what `pack` does today. It has two options `skip-code` and `skip-metadata`. If either of those is specified we also skip the generation of the pack. Specifying both is an error (obviously).

I like this idea, since in most cases you would want to do both - and incremental compilation means that even compiling both will not cost too much after the first time.

src/cmd/metadata.rs
Robbepop: I am fine with both!

cmichi:
> There is only the build command which does what pack does today. It has two options skip-code and skip-metadata. If either of those is specified we also skip the generation of the pack. Specifying both is an error (obviously).

So I also like it, but I'm not sure about the `--skip-code`. As I understand it, you meant it as "not generating the Wasm blob", but in order to generate the metadata we need to build that blob anyway. So there's no cost in outputting the file name to that one as well.

What I suggest instead is to have the `skip-metadata` like you suggested and then a `skip-packing`:
```
--skip-metadata    Only the Wasm is created, generation of metadata and a packed .contract file is skipped
--skip-packing     Only the Wasm and the metadata are generated, no packed .contract file is created
```
I've added a commit which does that. Lmk what you think.

src/main.rs
```diff
@@ -162,23 +162,18 @@ enum Command {
...
-    },
-    /// Generate contract metadata artifacts
-    #[structopt(name = "generate-metadata")]
-    GenerateMetadata {
-        /// Path to the Cargo.toml of the contract for which to generate metadata
-        #[structopt(long, parse(from_os_str))]
-        manifest_path: Option<PathBuf>,
+        /// Only the Wasm and the metadata are generated, no packed .contract file is created
+        #[structopt(long = "skip-packing", conflicts_with = "skip-metadata")]
+        skip_packing: bool,
```
athei, line 173:
```suggestion
        skip_bundle: bool,
```

src/main.rs
```diff
@@ -162,23 +162,18 @@ enum Command {
...
+        /// Only the Wasm is created, generation of metadata and a packed .contract file is skipped
+        #[structopt(long = "skip-metadata", conflicts_with = "skip-packing")]
+        skip_metadata: bool,
```
athei, line 176:
```suggestion
        /// Only the Wasm and the metadata are generated, no bundled .contract file is created
        #[structopt(long = "skip-bundle", conflicts_with = "skip-metadata")]
        skip_bundle: bool,
        /// Only the Wasm is created, generation of metadata and a packed .contract file is skipped
        #[structopt(long = "skip-metadata", conflicts_with = "skip-bundle")]
        skip_metadata: bool,
```

src/main.rs
```diff
@@ -249,33 +244,50 @@ fn exec(cmd: Command) -> Result<String> {
...
         } => {
+            if *(skip_metadata) {
+                let manifest_path = ManifestPath::try_from(manifest_path.as_ref())?;
+                let dest_wasm = cmd::build::execute(
+                    &manifest_path,
+                    verbosity.try_into()?,
+                    unstable_options.try_into()?,
+                )?;
+                return Ok(format!(
+                    "\nYour contract is ready. You can find it here:\n{}",
```
athei, line 259:
```suggestion
                    "\nYour contract's code is ready. You can find it here:\n{}",
```

For consistency

src/main.rs
```diff
@@ -249,33 +244,50 @@ fn exec(cmd: Command) -> Result<String> {
...
-        }
-        Command::GenerateMetadata {
-            manifest_path,
-            verbosity,
-            unstable_options,
-        } => {
-            let manifest_path = ManifestPath::try_from(manifest_path.as_ref())?;
-            let metadata_file = cmd::metadata::execute(
-                manifest_path,
+            if *(skip_packing) {
```
athei, line 270:
```suggestion
            if *(skip_bundle) {
```

src/main.rs
```diff
@@ -162,23 +162,18 @@ enum Command {
         #[structopt(short, long, parse(from_os_str))]
         target_dir: Option<PathBuf>,
     },
-    /// Compiles the smart contract
+    /// Compiles the contract, generates metadata, packs both together in one <package-name>.contract file
```
athei, line 165:
```suggestion
    /// Compiles the contract, generates metadata, bundles both together in one <package-name>.contract file
```

src/main.rs
```diff
@@ -249,33 +244,50 @@ fn exec(cmd: Command) -> Result<String> {
         Command::Build {
             manifest_path,
             verbosity,
+            skip_packing,
```
athei, line 247:
```suggestion
            skip_bundle,
```

src/main.rs
```diff
@@ -249,33 +244,50 @@ fn exec(cmd: Command) -> Result<String> {
...
-                manifest_path,
+            if *(skip_packing) {
+                return Ok(format!(
+                    "\nYour contract's code is ready. You can find it here:\n{}
+                    \nYour contract's metadata is ready. You can find it here:\n{}",
+                    metadata_result.wasm_file.display().to_string().bold(),
+                    metadata_result.metadata_file.display().to_string().bold(),
+                ));
+            }
+            let pack_result = cmd::metadata::execute(
```
athei, line 278:
```suggestion
            let bundle_result = cmd::metadata::execute(
```

src/main.rs
```diff
@@ -249,33 +244,50 @@ fn exec(cmd: Command) -> Result<String> {
...
             )?;
             Ok(format!(
-                "Your metadata file is ready.\nYou can find it here:\n{}",
-                metadata_file.display()
+                "\nYour contract's code is ready. You can find it here:\n{}
+                \nYour contract's metadata is ready. You can find it here:\n{}
+                \nYour contract bundle (code + metadata) is ready. You can find it here:\n{}",
+                pack_result.wasm_file.display().to_string().bold(),
+                metadata_result.metadata_file.display().to_string().bold(),
+                pack_result.metadata_file.display().to_string().bold()
```
athei, line 290:
```suggestion
                bundle_result.wasm_file.display().to_string().bold(),
                metadata_result.metadata_file.display().to_string().bold(),
                bundle_result.metadata_file.display().to_string().bold()
```

athei (commented):
You are right. `--skip-code` makes no sense. My only complaint is that we should not use the word "packing". I think "bundle" better expresses what we are doing. "Pack" could also mean some kind of compression which is not what we are doing here. I mean we could apply compression on top but the main thing we are doing is still "bundle".

src/main.rs
Robbepop, line 176: Can we please name things positively instead of negatively? I.e. `bundle` or `bundle_contract` or `build_bundle` instead of `skip_bundle`? Same for `skip_metadata`.

cmichi: While I agree with the general sentiment of prioritizing positive naming, I'm not sure how we can do that here, since default behavior is wasm+metadata+bundle and we want to provide a possibility to leave certain things out.

So we need better flag names for 
```
--skip-metadata    Only the Wasm is created, generation of metadata and a packed .contract file is skipped
--skip-packing     Only the Wasm and the metadata are generated, no packed .contract file is created
```

Or do you suggest changing default behavior?

Robbepop: Why do we need these flags in the first place? I think this should be automated just like you don't need to specify to `cargo` which rust files to build.

Robbepop: If your contract implementation has changed you always want to build the Wasm AND if your interfaces have changed you always want to build metadata. This already eliminates the `wasm` flag because bundling pretty much always involved building the `wasm`. So the only question is: Do we want to always build `metadata`? Do we want to infer it? And if so, how? Or do we want to make it skip-opt-in as proposed so far?

cmichi: I don't follow, can you elaborate?

We changed default behavior of `cargo contract build` to generate wasm+metadata+pack. And now we want to provide flags to skip certain steps. Which part of that can be automated?

Robbepop: Btw.: Changing of interfaces is done whenever
- message or constructor signature changes
- docs changing (!)
- types changing, even if it is only a type alias renaming to the same type (!)
- pre- or post conditions are changing
- storage is changing

So in pretty much the same cases as "contract is changing" unless the change is clearly not visible to the outside - e.g. kept in the implementation.

Robbepop: > I don't follow, can you elaborate?
> 
> We changed default behavior of `cargo contract build` to generate wasm+metadata+pack. And now we want to provide flags to skip certain steps. Which part of that can be automated?

Ah, I was not aware of this. This obviously changes perception of this change to me. Hmm...

Robbepop: What is the motivation to change behaviour of `cargo contract build` instead of adding yet another command `cargo contract bundle`?

Robbepop: I estimate that those commands will be used most frequently:

- `cargo contract build` (very often)
- `cargo contract bundle` (rarely)
- `cargo contract metadata` (nearly never)

So we could also remove `cargo contract metadata` and add a `--skip-metadata` flag to `cargo contract bundle` and keep `cargo contract build`.

cmichi: >What is the motivation to change behaviour of cargo contract build instead of adding yet another command cargo contract bundle?

See earlier discussion in this PR: https://github.com/paritytech/cargo-contract/pull/97#pullrequestreview-523161594

cmichi: >So we could also remove cargo contract metadata and add a --skip-metadata flag to cargo contract bundle and keep cargo contract build. Building only metadata seldomly makes real sense.

So what this PR already does is that it removes the `cargo contract generate-metadata` cmd, makes `cargo contract build` build everything by default, and adds skip flags for omitting metadata/bundling.

Robbepop: Then using the most frequently used command forces users to type in `cargo contract build --skip-metadata` which is messy imo. Generating metadat by default is bad here because it adds tons of overhead and users will question ink! with taking 2 minutes to do a fresh build for a hello world contract. We should focus to make the most frequent command the simplest to use:

- Most frequent: `cargo contract build`
- Less frequent: `cargo contract bundle` or `cargo contract bundle --skip-metadata`
- Least frequent: `cargo contract generate-metadata`

Robbepop: Building a bundle without updating the `.wasm` is pretty much always not what you want.

Robbepop:
> I think generating that pack is the default case. We should make that as easy as possible. Imho the cli should look like this:
> 
> There is only the `build` command which does what `pack` does today. It has two options `skip-code` and `skip-metadata`. If either of those is specified we also skip the generation of the pack. Specifying both is an error (obviously).

Building the pack should not be the default case since it incurs tons of overhead due to building `.wasm` AND metadata basically doubling the compile times.

athei:
Isn't that what incremental compilation is for? With metadata not using link time optimization it will not add much time. I predict that people just want a new *.contract file to upload almost always. We should optimize for this case. Who will fumble with two different files and think about what to rebuild in order to save a a second to skip metadata build?

Robbepop:
> Isn't that what incremental compilation is for? With metadata not using link time optimization it will not add much time. I predict that people just want a new *.contract file to upload almost always. We should optimize for this case. Who will fumble with two different files and think about what to rebuild in order to save a a second to skip metadata build?

Have you tried building both `cargo contract build` and `cargo contract generate-metadata`? They build all ink! crates with different crate features and therefore cannot reuse a lot of cached build artifacts from one another. Also `generate-metadata` even builds more stuff generally so tends to be even more of a doubling of fresh build time in general.

athei:
They cannot use the cache from one another but their own cache from a previous run, don't they?

Robbepop:
> They cannot use the cache from one another but their own cache from a previous run, don't they?

Yes they can but still build times are going to be significantly worse when making bundling and all of its extra steps the default.

On my system these are the benchmarks for:
- `cargo contract build` (fresh): 40 seconds
- `cargo contract generate-metadata` (fresh): 120 seconds
- `cargo contract build (incremental): 2.65 seconds
- `cargo contract generate-metadata (incremental): 7.39 seconds
- `cargo contract build (no-op): 0.8 seconds
- `cargo contract generate-metadata (no-op): 2.26 seconds

Incremental builds are measured after `touch src/lib.rs`. No-op builds are measured not changing anything in between.

cmichi:
@Robbepop Why do you think generating a wasm without metadata ist the most common operation a user wants to do? My impression was that we nearly always mention the line `cargo contract build && cargo contract generate-metadata` in docs/tutorials/etc..

Robbepop:
> @Robbepop Why do you think generating a wasm without metadata ist the most common operation a user wants to do? My impression was that we nearly always mention the line `cargo contract build && cargo contract generate-metadata` in docs/tutorials/etc..

Good question!
I compare `cargo contract build` mostly with `cargo check` and `cargo contract bundle` mostly with a composition of `cargo build` and/or `cargo publish`. You need `cargo contract build` in order to check if your contract compiles successfully on Wasm, you need `cargo test` to build tests, you basically never need `cargo contract generate-metadata` and you are going to need `cargo contract bundle` if you plan to deploy on-chain. Note that `cargo contract bundle` is comparable to `cargo publish` because unlike `cargo build` it will also have some additional checks to verify that no important metadata information is missing for deployment.

athei:
I think it is not helpful to benchmark this for a trivial contract and then claim that it adds 4x overhead. It won't scale with the contract compilation time because it is purely the overhead of running the metadata generation at all (I think).

And I think this overhead can be reduced. We should work on that rather than forcing the user to solve the problem for us (deciding when to generate metadata).

> I compare cargo contract build mostly with cargo check and cargo contract bundle mostly with a composition of cargo build and/or cargo publish. You need cargo contract build in order to check if your contract compiles successfully on Wasm, you need cargo test to build tests, you basically never need cargo contract generate-metadata and you are going to need cargo contract bundle if you plan to deploy on-chain. Note that cargo contract bundle is comparable to cargo publish because unlike cargo build it will also have some additional checks to verify that no important metadata information is missing for deployment.

I don't think this is a beginner friendly workflow and something that contradicts rust developer assumptions. The user rightfully can expect that `build` "just builds" a deployable artifact with everything they need and any additional hassle. Why not add a `check` command to cargo contract that does this "quick check" whether it compiles?

Robbepop:
> I think it is not helpful to benchmark this for a trivial contract and then claim that it adds 4x overhead. It won't scale with the contract compilation time because it is purely the overhead of running the metadata generation at all (I think).
> 
> And I think this overhead can be reduced. We should work on that rather than forcing the user to solve the problem for us (deciding when to generate metadata).
> 
> > I compare cargo contract build mostly with cargo check and cargo contract bundle mostly with a composition of cargo build and/or cargo publish. You need cargo contract build in order to check if your contract compiles successfully on Wasm, you need cargo test to build tests, you basically never need cargo contract generate-metadata and you are going to need cargo contract bundle if you plan to deploy on-chain. Note that cargo contract bundle is comparable to cargo publish because unlike cargo build it will also have some additional checks to verify that no important metadata information is missing for deployment.
> 
> I don't think this is a beginner friendly workflow and something that contradicts rust developer assumptions. The user rightfully can expect that `build` "just builds" a deployable artifact with everything they need without any additional hassle. Why not add a `check` command to cargo contract that does this "quick check" whether it compiles?

So to summarize, your proposal is the following:

- Add `cargo contract check` to check whether a contract successfully compiles to Wasm similar to `cargo check`.
- Make `cargo contract build` generate the `.wasm` and the `metadata` by default and add a flag `--skip-generate-metadata` to avoid generating metadata because it often won't be required. I want to emphasize here again that a `--skip-generate-wasm` does not make sense for this command.

Under the above design, do we want to keep the `cargo contract generate-metadata` command?
Is this what you want to propose?

I am fine with having `cargo contract check` and changing the defaults as described above with `cargo contract build`.

Robbepop:
> > I think it is not helpful to benchmark this for a trivial contract and then claim that it adds 4x overhead. It won't scale with the contract compilation time because it is purely the overhead of running the metadata generation at all (I think).
> > And I think this overhead can be reduced. We should work on that rather than forcing the user to solve the problem for us (deciding when to generate metadata).
> > > I compare cargo contract build mostly with cargo check and cargo contract bundle mostly with a composition of cargo build and/or cargo publish. You need cargo contract build in order to check if your contract compiles successfully on Wasm, you need cargo test to build tests, you basically never need cargo contract generate-metadata and you are going to need cargo contract bundle if you plan to deploy on-chain. Note that cargo contract bundle is comparable to cargo publish because unlike cargo build it will also have some additional checks to verify that no important metadata information is missing for deployment.
> > 
> > 
> > I don't think this is a beginner friendly workflow and something that contradicts rust developer assumptions. The user rightfully can expect that `build` "just builds" a deployable artifact with everything they need without any additional hassle. Why not add a `check` command to cargo contract that does this "quick check" whether it compiles?
> 
> So to summarize, your proposal is the following:
> 
>     * Add `cargo contract check` to check whether a contract successfully compiles to Wasm similar to `cargo check`.
> 
>     * Make `cargo contract build` generate the `.wasm` and the `metadata` by default and add a flag `--skip-generate-metadata` to avoid generating metadata because it often won't be required. I want to emphasize here again that a `--skip-generate-wasm` does not make sense for this command.
> 
> 
> Under the above design, do we want to keep the `cargo contract generate-metadata` command?
> Is this what you want to propose?
> 
> I am fine with having `cargo contract check` and changing the defaults as described above with `cargo contract build`.

Robbepop:
Sorry for accidentally closing ... hate the GH UI, happens too often.

ascjones:
:rofl:

ascjones:
> Under the above design, do we want to keep the cargo contract generate-metadata command?

Remove it - It's redundant because it would be doing the same as `build` (both wasm and metadata).

athei:
> So to summarize, your proposal is the following:

Yes. 

> > Under the above design, do we want to keep the cargo contract generate-metadata command?
> 
> Remove it - It's redundant because it would be doing the same as `build` (both wasm and metadata).

It is already removed in this PR. As a matter of fact my proposal is already implemented here with the exception of the `check` command which basically does what the old `build` command did.

ascjones:
Perhaps we should keep the `generate-metadata` command and add a `deprecated` message since there are still legacy tutorials/presentations etc out there referencing it.

athei:
> Perhaps we should keep the `generate-metadata` command and add a `deprecated` message since there are still legacy tutorials/presentations etc out there referencing it.

I think the deprecated message would suffice, right? No need to still do anything. Remember that people will only read text if they are forced to do so.

Robbepop:
The deprecation message should suffice if we add a note what do use instead now.

athei:
I think that the `cargo contract check` command should **not** copy/optimize the binary to the top level `target` folder. This way we avoid the confusion of having a metadata mismatch when executing `check` but not `build` before uploading.

cmichi:
I updated the PR, you can all take another look :-).

>I think that the cargo contract check command should not copy/optimize the binary to the top level target folder. This way we avoid the confusion of having a metadata mismatch when executing check but not build before uploading.

Good idea! The binary is written to `target` anyway, I appended the `.unoptimized` suffix now and don't output the file name to the user, so it should reduce the possibility to mistake the Wasm for the optimized one. We could also move it to another folder in `target` if you think that makes more sense.

Robbepop:
> I updated the PR, you can all take another look :-).
> 
> > I think that the cargo contract check command should not copy/optimize the binary to the top level target folder. This way we avoid the confusion of having a metadata mismatch when executing check but not build before uploading.
> 
> Good idea! The binary is written to `target` anyway, I appended the `.unoptimized` suffix now and don't output the file name to the user, so it should reduce the possibility of mistaking the Wasm for the optimized one. We could also move it to another folder in `target` if you think that makes more sense.

It would definitely be ideal if the `.wasm` would not end up in the `target/ink` folder or in the top-level of the `target` folder for `cargo contract check`. Is this hard to do?

src/main.rs
```diff
@@ -162,32 +162,38 @@ enum Command {
...
         unstable_options: UnstableOptions,
     },
-    /// Compiles the contract, generates metadata, packs both together in one file
-    #[structopt(name = "pack")]
-    Pack {
-        /// Path to the Cargo.toml of the contract to build and pack
+    /// Command has been deprecated, use 'cargo contract build' instead
+    #[structopt(name = "generate-metadata")]
+    GenerateMetadata {
+        /// Path to the Cargo.toml of the contract to build
```
ascjones, line 185: This command variant can be empty now since none of the fields are used for just the deprecation warning

src/main.rs
```diff
@@ -260,53 +266,74 @@ fn exec(cmd: Command) -> Result<String> {
...
+            if *(skip_bundle) {
+                return Ok(format!(
+                    "\nYour contract's code is ready. You can find it here:\n{}
+                    \nYour contract's metadata is ready. You can find it here:\n{}",
+                    metadata_result.wasm_file.display().to_string().bold(),
+                    metadata_result.metadata_file.display().to_string().bold(),
+                ));
+            }
+            let bundle_result = cmd::metadata::execute(
+                &manifest_path,
```
ascjones, line 302: Please move as much of the logic as possible into `cmd/build.rs`.

Even better see https://github.com/paritytech/cargo-contract/pull/79 where I have extracted command variants into their own struct e.g. https://github.com/paritytech/cargo-contract/blob/5d5f7bee958e403e654ce4f3dda92db4c8757762/src/cmd/extrinsics/call.rs

Goal is to make `main.rs` as slim as possible eventually just dispatching the commands.

ascjones (changes requested):

athei:
> I updated the PR, you can all take another look :-).
> 
> > I think that the cargo contract check command should not copy/optimize the binary to the top level target folder. This way we avoid the confusion of having a metadata mismatch when executing check but not build before uploading.
> 
> Good idea! The binary is written to `target` anyway, I appended the `.unoptimized` suffix now and don't output the file name to the user, so it should reduce the possibility of mistaking the Wasm for the optimized one. We could also move it to another folder in `target` if you think that makes more sense.

Is it possible to not copy it there? I mean cargo will ouput it to the `wasm32-unknown-unknown/release` directory. Just leave it there. Basically what Robin says.

README.md
```diff
@@ -32,8 +32,8 @@ OPTIONS:
 
 SUBCOMMANDS:
     new                  Setup and create a new smart contract project
-    build                Compiles the smart contract
-    generate-metadata    Generate contract metadata artifacts
+    build                Compiles the contract, generates metadata, bundles both together in a '.contract' file
+    check                Check that the Wasm builds; does not optimize, generate metadata, or bundle
```
athei, line 36:
```suggestion
    check                Check that the code builds as wasm; does not output any build artifact to the top level `target/` directory.
```

This statement is of course only true if we prevent copying stuff to the top level `target/` dir on "check".

athei (commented):
I would really like to see that no artifact is outputted to the top level dir. `check` should output nothing from the perspective of the user.

src/main.rs
```diff
@@ -249,35 +258,70 @@ fn exec(cmd: Command) -> Result<String> {
...
-            Ok(format!(
-                "Your metadata file is ready.\nYou can find it here:\n{}",
-                metadata_file.display()
-            ))
+            assert!(maybe_dest_wasm.is_none(), "no dest_wasm should exist");
+            Ok(format!("\nYour contract's code was built successfully."))
         }
+        Command::GenerateMetadata {} => Err(anyhow::anyhow!(format!(
+            "Command deprecated, use `cargo contract build` instead"
+        ))),
```
athei, line 324:
```suggestion
        Command::GenerateMetadata {} => Err(anyhow::anyhow!(
            "Command deprecated, use `cargo contract build` instead"
        )),
```

The anyhow macro supports formatting syntax, already.

cmichi: Nice, thx!

athei:
I checked out the branch and tested the new `build` command with a fresh flipper. Something doesn't seem right because everything is just doubled: Code is generated two times and metadata, too. Also it would be super awesome if on build the metadata generation would integrate itself as an additional "step [4/4]" in the build process. Also it should output that before building metadata-gen as this is what takes the chunk of the time.

```
alexander:~/Downloads/flipper% cargo +nightly contract build
  Generating metadata
 [1/3] Building cargo project
    Finished release [optimized] target(s) in 0.04s
 [2/3] Post processing wasm file
 [3/3] Optimizing wasm file
 Original wasm size: 19.1K, Optimized: 2.5K
    Updating crates.io index
   Compiling metadata-gen v0.1.0 (/var/folders/8g/_rkmq65n0qnf70c41xvp174h0000gn/T/cargo-contract_PWanF3/.ink/metadata_gen)
    Finished release [optimized] target(s) in 1.05s
     Running `target/release/metadata-gen`
  Generating metadata
 [1/3] Building cargo project
    Finished release [optimized] target(s) in 0.04s
 [2/3] Post processing wasm file
 [3/3] Optimizing wasm file
 Original wasm size: 19.1K, Optimized: 2.5K
    Updating crates.io index
   Compiling metadata-gen v0.1.0 (/var/folders/8g/_rkmq65n0qnf70c41xvp174h0000gn/T/cargo-contract_oOz1St/.ink/metadata_gen)
    Finished release [optimized] target(s) in 1.08s
     Running `target/release/metadata-gen`
	
Your contract's code is ready. You can find it here:
/Users/alexander/Downloads/flipper/target/flipper.wasm
                
Your contract's metadata is ready. You can find it here:
/Users/alexander/Downloads/flipper/target/metadata.json
Your contract bundle (code + metadata) is ready. You can find it here:
/Users/alexander/Downloads/flipper/target/flipper.contract
```

ascjones:
> metadata generation would integrate itself as an additional "step [4/4]"

:+1: Good idea

Robbepop:
> > metadata generation would integrate itself as an additional "step [4/4]"
> 
> +1 Good idea

Step [5/5] could be the actual bundling of metadata and `.wasm` in one file and the future checks that are going to be resolved, e.g. missing authors etc.

cmichi:
Alright, updated. Progress output now:
```
$ cargo run -- contract build --manifest-path ../ink/examples/flipper/Cargo.toml 
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/cargo-contract contract build --manifest-path ../ink/examples/flipper/Cargo.toml`
 [1/5] Building cargo project
    Finished release [optimized] target(s) in 0.04s
 [2/5] Post processing wasm file
 [3/5] Optimizing wasm file
 Original wasm size: 19.1K, Optimized: 2.5K
 [4/5] Generating metadata
    Updating crates.io index
   Compiling metadata-gen v0.1.0 (/tmp/cargo-contract_ULeALG/.ink/metadata_gen)
    Finished release [optimized] target(s) in 1.63s
     Running `target/release/metadata-gen`
 [5/5] Generating bundle

Your contract bundle (code + metadata) is ready. You can find it here:
/home/michi/projects/ink/examples/flipper/target/flipper.contract
Your contract's code is ready. You can find it here:
/home/michi/projects/ink/examples/flipper/target/flipper.wasm
Your contract's metadata is ready. You can find it here:
/home/michi/projects/ink/examples/flipper/target/metadata.json
```

src/cmd/build.rs
```diff
@@ -240,19 +250,43 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
...
     manifest_path: &ManifestPath,
     verbosity: Option<Verbosity>,
-    unstable_options: UnstableFlags,
     optimize_contract: bool,
-) -> Result<PathBuf> {
-    let mut crate_metadata = CrateMetadata::collect(manifest_path)?;
-    if !optimize_contract {
-        crate_metadata.dest_wasm.set_extension("wasm.unoptimized");
+    skip_bundle: bool,
+    skip_metadata: bool,
```
ascjones, line 255: I think `false, true` here is an invalid combination (bundling has to create the metadata anyway). 

To make it more readable and prevent invalid states I'd prefer an enum here rather than multiple bool flags e.g.

```
enum BuildArtifacts {
  All, // false, false
  CodeOnly, // true, true
  MetadataOnly, // false, true
}

impl BuildArtifacts {
  fn from_flags(..) -> Result<BuildArtifacts> { ..}
```

Also, at the risk of reopening the bikeshedding, we could consider using equivalent command options e.g. `--bundle` (default) `--code-only` and `--metadata-only`. :see_no_evil:

ascjones (commented):
image

This could be more concise: something like 

```
Your contract artifacts are ready. You can find them in: 
/home/andrew/code/paritytech/ink/examples/flipper/target
  - flipper.contract (code + metadata bundle)
  - flipper.wasm (the contract's code)
  - metadata.json (the contract's metadata)
```

image
 Useful info about contract size is now hidden, would be useful in the summary.

cmichi:
All comments addressed. Output now:
```
$ cargo run -- contract build --manifest-path ../ink/examples/flipper/Cargo.toml                                                                               
   Compiling cargo-contract v0.7.1 (/home/michi/projects/cargo-contract)                                                                                       
    Finished dev [unoptimized + debuginfo] target(s) in 15.97s                                                                                                 
     Running `target/debug/cargo-contract contract build --manifest-path ../ink/examples/flipper/Cargo.toml`                                                   
 [1/5] Building cargo project                                                  
    Finished release [optimized] target(s) in 0.09s                                                                                                            
 [2/5] Post processing wasm file                                               
 [3/5] Optimizing wasm file                                                    
    Updating crates.io index                                                   
   Compiling metadata-gen v0.1.0 (/tmp/cargo-contract_NFmb1e/.ink/metadata_gen)                                                                                
    Finished release [optimized] target(s) in 2.59s                                                                                                            
     Running `target/release/metadata-gen`                                                                                                                     
 [4/5] Generating bundle                                                                                                                                       
 [5/5] Generating metadata                                                     
                                                                                                                                                               
Original wasm size: 19.1K, Optimized: 2.5K                                                                                                                     
                                                                               
Your contract artifacts are ready. You can find them in:                       
/home/michi/projects/ink/examples/flipper/target                               
  - flipper.contract (code + metadata)                                         
  - flipper.wasm (the contract's code)                                                                                                                         
  - metadata.json (the contract's metadata)  
```

ascjones (commented):
Nice, like the new output. Just added a couple of extra comments

src/cmd/build.rs
```diff
@@ -33,6 +34,87 @@ use parity_wasm::elements::{External, MemoryType, Module, Section};
...
+#[derive(Copy, Clone, Eq, PartialEq, Debug, StructOpt)]
+#[structopt(name = "build-artifacts")]
+pub enum BuildArtifacts {
+    /// Generate the Wasm, the metadata and a bundled `<name>.contract` file
+    #[structopt(name = "all")]
+    All,
+    /// Only the Wasm is created, generation of metadata and a bundled `<name>.contract` file is skipped
+    #[structopt(name = "code-only")]
+    CodeOnly,
+    /// Only the Wasm and the metadata are generated, no bundled `<name>.contract` file is created
```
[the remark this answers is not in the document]
cmichi: Yeah I thought about that too, it only works if there is an already existing wasm in the filesystem though. That's what you want to get at, right?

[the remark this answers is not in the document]
ascjones: It is not required, currently. It's just there as a nice to have. In the future the client tools could use that (if present) to verify the metadata matches the deployed code.

So I would say that if `metadata-only is specified then that field should be omitted, even if the wasm is already there because it is not guaranteed to match.

src/cmd/metadata.rs
```diff
@@ -63,14 +61,19 @@ impl GenerateMetadataCommand {
...
         // build the extended contract project metadata
-        let (dest_wasm, source_meta, contract_meta, user_meta, optimization_result) =
-            self.extended_metadata()?;
+        let ExtendedMetadataResult {
+            dest_wasm,
+            source,
+            contract,
+            user,
+            optimization_result,
+        } = self.extended_metadata()?;
```
ascjones, line 73: My preference would be to use the result direactly to access the fields e.g. `result.dest_wasm`, instead of destructuring it. I can see that it has benefits in the construction of the `GenerationResult` below. That's only my preference though so if you prefer this then :+1:

src/cmd/metadata.rs
```diff
@@ -107,13 +168,29 @@ impl GenerateMetadataCommand {
...
         let license = contract_package.license.clone();
-        let hash = self.wasm_hash()?;
-
+        //{
+        let (dest_wasm, hash, optimization_result) =
+            if self.build_artifact != GenerateArtifacts::MetadataOnly {
+                let (wasm, hash, optimization) = self.wasm_hash()?;
+                (Some(wasm), Some(hash), Some(optimization))
+            } else {
+                (None, None, None)
```
ascjones, line 177: Since this is either all `None` or all `Some`, this indicates they should be grouped together.

src/cmd/metadata.rs
cmichi: Okay, I'll leave it as is then, since it keeps the subsequent code shorter and has the benefits you mentioned.

metadata/lib.rs
```diff
@@ -433,6 +436,14 @@ impl ContractBuilder {
     }
 }
 
+fn serialize_as_byte_str_foo<S>(bytes: &Option<[u8; 32]>, serializer: S) -> Result<S::Ok, S::Error>
```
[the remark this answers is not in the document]
cmichi: That's a good suggestion, I migrated to the wrapper type. Imho it has the advantage of having a semantic meaning as well, opposed to the kinda generic `H256`.

src/cmd/metadata.rs
cmichi: So I agree, but thing is that in the subsequent code they are each used in a different context which requires them each as an `Option`. So imho this a short way of assigning the missing ones `None` here, while wrapping the others in `Some`. Otherwise this would need to be done in the subsequent code.

ascjones (approved):
LGTM. I'd like either @Robbepop or @athei to give it a :heavy_check_mark: before merging

README.md
```diff
@@ -32,8 +32,8 @@ OPTIONS:
 
 SUBCOMMANDS:
     new                  Setup and create a new smart contract project
-    build                Compiles the smart contract
-    generate-metadata    Generate contract metadata artifacts
+    build                Compiles the contract, generates metadata, bundles both together in a '.contract' file
+    check                Check that the code builds as Wasm; does not output any build artifact to the top level `target/` directory
```
Robbepop, line 36: Maybe we should mention that this step also deploys (or will deploy) some custom checks that we implement on top besides just making sure that things compile for Wasm.

metadata/lib.rs
```diff
@@ -92,27 +93,84 @@ impl ContractMetadata {
...
+        serialize_as_byte_str(&self.0[..], serializer)
+    }
 }
 
 #[derive(Debug, Serialize)]
 pub struct Source {
-    #[serde(serialize_with = "serialize_as_byte_str")]
-    hash: [u8; 32],
+    #[serde(skip_serializing_if = "Option::is_none")]
+    hash: Option<CodeHash>,
```
Robbepop, line 118: When is this ever `None`? I guess it is always useful even if we provide the sources.

athei:
I updated the branch and built a fresh flipper contract. The output seems to be wrong again:

The "Generating metadata" is at the wrong position. It should be directly behind step `[3/5]`.


```sh
alexander:~% cargo contract build
 [1/5] Building cargo project
   ...
   Compiling flipper v0.1.0 (/var/folders/8g/_rkmq65n0qnf70c41xvp174h0000gn/T/cargo-contract_qOrMhF)
    Finished release [optimized] target(s) in 23.10s
 [2/5] Post processing wasm file
 [3/5] Optimizing wasm file
   ... 
   Compiling flipper v0.1.0 (/var/folders/8g/_rkmq65n0qnf70c41xvp174h0000gn/T/cargo-contract_Qv7HEC)
   Compiling metadata-gen v0.1.0 (/var/folders/8g/_rkmq65n0qnf70c41xvp174h0000gn/T/cargo-contract_Qv7HEC/.ink/metadata_gen)
    Finished release [optimized] target(s) in 36.29s
     Running `target/release/metadata-gen`
 [4/5] Generating bundle
 [5/5] Generating metadata
	
Original wasm size: 19.1K, Optimized: 2.5K

Your contract artifacts are ready. You can find them in:
/Users/alexander/Developer/parity/contracts/flipper/target

  - flipper.contract (code + metadata)
  - flipper.wasm (the contract's code)
  - metadata.json (the contract's metadata)
```

Robbepop (commented):
WIP: review - will continue later

metadata/lib.rs
cmichi: Discussion happened above, [here](https://github.com/paritytech/cargo-contract/pull/97#discussion_r519683048).

README.md
cmichi: I agree, but would only add it once we actually have some checks on top.

athei:
It is still saying "Optimizing wasm file" when it actually goes on and starts building "metadata-gen":
```
alexander:~/Developer/parity/contracts/flipper% cargo contract build
 [1/5] Building cargo project
    Finished release [optimized] target(s) in 0.04s
 [2/5] Post processing wasm file
 [3/5] Optimizing wasm file
    Updating crates.io index
   Compiling metadata-gen v0.1.0 (/var/folders/8g/_rkmq65n0qnf70c41xvp174h0000gn/T/cargo-contract_LmksAM/.ink/metadata_gen)
    Finished release [optimized] target(s) in 1.37s
     Running `target/release/metadata-gen`
 [4/5] Generating metadata
 [5/5] Generating bundle
	
Original wasm size: 19.1K, Optimized: 2.5K

Your contract artifacts are ready. You can find them in:
/Users/alexander/Developer/parity/contracts/flipper/target

  - flipper.contract (code + metadata)
  - flipper.wasm (the contract's code)
  - metadata.json (the contract's metadata)
```

athei (approved):
LGTM

src/cmd/build.rs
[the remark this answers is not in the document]
Robbepop: When would you ever need `metadata-only` is my first question? Please tell me some use cases because I cannot come up with any that would not involve also building the `.wasm`.

metadata/lib.rs
Robbepop: What I read from the discussion is that we still always want the `hash` field to be available. And also I don't think that we need `--metadata-only`. Please convince me of the opposite by coming up with a useful use case.

cmichi: >What I read from the discussion is that we still always want the hash field to be available.

I understood the discussion in the way that we want to omit the hash:
>It is not required, currently. It's just there as a nice to have. In the future the client tools could use that (if present) to verify the metadata matches the deployed code.
>
>_So I would say that if metadata-only is specified then that field should be omitted_, even if the wasm is already there because it is not guaranteed to match.

PR #99, opened 2020-11-04
Replace xbuild with cargo build-std

ascjones:
The [`build-std`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#build-std) feature of cargo now allows to specify the `panic-immediate-abort` feature. See: https://github.com/rust-osdev/cargo-xbuild#alternative-the-build-std-feature-of-cargo.

So we can use that (unstable) feature instead of depending on `xbuild`

Produces the same size wasm output as with `xbuild` e.g. flipper:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -500,8 +500,7 @@ dependencies = [
  "async-std",
  "binaryen",
  "blake2",
- "cargo-xbuild",
- "cargo_metadata 0.11.4",
+ "cargo_metadata",
  "colored",
  "contract-metadata",
  "env_logger",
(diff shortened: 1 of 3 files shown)


athei:
Regarding the TODO: I think passing hardcoded flags to `cargo` as it is done in this PR is fine. I wouldn't generate/mess with a `.cargo/config` that a user might want to edit manually.

What is not fine is hardcoding `RUSTFLAGS`. As far as I see it we just set the env variable to some hardcoded value. But that should be fixed in the mentioned issue. Not here.

ascjones:
> I wouldn't generate/mess with a .cargo/config that a user might want to edit manually.

The idea would be that it would set the default hardcoded `rustflags` iff they are not specified by the user (like we do with the `[profile.release]` section. And also would ensure (since the env var overrides the value from `.cargo/config`) that a user supplied `RUSTFLAGS` is not ignored.

> But that should be fixed in the mentioned issue. Not here.

Anyway I agree that this should be fixed separately with a little more thought.

athei (approved):

ascjones:
CI currently broken, just waiting for a new image to rebuild before merging

PR #104, opened 2020-11-10
Make `source.hash` non-optional, remove `metadata-only`

cmichi:


src/cmd/metadata.rs
```diff
@@ -226,11 +216,11 @@ impl GenerateMetadataCommand {
         let user = self.crate_metadata.user.clone().map(User::new);
 
         Ok(ExtendedMetadataResult {
-            dest_wasm,
+            dest_wasm: Some(dest_wasm),
```
Robbepop, line 219: Useful to be an `Option`?

cmichi: So this is mostly used for `cargo contract check`, which should never result in a `dest_wasm` in the target directory. So for the `GenerationResult` object `dest_wasm` was made an `Option`. There is also an assert for `is_none()` in the `cargo contract check` logic.

Robbepop: okay cool, thanks!

Robbepop (approved):
LGTM!

ascjones (approved):

PR #105, opened 2020-11-10
Do not prettify JSON for bundle

cmichi:
Makes quite a difference.

`flipper.contract` with `serde_json::to_string_pretty()`: `7.0K`
`flipper.contract` with `serde_json::to_string()`: `6.1K`

### DIFF
--- a/src/cmd/metadata.rs
+++ b/src/cmd/metadata.rs
@@ -110,7 +110,7 @@ impl GenerateMetadataCommand {
                     format!("[{}/{}]", current_progress, self.build_artifact.steps()).bold(),
                     "Generating bundle".bright_green().bold()
                 );
-                let contents = serde_json::to_string_pretty(&metadata)?;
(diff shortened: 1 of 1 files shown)


ascjones:
I'm torn on this because even though it is not supposed to be human readable, I personally like to look at the files. Though I realise I am not a typical user. Yes I know I could use an editor to prettify but I am lazy. So we could add a `--pretty` option but then I don't want to have too many options. Help!

One thing I would like to do is to add metadata inspection features which would print to the console in human readable form: constructors, messages, layout etc.

cmichi:
I like the `--pretty` idea. 

I see it like this: for `metadata.json` we still print pretty atm, but imo the bundled `.contract` files primary purpose is to be used for deployment. I mean the whole thing is basically a UX improvement. So making the file size smaller just seems like a further UX improvement.

ascjones (approved):
Cool didn't see that this was just for the `bundle`, thought maybe it was for both.

PR #109, opened 2020-11-13
Exit with 1 on Err

cmichi:
You can provoke this by e.g. applying this patch to `ink/examples/dns/Cargo.toml`:
```
 [lib]
-name = "dns"
+name = "d ns"
```

and then
```
cargo run -- contract build --manifest-path ../ink/examples/dns/Cargo.toml 
echo $?
```

After the merge we should move the `ink-ci-v0.8.0` tag ahead to this commit.

This is another reason why some of the stages in the ink! CI succeed even though `cargo-contract` fails.


ascjones (approved):
LGTM

PR #115, opened 2020-12-01
Nightly test new CI image

TriplEight:
Put the image name into a var so it can be changed by an upstream pipeline trigger.

Companion PR: https://github.com/paritytech/scripts/pull/238

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,11 +16,7 @@ variables:
   CARGO_HOME:                      "/ci-cache/${CI_PROJECT_NAME}/cargo/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   RUST_LIB_BACKTRACE:              "0"
-
-# Necessary for building binaryen-sys, which is part of the binaryen dependency.
-# Should be reverted once the `cargo-contract` ci image provides a clang version
-# for which building `binaryen-sys` works (paritytech/scripts/#237).
-  CXX:                             "/usr/bin/clang++-8"
+  CI_IMAGE:                        "paritytech/contracts-ci-linux:production"
 
 workflow:
   rules:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -52,7 +48,6 @@ workflow:
     - if: $CI_PIPELINE_SOURCE == "schedule"
     - if: $CI_COMMIT_REF_NAME == "master"
     - if: $CI_COMMIT_REF_NAME == "tags"
-    - if: $CI_COMMIT_BRANCH
```
cmichi, line 55: What does it imply that this variable is removed?

TriplEight: nothing really, it's supposed to launch CI pipelines on every branch which is not desirable, we run the usual pipelines only on PR branches, it's covered by the next line. Besides, branches other than PR ones are not getting mirrored to GitLab.

.gitlab-ci.yml
```diff
@@ -16,11 +16,7 @@ variables:
   CARGO_HOME:                      "/ci-cache/${CI_PROJECT_NAME}/cargo/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   RUST_LIB_BACKTRACE:              "0"
-
-# Necessary for building binaryen-sys, which is part of the binaryen dependency.
-# Should be reverted once the `cargo-contract` ci image provides a clang version
-# for which building `binaryen-sys` works (paritytech/scripts/#237).
-  CXX:                             "/usr/bin/clang++-8"
+  CI_IMAGE:                        "paritytech/contracts-ci-linux:production"
```
ascjones, line 19: Just to make sure I understand, if the `contracts-ci-linux` image changes, this CI is triggered but with the above variable pointing at `staging`. If it succeeds the new image will be published as `production`, if it fails we will get a :x: here?

Would be good to get a comment on this side which documents what is going on for future maintainers.

TriplEight: Yeah, you'll get a failed pipeline here https://gitlab.parity.io/parity/cargo-contract/-/pipelines?page=1&scope=all&ref=master

HDYT is it enough tho link this PR in the comment?

TriplEight: done

ascjones (approved):
LGTM

PR #116, opened 2020-12-02
Refactorings for `build` and `check`

cmichi:
I had it in the back of my head for a while now to have `main.rs` be just the dispatcher for `build` and `check` and have the logic reside in the appropriate modules. This was something that came up in a review a while ago.

There are some more minor refactorings in here, like consistent naming and moving a function to utilities.

Also a small fix is in here: for `check` the output of the steps in the progress (`[{}/{}]`) was one off.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -39,11 +39,6 @@ const MAX_MEMORY_PAGES: u32 = 16;
 /// Executes build of the smart-contract which produces a wasm binary that is ready for deploying.
 ///
 /// It does so by invoking `cargo build` and then post processing the final binary.
-///
-/// # Note
-///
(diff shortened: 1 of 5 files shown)


src/cmd/build.rs
```diff
@@ -24,15 +25,94 @@ use crate::{
...
+    )]
+    build_artifact: BuildArtifacts,
+    #[structopt(flatten)]
+    verbosity: VerbosityFlags,
+    #[structopt(flatten)]
+    unstable_options: UnstableOptions,
+}
+
+impl BuildCommand {
+    /// Executes build of the smart-contract which produces a wasm binary that is ready for deploying.
```
ascjones, line 65: This general comment would be better on the `BuiildCommand` struct

ascjones (approved):
:+1: just needs some comments updating

ascjones (approved):

PR #118, opened 2020-12-08
Set up dependabot

cmichi:
Analog to https://github.com/libp2p/rust-libp2p/pull/1744.

See [the GitHub docs](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-and-disabling-version-updates) for in-depth documentation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "daily"


mxinden (approved):
A frequently asked question is whether Dependabot will try to update all dependencies at once. This is not the case. It will only create pull requests for 5 dependencies at a time.

Dependabot will update both `Cargo.toml` and `Cargo.lock`. Given the checked in `Cargo.lock` file it treats the project as a binary, thus updating not only major (below `1.0` minor) versions, but also minor and patch versions.

ascjones (approved):

PR #122, opened 2020-12-11
Put build artifacts under `target/ink/`

### LINKED ISSUE #92: Build artifacts should be put under `target/ink` directory
Currently `cargo-contract` builds under `target` which can interfere with other tools that do the same. The established common rule for tools like `cargo-contract` is to put every artifact under `target/<name>` and we should do this, too.

cmichi:
Closes #92.

### DIFF
--- a/src/cmd/metadata.rs
+++ b/src/cmd/metadata.rs
@@ -55,8 +55,6 @@ impl GenerateMetadataCommand {
     pub fn exec(&self) -> Result<BuildResult> {
         util::assert_channel()?;
 
-        let cargo_meta = &self.crate_metadata.cargo_meta;
-
         let target_directory = self.crate_metadata.target_directory.clone();
         let out_path_metadata = target_directory.join(METADATA_FILE);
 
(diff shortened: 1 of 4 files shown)


src/crate_metadata.rs
```diff
@@ -41,7 +41,8 @@ pub struct CrateMetadata {
 impl CrateMetadata {
     /// Parses the contract manifest and returns relevant metadata.
     pub fn collect(manifest_path: &ManifestPath) -> Result<Self> {
-        let (metadata, root_package) = get_cargo_metadata(manifest_path)?;
+        let (mut metadata, root_package) = get_cargo_metadata(manifest_path)?;
+        metadata.target_directory.push("ink");
```
ascjones, line 45: It would be better to introduce a new field directly on our metadata struct for this rather than modifying the original `cargo` metadata struct. Current usages would use that field (or a getter) instead of `&crate_metadata.cargo_meta.target_directory`

src/cmd/metadata.rs
```diff
@@ -56,12 +56,12 @@ impl GenerateMetadataCommand {
         util::assert_channel()?;
 
         let cargo_meta = &self.crate_metadata.cargo_meta;
```
ascjones, line 58: This binding is now redundant, can just move to usage site now (`Workspace::new`). A further refactoring might be to just have `Workspace::new` accept `&CrateMetadata`.

ascjones (approved):
LGTM

PR #124, opened 2020-12-14
Do not generate build artifacts under `target` for `check`

cmichi:
Our documentation for `cargo contract check`:
```
/// Check that the code builds as Wasm;
/// does not output any build artifact to the top level `target/` directory
```

I found that despite this comment we actually output to the `target` dir. This PR changes this behavior by making `check` build into a tmp dir.

### DIFF
--- a/src/crate_metadata.rs
+++ b/src/crate_metadata.rs
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
 
-use crate::{BuildArtifacts, ManifestPath};
+use crate::ManifestPath;
 use anyhow::{Context, Result};
 use cargo_metadata::{Metadata as CargoMetadata, MetadataCommand, Package};
 use semver::Version;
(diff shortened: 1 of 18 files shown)


src/crate_metadata.rs
```diff
@@ -41,11 +41,22 @@ pub struct CrateMetadata {
 
 impl CrateMetadata {
     /// Parses the contract manifest and returns relevant metadata.
-    pub fn collect(manifest_path: &ManifestPath) -> Result<Self> {
+    pub fn collect(
+        manifest_path: &ManifestPath,
+        build_artifact: Option<BuildArtifacts>,
```
ascjones, line 46: Rather than passing in the `BuildArtifact`, I think instead this should be `target_directory: Option<PathBuf>`, and then move the block with `CheckOnly` and passing `Some(tmp_dir)` into `cmd/build.rs`, otherwise `None`.

All other call sites will pass `None`. This moves the logic to earlier in the code path and also means that `CrateMetadata` doesn't need to know about `BuildArtifacts`

ascjones (commented):
Rather than building into a tmp dir, did you consider instead actually running `cargo check` instead of `cargo build` when invoking cargo for wasm and metadata gen?

src/crate_metadata.rs
cmichi: Ok, so in general I agree about shifting that logic, but this function creates `CrateMetadata`, which needs to contain the path to `original_wasm` and `dest_wasm`, which both have to contain `target_directory`. So it should be `target_directory: PathBuf`, right?

ascjones: I think I meant that if you pass `Some(target_dir)` it uses that otherwise it still uses `metadata.target_directory`, something like:

```rust
let target_directory = target_directory.unwrap_or_else(|| 
    metadata.target_directory.clone().as_path().join("ink")
);
```

cmichi:
@ascjones Thanks for the comments, I applied all of them. Can you take another look please?

src/crate_metadata.rs
```diff
@@ -43,9 +43,7 @@ impl CrateMetadata {
     /// Parses the contract manifest and returns relevant metadata.
     pub fn collect(manifest_path: &ManifestPath) -> Result<Self> {
         let (metadata, root_package) = get_cargo_metadata(manifest_path)?;
-
-        let mut target_directory = metadata.target_directory.clone();
-        target_directory.push("ink");
+        let target_directory = metadata.target_directory.clone().as_path().join("ink");
```
ascjones, line 46: Is the `clone()` redundant now?

ascjones (approved):
LGTM

src/crate_metadata.rs
cmichi: Good catch!

PR #131, opened 2020-12-29
Enable debug info with flag in command build

clearloop:
## Desc

Currently `cargo-contract` trims `name section` by default while building contracts, `strip_custom_sections`, `wasm-opt` and an operation in `parity-wasm`(https://github.com/paritytech/parity-wasm/pull/300) will remove `name section`, which makes it hard to debug contracts.

We are trying to set up a flag `--debug` to keep `name section` in this PR, we have kept the `name section` after `strip_custom_sections`  but not in `wasm_opt`(not sure if it is a bug,  `name section` optimized even passing `debug_info: true` to `binaryen`, but `wasm-opt` works fine using the same arguments in command-line `wasm-opt -O3 -s 1 --debuginfo contract.wasm`)

Instead, we left `target/ink/<contract>.src.wasm` after the process, which contains name section.

## Refs

#169

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -132,15 +132,18 @@ fn build_cargo_project(
     crate_metadata: &CrateMetadata,
     verbosity: Option<Verbosity>,
     unstable_flags: UnstableFlags,
+    debug: bool,
 ) -> Result<()> {
     util::assert_channel()?;
 
     // set linker args via RUSTFLAGS.
     // Currently will override user defined RUSTFLAGS from .cargo/config. See https://github.com/paritytech/cargo-contract/issues/98.
-    std::env::set_var(
-        "RUSTFLAGS",
-        "-C link-arg=-z -C link-arg=stack-size=65536 -C link-arg=--import-memory",
-    );
+    let mut flags =
+        "-C link-arg=-z -C link-arg=stack-size=65536 -C link-arg=--import-memory".to_string();
+    if debug {
+        flags.push_str(" -C opt-level=1");
+    }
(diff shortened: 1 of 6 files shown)


Robbepop:
I think as long as it is still guaranteed that non-`debug` builds behaves as before this is fine and could be a valuable addition.

src/cmd/build.rs
```diff
@@ -127,15 +132,18 @@ fn build_cargo_project(
...
-    std::env::set_var(
-        "RUSTFLAGS",
-        "-C link-arg=-z -C link-arg=stack-size=65536 -C link-arg=--import-memory",
-    );
+    let mut flags =
+        "-C link-arg=-z -C link-arg=stack-size=65536 -C link-arg=--import-memory".to_string();
+    if debug {
+        flags.push_str(" -C opt-level=1");
+    }
+    std::env::set_var("RUSTFLAGS", flags);
```
clearloop, line 146: Not sure using the rustc flag or building debug wasm binary under `--debug`

src/cmd/build.rs
```diff
@@ -339,30 +375,32 @@ pub(crate) fn execute_with_crate_metadata(
     optimize_contract: bool,
     build_artifact: BuildArtifacts,
     unstable_flags: UnstableFlags,
-) -> Result<(Option<PathBuf>, Option<OptimizationResult>)> {
+    debug: bool,
+) -> Result<(Option<PathBuf>, Option<PathBuf>, Option<OptimizationResult>)> {
```
clearloop, line 379: This modification is quite dirty, would mind offering some advice if this PR is welcome in the next days? @Robbepop

PR #132, opened 2021-01-05
delete cmd::new:: in new.rs

honeywest:
Unified function call
https://github.com/paritytech/cargo-contract/blob/4c156d98176ca767c2cc62cab03d091297ca4e88/src/cmd/new.rs#L121-L122


ascjones (approved):
LGTM

PR #133, opened 2021-01-06
add build quiet and verbose usage

honeywest:
now:
FLAGS:
    -h, --help       Prints help information
        --quiet      No output printed to stdout
    -V, --version    Prints version information
        --verbose    Use verbose output

before:
FLAGS:
    -h, --help       Prints help information
        --quiet      
    -V, --version    Prints version information
        --verbose

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -94,8 +94,10 @@ impl ExtrinsicOpts {
 
 #[derive(Clone, Debug, StructOpt)]
 pub struct VerbosityFlags {
+    /// No output printed to stdout
     #[structopt(long)]
     quiet: bool,
+    /// Use verbose output
     #[structopt(long)]
     verbose: bool,
 }


src/main.rs
```diff
@@ -94,8 +94,10 @@ impl ExtrinsicOpts {
 
 #[derive(Clone, Debug, StructOpt)]
 pub struct VerbosityFlags {
+    /// No output printed to stdout
```
ascjones, line 97: This is not actually true, it just passes `--quiet` to the underlying `cargo` invocation. We would need to add conditionals around all our `println!`s for it to be quiet in this sense.

honeywest: Some(Verbosity::Quiet) => cmd.arg("--quiet"),
Some(Verbosity::Verbose) => cmd.arg("--verbose"),
I see. I just copy cargo --quiet.
What do you think it should be?

honeywest: Cargo No output printed to stdout.   How about this one?

ascjones: What about `No cargo output printed to stdout`.

Also opened an issue about making it truly quiet: https://github.com/paritytech/cargo-contract/issues/134

cmichi:
Superseded by #165. 

Thanks for bringing this up, @honeywest!

PR #165, opened 2021-01-29
Suppress output on `--quiet`

### LINKED ISSUE #134: Quiet verbosity effects
Currently the `--quiet` flag is just passed through to the underlying `cargo` invocation (see https://github.com/paritytech/cargo-contract/pull/133#discussion_r552469809).

We should consider making `--quiet` not print anything to `stdout` as it does in cargo, whether by wrapping all our `println!`s in conditionals and/or investigate whether we should print progress updates to `stderr` as I believe `cargo` does (needs confirming).

cmichi:
Closes #134.

Supersedes #133.

From the issue:
>[…] investigate whether we should print progress updates to stderr as I believe cargo does (needs confirming).

I investigated and the `cargo` behavior is that progress output is suppressed entirely for `--quiet`, but errors are still printed to `stderr`. This is the behavior which I implemented in this PR.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -157,7 +157,7 @@ fn build_cargo_project(
         Ok(())
     };
 
-    if unstable_flags.original_manifest {
+    if unstable_flags.original_manifest && util::is_verbose(&verbosity) {
         println!(
             "{} {}",
             "warning:".yellow().bold(),
@@ -318,6 +318,7 @@ fn execute(
             target_directory: crate_metadata.target_directory,
             optimization_result: maybe_optimization_result,
             build_artifact,
+            verbosity,
         };
         return Ok(res);
     }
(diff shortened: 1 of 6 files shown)


src/cmd/build.rs
```diff
@@ -342,26 +343,32 @@ pub(crate) fn execute_with_crate_metadata(
...
-        format!("[1/{}]", build_artifact.steps()).bold(),
-        "Building cargo project".bright_green().bold()
-    );
+    if util::is_verbose(&verbosity) {
+        println!(
+            " {} {}",
+            format!("[1/{}]", build_artifact.steps()).bold(),
+            "Building cargo project".bright_green().bold()
+        );
+    }
```
ascjones, line 352: This could be a macro in place of `println!`

src/main.rs
```diff
@@ -356,20 +367,30 @@ fn main() {
...
 }
 
-fn exec(cmd: Command) -> Result<String> {
+fn exec(cmd: Command) -> Result<Option<String>> {
     match &cmd {
         Command::New { name, target_dir } => cmd::new::execute(name, target_dir.as_ref()),
         Command::Build(build) => {
             let result = build.exec()?;
-            Ok(result.display())
+            if util::is_verbose(&result.verbosity) {
```
ascjones, line 375: Not nice to have to have these conditional blocks everywhere.

We could either replace the success status message propogation, and just call our `println!` ersatz macro directly like we would do in other places, or rework the args so the verbosity flags are available at the top level command.

src/util.rs
```diff
@@ -93,6 +93,15 @@ pub(crate) fn base_name(path: &PathBuf) -> &str {
         .expect("must be valid utf-8")
 }
 
+/// Returns `true` if output should be printed (i.e. verbose output is set).
+pub(crate) fn is_verbose(verbosity: &Option<Verbosity>) -> bool {
```
ascjones, line 97: If we make `Verbosity` a three variant enum e.g. `Default, Quiet, Verbose`, we can replace `Option<Verbosity>` and make `is_verbose` a method.

ascjones (approved):
LGTM!

PR #166, opened 2021-01-31
add rust-toolchain

honeywest:


ascjones (commented):
`cargo-contract` itself can build with `stable`, `nightly` is only required when using it to build contracts

PR #168, opened 2021-02-03
Use either `binaryen-rs` dep or `wasm-opt` binary

### LINKED ISSUE #113: binaryen dependency feature
In https://github.com/paritytech/cargo-contract/pull/95 we add the `binaryen` as a dependency, rather than requiring manual installation. However this 

- Increases the build time which we want to keep as fast and easy as possible. (*edit*: it might not be by much? need to confirm with fresh build)
- Takes up quite a lot of disk space 
- Requires the prerequisities to be installed. 

We could still provide this but behind a feature, and default to using an installed binary, and perhaps we could provide better information to be able to download a prebuilt binary e.g. there is one here https://packages.debian.org/buster/binaryen.

Potentially we could detect the presence of the binary on the `PATH`, and then choose whether to enable this feature but it would require some trickery in `build.rs` so not sure about that.

/cc @Robbepop

cmichi:
Closes #113.

Soo, this will get a bit hairy and I expect discussion in this PR.

Our initial idea of using our build script to dynamically enable the `binaryen-rs` dependency based on availability of the `wasm-opt` binary won't work:

>The build script is run (once) after the dependencies are built, so without time travel it can not add dependencies.

(from https://github.com/rust-lang/cargo/issues/5499)

What we could do instead is to install `cargo-contract` either with
* `cargo install --features wasm-opt-unavailable`, if we want to use the `binaryen-rs` dep.
* or via the usual `cargo install`, if the `wasm-opt` binary is available.

This is currently implemented by this PR ‒ if the feature `wasm-opt-unavailable` is enabled the `binaryen-rs` dep is built and used for optimization.

To automate the decision which installation parameters to use we could e.g. build something like the Substrate installation script  (`curl https://getsubstrate.io -sSf | bash`). I'm also not super-happy with this approach, but it's the best workaround I've found so far. Open for other ideas, of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,14 @@ A CLI tool for helping setting up and managing WebAssembly smart contracts writt
 
 `rust-src` is a prerequisite: `rustup component add rust-src`.
 
-We optimize the resulting contract Wasm using `binaryen`. You have two options of how
-to invoke `binaryen`:
+We optimize the resulting contract Wasm using `binaryen`. You have two options for installing it:
 
-  - The preferred way:
+  - _The preferred way:_  
     Install [`binaryen`](https://github.com/WebAssembly/binaryen#tools). Many package managers
     have it available nowadays (it's a package for e.g. Debian/Ubuntu, Homebrew, Arch Linux, etc.).
-    After you've installed `wasm-opt` execute `cargo install --force cargo-contract`.
+    After you've installed the package execute `cargo install --force cargo-contract`.
 
-  - Build `binaryen` as a dependency when installing `cargo-contract`:
+  - _Build `binaryen` as a dependency when installing `cargo-contract`:_  
     A C++14 compiler and python >= 3.5 is required.
     Execute `cargo install --force --features wasm-opt-unavailable cargo-contract`.
 
--- a/README.md
+++ b/README.md
@@ -8,15 +8,19 @@ A CLI tool for helping setting up and managing WebAssembly smart contracts writt
 
 ## Installation
 
-- **Prerequisites**
+`rust-src` is a prerequisite: `rustup component add rust-src`.
 
-  - **rust-src**: `rustup component add rust-src`
-  - A C++14 compiler and python >= 3.5 is required for building the
-    [binaryen](https://github.com/WebAssembly/binaryen) dependency.
-    `binaryen` is built automatically during the `cargo-contract` build process.
+We optimize the resulting contract Wasm using `binaryen`. You have two options of how
+to invoke `binaryen`:
 
-- **Install latest version from [crates.io](https://crates.io/crates/cargo-contract)**
-  - `cargo install cargo-contract`
+  - The preferred way:
+    Install [`binaryen`](https://github.com/WebAssembly/binaryen#tools). Many package managers
+    have it available nowadays (it's a package for e.g. Debian/Ubuntu, Homebrew, Arch Linux, etc.).
+    After you've installed `wasm-opt` execute `cargo install --force cargo-contract`.
+
+  - Build `binaryen` as a dependency when installing `cargo-contract`:
+    A C++14 compiler and python >= 3.5 is required.
+    Execute `cargo install --force --features wasm-opt-unavailable cargo-contract`.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ We optimize the resulting contract Wasm using `binaryen`. You have two options f
 
   - _The preferred way:_  
     Install [`binaryen`](https://github.com/WebAssembly/binaryen#tools). Many package managers
-    have it available nowadays (it's a package for e.g. Debian/Ubuntu, Homebrew, Arch Linux, etc.).
+    have it available nowadays ‒ e.g. it's a package for [Debian/Ubuntu](https://tracker.debian.org/pkg/binaryen),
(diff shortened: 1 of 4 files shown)


src/cmd/build.rs
```diff
@@ -293,6 +284,63 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<OptimizationResult> {
     })
 }
 
+/// Optimizes the Wasm supplied as `wasm` using the `binaryen-rs` dependency.
+///
+/// If successful, the optimized Wasm is returned as a `Vec<u8>`.
+#[cfg(feature = "wasm-opt-unavailable")]
```
Robbepop, line 290: I would rename the crate feature to something "positive" such as `wasm-opt-installed` and then negate these conditions via: `#[cfg(not(feature = "wasm-opt-installed"))]`

src/cmd/build.rs
```diff
@@ -293,6 +284,63 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<OptimizationResult> {
...
+    Ok(module.write())
+}
+
+/// Optimizes the Wasm supplied as `crate_metadata.dest_wasm` using
+/// the `wasm-opt` binary.
+///
+/// If successful, the optimized Wasm file is created under `optimized`
+/// and returned as a `Vec<u8>`.
+#[cfg(not(feature = "wasm-opt-unavailable"))]
+fn get_optimized_wasm(
```
ascjones, line 312: Maybe `do_optimization` or similar, `get` gives the impression that it is just accessing some existing resource rather than performing a process.

ascjones (commented):
I'm happy with this approach, it appears to be the best we can do. I would like to see this clearly documented in the README to avoid any confusion (though no doubt there will still be plenty)

src/cmd/build.rs
```diff
@@ -293,6 +284,63 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<OptimizationResult> {
...
+            "wasm-opt is not installed. Install this tool on your system in order to \n\
+             reduce the size of your contract's Wasm binary. \n\
+             See https://github.com/WebAssembly/binaryen#tools"
+                .bright_yellow()
+        );
+    }
+
+    let output = Command::new("wasm-opt")
+        .arg(crate_metadata.dest_wasm.as_os_str())
+        .arg("-O3") // execute -O3 optimization passes (spends potentially a lot of time optimizing)
```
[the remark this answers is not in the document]
cmichi: Good point, I also created a follow-up for giving contract authors control over it: https://github.com/paritytech/cargo-contract/issues/169.

cmichi:
I updated the readme.

[Click here for the preview](https://github.com/paritytech/cargo-contract/blob/cmichi-build-binaryen-only-as-fallback/README.md)

src/cmd/build.rs
```diff
@@ -268,7 +268,7 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<OptimizationResult> {
     let mut dest_wasm_file_content = Vec::new();
     dest_wasm_file.read_to_end(&mut dest_wasm_file_content)?;
 
-    let optimized_wasm = get_optimized_wasm(crate_metadata, &optimized, &dest_wasm_file_content)?;
+    let optimized_wasm = do_optimization(crate_metadata, &optimized, &dest_wasm_file_content, 3)?;
```
ascjones, line 271: Would be good to make a type here to represent the valid opt levels, but it doesn't necessarily have to be part of this PR

```
 -O                                           execute default optimization 
                                                passes 
   -O0                                          execute no optimization passes 
   -O1                                          execute -O1 optimization passes
                                                (quick&useful opts, useful for 
                                                iteration builds) 
   -O2                                          execute -O2 optimization passes
                                                (most opts, generally gets most
                                                perf) 
   -O3                                          execute -O3 optimization passes
                                                (spends potentially a lot of 
                                                time optimizing) 
   -O4                                          execute -O4 optimization passes
                                                (also flatten the IR, which can
                                                take a lot more time and memory,
                                                but is useful on more nested / 
                                                complex / less-optimized input)
   -Os                                          execute default optimization 
                                                passes, focusing on code size 
   -Oz                                          execute default optimization 
                                                passes, super-focusing on code 
                                                size 
```

ascjones (approved):
LGTM. It is possible to supply invalid opt levels, but since this is hardcoded for now I think that is okay.

src/cmd/build.rs
cmichi: Good point, I plan on doing the follow-up of exposing these via the cli anyway, so I'll include it then.

README.md
```diff
@@ -8,15 +8,19 @@ A CLI tool for helping setting up and managing WebAssembly smart contracts writt
...
-    `binaryen` is built automatically during the `cargo-contract` build process.
+We optimize the resulting contract Wasm using `binaryen`. You have two options for installing it:
 
-- **Install latest version from [crates.io](https://crates.io/crates/cargo-contract)**
-  - `cargo install cargo-contract`
+  - _The preferred way:_  
+    Install [`binaryen`](https://github.com/WebAssembly/binaryen#tools). Many package managers
+    have it available nowadays ‒ e.g. it's a package for [Debian/Ubuntu](https://tracker.debian.org/pkg/binaryen),
+    [Homebrew](https://formulae.brew.sh/formula/binaryen) and [Arch Linux](https://archlinux.org/packages/community/x86_64/binaryen/).
+    After you've installed the package execute `cargo install --force cargo-contract`.
```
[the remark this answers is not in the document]
cmichi: Yep, I thought that too.

PR #171, opened 2021-02-05
Implement Wasm validation for known issues/markers

### LINKED ISSUE #163: Ensure building/checking depends on Wasm import functions being correct
We want to examine the resulting contract Wasm in order to ensure that certain criteria are met.

# ToDo

- [ ] Ensure that building/checking happens only if the following conditions are fulfilled.
- [ ] Ensure that only `seal_*` functions are imported in the resulting contract Wasm.
- [ ] For functions which are not `seal_*` we want to give helpful messages:
    - [ ] Recommend `overflow-checks = false` when [this bug](https://github.com/rust-lang/rust/issues/78744) occurs.
    - [ ] If generated ink! constructs [like this](https://github.com/paritytech/ink/blob/master/crates/lang/codegen/src/generator/cross_calling.rs#L209) are found.

cmichi:
Closes #163.

Output currently looks like this:
```
…
 [2/3] Post processing wasm file                                                                                                                               
ERROR: Validation of the Wasm failed.                                          
                                                                               
                                                                               
ERROR: An error was found while compiling the contract:                       
The ink! message with the selector `0xAD931DAA` contains an invalid trait call. 
                                                                               
Please check if the mutable parameter of the function to call is consistent                                                                                    
with the scope in which it is called.                                                                                                                          
                                                                               
                                                                                                                                                               
ERROR: An unexpected panic function import was found in the contract Wasm.                                                                                     
This typically goes back to a known bug in the Rust compiler:                                                                                                  
https://github.com/rust-lang/rust/issues/78744

As a workaround try to insert `overflow-checks = false` into your `Cargo.toml`. 
This will disable safe math operations, but unfortunately we are currently not  
aware of a better workaround until the bug in the compiler is fixed.
```


src/cmd/build.rs
```diff
@@ -249,10 +257,74 @@ fn post_process_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
...
     parity_wasm::serialize_to_file(&crate_metadata.dest_wasm, module)?;
     Ok(())
 }
 
+/// Validates the import section in the Wasm.
+///
+/// The checks currently fall into two categories:
+/// - Known bugs for which we want to recommend a solution.
+/// - Markers inserted by the ink! codegen for errors which can't be checked at compile time.
+fn validate_import_section(module: &Module) -> Result<()> {
```
cmichi, line 271: I guess the main thing to discuss is if we want to have this validation in a more structured manner (e.g. as a separate module with separate functions for each validation criteria). For now there are only these three checks, but we already have some ideas of stuff to add.

Robbepop: Generally the scheme is to have all of ink!'s generated errors have the same form which `cargo-contract` can then query for.
Currently the form is: `__ink_enforce_error_for_{0}_{1}`
Where `{0}` is the subject, which currently is only either message or construct but which will potentially have more subjects to come in the future and the `{1}` is additional context for the error message.
Ideally we would tune the additional error context information for our current errors to encompass much more information than what they currently do - which is just the selector of the message. Ideally the additional error context was better structured like so:
`__ink_enforced_error_{0}_{2..n}`
Where `{0}` still is an error code known to `cargo-contract`. Every error code required an additional `N` number of "parameters" that are given by `_` separated additional values. Then `cargo-contract` could construct a fully qualified error message to the user.
An example for this was:
```rust
// The below error code represents calling a `&mut self` message in a context that
// only allows for `&self` messages. This may happen under certain circumstances
// when ink! trait implementations are involved with long-hand calling notation.
//
// The error code requires 2 parameters:
// 1. The name of the called message.
// 2. The selector of the called message.
// 3. The trait that defines the called message.
const CALL_MUT_MESSAGE_INVALID_CTX_ERRC: u32 = 0x1;

fn __ink_enforced_error_1_foo_0xc0decafe_FooTrait
```
Then `1` refers to `CALL_MUT_MESSAGE_INVALID_CTX_ERRC`, `foo` is the message's name, `0xc0decafe` is the message's selector and `FooTrait` is the name of the trait that defines the function.
What we need to do is to articulate a proper list between ink! and `cargo-contract` for all the valid error codes and the parameters that `cargo-contract` can expect for them.

cmichi: That's great.

>What we need to do is to articulate a proper list between ink! and cargo-contract for all the valid error codes and the parameters that cargo-contract can expect for them.

Hmm, how about creating a crate in ink! for this and sharing it as a dependency?

ascjones: Perhaps it could go in the existing `metadata` crate? It could be a separate "static" metadata module.

Robbepop: > That's great.
> 
> > What we need to do is to articulate a proper list between ink! and cargo-contract for all the valid error codes and the parameters that cargo-contract can expect for them.
> 
> Hmm, how about creating a crate in ink! for this and sharing it as a dependency?

How do you imagine this to work out exactly?
How are the structures encoding the errors, their required parameters and their display messages?

Robbepop: > Perhaps it could go in the existing `metadata` crate? It could be a separate "static" metadata module.

Not sure metadata fits the use case. Can you elaborate?

ascjones (commented):
It would be possible to make it forward compatible if we wanted by adding these to an extra section in the generated metadata, though this would of course rely on the metadata being generated first.

Robbepop:
> It would be possible to make it forward compatible if we wanted by adding these to an extra section in the generated metadata, though this would of course rely on the metadata being generated first.

This protocol should not exist in the contract metadata since it has absolutely no value beyond `cargo-contract`. If it escapes `cargo-contract`'s build it is already bad.

src/cmd/build.rs
ascjones: Just thinking out loud really, I initially thought of it because `metadata` is a crate which is already shared by `cargo-contract`. It does already contain some information about the environment so not too much of a stretch to imagine adding these error mappings.

ascjones:
That makes sense.

There is still a way at least to provide some forward compatibility by detecting the common prefix `__ink_enforce_error` and enumerating the subsequent arguments even if we wouldn't be able to provide a nice error message.

src/cmd/build.rs
ascjones: > I guess the main thing to discuss is if we want to have this validation in a more structured manner (e.g. as a separate module with separate functions for each validation criteria). For now there are only these three checks, but we already have some ideas for stuff to add.

If we know we will add more checks then I think we should just have them defined in a separate module at least to avoid cluttering up this file.

Robbepop (approved):
LGTM

nice! :rocket:

src/validation/validate_wasm.rs
```diff
@@ -0,0 +1,237 @@
...
+pub fn validate_import_section(module: &Module) -> Result<()> {
+    let imports = module
+        .import_section()
+        .expect("import section must exist")
+        .entries()
+        .iter();
+    let original_imports_len = imports.len();
+    let mut errs = Vec::new();
+
+    let filtered_imports = imports.filter(|section| {
```
ascjones, line 79: This whole block would be (for me) more readable if it was either a pure loop which pushed to `errs` and increments `import_len` or purely functional by aggregating the errors. Something like this *might* work

```rust
let result = 
        imports
            .map(|import| {
                let field = section.field();
                if field.contains("panic") {
                    Err(anyhow::anyhow!(String::from(
                        "An unexpected panic function import was found in the contract Wasm.\n\
                        This typically goes back to a known bug in the Rust compiler:\n\
                        https://github.com/rust-lang/rust/issues/78744\n\n\
                        As a workaround try to insert `overflow-checks = false` into your `Cargo.toml`.\n\
                        This will disable safe math operations, but unfortunately we are currently not \n\
                        aware of a better workaround until the bug in the compiler is fixed.",
                    )))
                } else if field.starts_with(INK_ENFORCE_ERR) {
                    Err(anyhow::anyhow!(parse_linker_error(field)))
                } else if !is_import_allowed(field) {
                    Err(anyhow::anyhow!("TODO import not allowed"))
                } else {
                    Ok(())
                }
            })
            .collect::<Result<(), Vec<String>>>();
    
    result.map_err(|errs| {
        anyhow::anyhow!(
            "Validation of the Wasm failed.\n\n\n{}",
            errs.into_iter()
                .map(|err| format!("{} {}", "ERROR:".to_string().bold(), err))
                .collect::<Vec<String>>()
                .join("\n\n\n")
        )
    })
```

ascjones: Actually on second thoughts this won't work because the Result `FromIterator` only collects for the `Ok` variant.

src/validation/validate_wasm.rs
```diff
@@ -0,0 +1,237 @@
...
+
+/// Validates the import section in the Wasm.
+///
+/// The checks currently fall into two categories:
+/// - Known bugs for which we want to recommend a solution.
+/// - Markers inserted by the ink! codegen for errors which can't be checked at compile time.
+pub fn validate_import_section(module: &Module) -> Result<()> {
+    let imports = module
+        .import_section()
+        .expect("import section must exist")
```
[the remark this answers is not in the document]
cmichi: Come to think about it I think the correct way of handling this is to actually have the validation return `Ok(())` if no import section is available ‒ since there are no invalid imports to be found.

I adapted this PR and added a test for this behavior.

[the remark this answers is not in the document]
ascjones: But shouldn't it import at least `seal_deploy` and `seal_call`, or is that checked somewhere else?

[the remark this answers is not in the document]
Robbepop: `seal_deploy` and `seal_call` _exports_ need to be declared for a valid `contracts-pallet` Wasm smart contract.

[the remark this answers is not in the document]
ascjones: :facepalm:

src/validate_wasm.rs
```diff
@@ -0,0 +1,320 @@
...
+
+/// Extracts the ink! linker error marker from the `field`, parses it, and
+/// returns a human readable error message for it.
+fn parse_linker_error(field: &str) -> String {
+    let encoded = field
+        .strip_prefix(INK_ENFORCE_ERR)
+        .expect("error marker must exist as prefix");
+    let hex = serde_hex::from_hex(&encoded).expect("decoding hex failed");
+    let decoded =
+        <EnforcedErrors as codec::Decode>::decode(&mut &hex[..]).expect("decoding object failed");
```
ascjones, line 151: This should probably still return a friendly string explaining that the `EnforcedError` could not be decoded, and the probable cause (a mismatch between the `ink!` version and the local definition).

ascjones (approved):
LGTM

src/validate_wasm.rs
ascjones: This could occur when e.g. `ink!` adds a new variant, and someone is running an outdated version of `cargo-contract`.

src/validation/validate_wasm.rs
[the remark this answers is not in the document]
cmichi: Hmm, does it make sense then to validate the Wasm for those exports and fail building the contract otherwise?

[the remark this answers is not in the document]
Robbepop: Well, if those exports do not exist the Wasm is also invalid. Should normally not happen if you compile with ink! though.

Issue #60, opened 2020-06-23
Generate metadata for sub-contracts in multi contract workspace


ascjones:
Following on from https://github.com/paritytech/cargo-contract/pull/58, we should support generating of metadata for the non-root contract in a multi contract workspace.

For example, the [`delegator`](https://github.com/paritytech/ink/tree/master/examples/delegator) example hosts multiple contracts which are dependencies of the main contract.

## Proposal
- All metadata files will be written to the `target` directory of the workspace root, with the file name the same as the contract module name e.g. `delegator.json`, `incrementer.json` etc.
- `cargo contract generate-metadata` in the [root workspace dir](https://github.com/paritytech/ink/tree/master/examples/delegator) generates just the metadata for the root contract (in this case the `delegator`
- Sub contracts: e.g `cd adder && cargo contract generate-metadata` generates just the metadata for that sub contract.
- Alternatively `cargo contract generate-metadata -p adder` from the root does the same, mirroring `cargo` functionality.
- `cargo contract generate-metadata --all` will generate metadata files for all contracts in the workspace, e.g. `delegator.json, adder.json, subber.json`

## Implementation notes

- Need to auto remove the `ink-as-dependency` feature in the root workspace manifest for the sub contracts

ii-ii-ii:
hi, I would like to know if this issue is solved. Because I'm working on multiple contracts and I'll have some trouble if I can't generate abi for sub-contracts.

ascjones:
It's not solved yet. It should still work if you run `cargo-contract` from a sub contract's directory, or use the `--manifest-path` argument

ii-ii-ii:
When I try to execute `cargo +nightly contract build` in a subdirectory I get an error:

```
error[E0463]: can't find crate for `contract`
 --> .ink/metadata_gen/main.rs:1:1
  |
1 | extern crate contract;
  | ^^^^^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: could not compile `metadata-gen`

To learn more, run the command again with --verbose.
```

It can get a wasm file, but cannot generate abi. Using `--manifest-path` also gives the same error

ascjones:
My mistake, I thought it would have worked. I'll look at implementing this soon.

PR #174, opened 2021-02-12
Generate metadata explicitly for the contract which is build

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/60.


ascjones:
Nice! I'll have a proper look on Monday.

src/workspace/mod.rs
```diff
@@ -121,10 +123,32 @@ impl Workspace {
...
+        package_path: PathBuf,
+    ) -> Result<&mut Self> {
+        // We strip the workspace root path from the path where the package is found.
+        // This way we have the relative path of the package in the workspace.
+        let relative_package_path = package_path
+            .strip_prefix(self.workspace_root.clone())
+            .expect("package path must be prefixed with workspace path");
+        // We prepend this path since the metadata generation package will be under
+        // `.ink/metadata_gen/` and we need to traverse up from there.
+        let mut path = PathBuf::from("../../");
```
ascjones, line 142: This seems brittle (dependent on the directory structure) , it would be good to encapsulate this somehow in `manifest`.

cmichi: So the reason why I've put this here is that only in this module we have information about the `workspace_root`. In `manifest` we only have information about the local view of a package.

But in order to have our metadata generation crate reference the sub-contract we need to hand the information about the relative path to the sub-contract to the `manifest`. The only other way I see is to add something like `relative_root_path` to `manifest`. Wdyt?

ascjones: > We can't use the absolute path, since we copy the contract to a temp directory.

I think we can make it an absolute path, we have access to the tmp dir path here where we call into `manifest.write` https://github.com/paritytech/cargo-contract/blob/f342f2364f1e7ba8859ca2f1f64bdcd25e203021/src/workspace/mod.rs#L177

src/workspace/mod.rs
```diff
@@ -149,6 +163,15 @@ impl Workspace {
...
             let new_manifest = ManifestPath::new(new_path)?;
 
+            // replace the original path to the package for which metadata should be
+            // generated with the absolute path to the package in the temporary directory.
+            if let Some(package) = manifest.metadata_package_mut() {
+                let absolute_package_path = package.path.canonicalize()?;
+                let mut new_path: PathBuf = target.as_ref().into();
+                new_path.push(absolute_package_path.strip_prefix(&self.workspace_root)?);
+                package.path = new_path;
+            }
```
cmichi, line 173: I've moved this logic here now, since the block above does basically the same thing for the manifest path.

Also we would need to additionally pass `self.workspace_root` and the `target` directory into `manifest` if we would really like to put the logic there.

ascjones: It does remove the hardcoded relative path, but this pattern doesn't feel 100% right because the construction of this path is split across `cmd/metadata.rs` and `workspace/mod.rs`, then finally used in `manifest.rs`. If encapsulating it means passing extra context into `manifest` via fields or method arguments, then it should be considered (if practical).

ascjones (changes requested):
This introduces a regression, whereby the optimized contract is empty. Building `flipper` with

- `master`
image

- this PR
image

Haven't tried debugging it, but I think it's worth adding an assert for this in ine of the existing `build` tests to prevent this regression.

ascjones (approved):
LGTM, this does fix the issue of metadata generation for sub contracts.

However it doesn't completely close all the points in #60, but I have raised a follow up issue for that https://github.com/paritytech/cargo-contract/issues/182.

Also I came across https://github.com/paritytech/cargo-contract/issues/181 while testing, which also needs investigating but possibly as a follow up.

PR #175, opened 2021-02-14
Updated deps to fix extrinsics build

brson:
`cargo-contract` doesn't build with `--features=extrinsics`. This patch makes version bumps that fix the build.

It is mostly updating the lockfile, but also bumping the substrate-subxt crate in the local manifest.


ascjones (approved):
Thanks!

PR #178, opened 2021-02-15
Add Makefile for new project creation template

deadblackclover:
When I first worked with smart contracts, I noticed that the testing and build teams are quite long. And I thought, why not collect them in Makefile, to make the work easier. This is already done in the Substrate. I think this idea is not a bad one. Perhaps you can also add more commands.

### DIFF
--- a/templates/new/Makefile
+++ b/templates/new/Makefile
@@ -1,0 +1,11 @@
+.PHONY: build
+build:
+	cargo +nightly contract build
+
+.PHONY: test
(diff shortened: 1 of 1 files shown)


Robbepop:
Cargo allows you to create so-called `cargo aliases` with which you can very easily customize the build and test commands on a project level.

ascjones:
Yes I think we can leave it up to users to customize the commands locally.

PR #179, opened 2021-02-18
Fix bug with empty Wasm file when using system binaryen for optimization

cmichi:
@ascjones The regression which you found is unfortunately a bug which was introduced when making the `binaryen` dependency opt-in.

I've added a number of asserts now and fixed it.


src/cmd/build.rs
```diff
@@ -263,23 +269,20 @@ fn post_process_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
...
-    let mut dest_wasm_file = File::open(crate_metadata.dest_wasm.as_os_str())?;
-    let mut dest_wasm_file_content = Vec::new();
-    dest_wasm_file.read_to_end(&mut dest_wasm_file_content)?;
-
-    let optimized_wasm = do_optimization(crate_metadata, &optimized, &dest_wasm_file_content, 3)?;
+    let mut dest_optimized = crate_metadata.dest_wasm.clone();
+    dest_optimized.set_file_name(format!("{}-opt.wasm", crate_metadata.package_name));
 
-    let mut optimized_wasm_file = File::create(optimized.as_os_str())?;
-    optimized_wasm_file.write_all(&optimized_wasm)?;
```
cmichi, line 276: This was the bug here, the optimized Wasm file was created here and the returned bytes from `do_optimization` written into it.

But:
* This was only necessary for the `do_optimization` version of `feature = "binaryen-as-dependency"`.
* For `not(feature = "binaryen-as-dependency"`)` the `wasm-opt` cli was already invoked with `-o optimized-wasm-file`, thus already creating the file. Consequently the returned bytes from the function were always empty. I first tried to fix this bug by having `wasm-opt` return a raw bytestream, but found a bug in `wasm-opt` which currently prevents this: https://github.com/WebAssembly/binaryen/issues/3579.

cmichi: Maybe it would also make sense to add tests which try to build the contract either with `--features=binaryen-as-dependency` or without.

src/cmd/build.rs
```diff
@@ -294,11 +297,14 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<OptimizationResult> {
 /// If successful, the optimized Wasm is returned as a `Vec<u8>`.
```
ascjones, line 297: This comment should be updated to say that this now writes the file to `dest_wasm`

ascjones (approved):
LGTM. As you suggest we should add a CI test run with `binaryen-as-dependency` enabled.

PR #180, opened 2021-02-18
Bring `test-ci-only` back

cmichi:
We actually temporarily removed testing our CI with the feature `test-ci-only` in https://github.com/paritytech/cargo-contract/pull/114/files.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -439,7 +439,7 @@ pub(crate) fn execute_with_crate_metadata(
 #[cfg(feature = "test-ci-only")]
 #[cfg(test)]
 mod tests_ci_only {
-    use crate::{cmd, util::tests::with_tmp_dir, BuildArtifacts, ManifestPath, UnstableFlags};
+    use crate::{cmd, util::tests::with_tmp_dir, BuildArtifacts, ManifestPath, UnstableFlags, Verbosity};
 
     #[test]
     fn build_template() {
@@ -449,7 +449,7 @@ mod tests_ci_only {
                 ManifestPath::new(&path.join("new_project").join("Cargo.toml")).unwrap();
             let res = super::execute(
                 &manifest_path,
-                None,
+                Verbosity::Default,
                 true,
                 BuildArtifacts::All,
                 UnstableFlags::default(),
@@ -477,7 +477,7 @@ mod tests_ci_only {
             // when
             super::execute(
                 &manifest_path,
-                None,
+                Verbosity::Default,
                 true,
                 BuildArtifacts::CheckOnly,
                 UnstableFlags::default(),
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -439,7 +439,9 @@ pub(crate) fn execute_with_crate_metadata(
 #[cfg(feature = "test-ci-only")]
 #[cfg(test)]
 mod tests_ci_only {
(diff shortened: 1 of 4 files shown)


src/util.rs
```diff
@@ -64,7 +64,7 @@ where
         Verbosity::Default => &mut cmd,
     };
 
-    log::info!("invoking cargo: {:?}", cmd);
+    log::info!("Invoking cargo: {:?}", cmd);
```
cmichi, line 67: Just for consistency with the other `log` messages, which start uppercase.

src/cmd/build.rs
```diff
@@ -430,6 +430,9 @@ pub(crate) fn execute_with_crate_metadata(
         "Building cargo project".bright_green().bold()
     );
     build_cargo_project(&crate_metadata, build_artifact, verbosity, unstable_flags)?;
+    if build_artifact == BuildArtifacts::CheckOnly {
```
ascjones, line 433: Is this from some other PR?

src/main.rs
```diff
@@ -315,7 +315,7 @@ enum Command {
     /// Command has been deprecated, use `cargo contract build` instead
     #[structopt(name = "generate-metadata")]
     GenerateMetadata {},
-    /// Check that the code builds as Wasm; does not output any build artifact to the top level `target/` directory
+    /// Check that the code builds as Wasm; does not output any `<name>.contract` artifact to the `target/` directory
```
cmichi, line 318: We invoke `cargo check` with `--target-dir`. This actually results in a `target` folder in the `target-dir`.
```
--target-dir directory
Directory for all generated artifacts and intermediate files.
```

You can verify it locally:
```
$ cargo new foo
     Created binary (application) `foo` package
$ cd foo
$ cargo check
ls
    Checking foo v0.1.0 (/tmp/foo)
    Finished dev [unoptimized + debuginfo] target(s) in 2.10s
$ ls
Cargo.lock  Cargo.toml  src  target
```

src/cmd/build.rs
```diff
@@ -466,30 +471,33 @@ mod tests_ci_only {
                 ManifestPath::new(&path.join("new_project").join("Cargo.toml")).unwrap();
             let res = super::execute(
                 &manifest_path,
-                None,
+                Verbosity::Default,
```
cmichi, line 474: This is a required change, actually the code does not build without it. We just didn't notice because the `test-ci-only` feature was not included in the CI.

src/cmd/build.rs
```diff
@@ -430,6 +430,9 @@ pub(crate) fn execute_with_crate_metadata(
         "Building cargo project".bright_green().bold()
     );
     build_cargo_project(&crate_metadata, build_artifact, verbosity, unstable_flags)?;
+    if build_artifact == BuildArtifacts::CheckOnly {
+        return Ok((None, None));
+    }
```
cmichi, line 435: This early return is needed because the next step would otherwise try to post-process the contract. But check does not actually create a contract.

Without this change `cargo contract check` fails with 
```
 [2/2] Post processing wasm file
ERROR: Loading original wasm file '/tmp/foobar/target/ink/wasm32-unknown-unknown/release/foobar.wasm'

Caused by:
    Can't read from the file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

ascjones (approved):
LGTM

Issue #183, opened 2021-02-18
Installed wasm-opt optimization not working


ascjones:
Using locally installed `wasm-opt version 99 (version_73-1265-ga74bc72db)`

![image](https://user-images.githubusercontent.com/75586/108386713-cf66f980-7204-11eb-9506-d9385a76bbcd.png)

Discovered while testing https://github.com/paritytech/cargo-contract/pull/174, but it is also an issue on `master` (I did pull the latest this time!).

Possibly related to https://github.com/paritytech/cargo-contract/pull/179, but haven't done any troubleshooting yet.

cmichi:
Ah lol, I had the same today. It's due to an outdated `wasm-opt` on your system. Just do an upgrade of `binaryen` and you should be fine!

ascjones:
Built with https://github.com/WebAssembly/binaryen/commit/f7bfc85fae3bebf9ff818905c9bc6df51989ce70 which is the current `HEAD`

cmichi:
I've created https://github.com/WebAssembly/binaryen/issues/3587.

PR #187, opened 2021-02-19
Fix `wasm-opt` regression

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/183.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -23,7 +23,7 @@ use std::{
 };
 
 #[cfg(not(feature = "binaryen-as-dependency"))]
-use std::process::Command;
+use std::{process::Command, str};
 
 use crate::{
     crate_metadata::CrateMetadata,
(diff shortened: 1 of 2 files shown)


src/cmd/build.rs
```diff
@@ -352,6 +352,12 @@ fn do_optimization(
         .arg(format!("-O{}", optimization_level))
         .arg("-o")
         .arg(dest_optimized)
+
+        // the memory in our module is imported, `wasm-opt` needs to be told that
+        // the memory is initialized to zeros, otherwise it won't run the
+        // memory-packing pre-pass.
+        .arg("--zero-filled-memory")
```
ascjones, line 359: Does this also work with the old version of `wasm-opt`?

cmichi: That's a good point you raised. I updated the PR to show a helpful message in this case. That's how it currently looks:

```
ERROR: The wasm-opt optimization failed.               
A possible source of error could be that you have installed an outdated binaryen package.
                                                                               
The error which wasm-opt returned was:  
Unknown option '--zero-filled-memory' 
```

ascjones: Can we update the `README` to specify the minimum compatible installed version of `wasm-opt`? 

We could even check the version before executing to be at least `99` to be more accurate - if that is sufficient. But up to you.

cmichi: I added a note to the Readme!

ascjones (approved):
LGTM

Issue #34, opened 2020-02-10
path traversal of "cargo contract new"


taskooh:
I think "cargo contract new" should only allow alphabets , dash('-') and  numbers.

when I run this in Desktop,
```bash
Desktop % cargo contract new ../xxx
	Created contract ../xxx
```

directory "xxx" is created in home directory.
```bash
Desktop % cd ../xxx
xxx %
```

when I try to build from here,

```bash
cargo contract build
 [1/4] Collecting crate metadata
error: Error during execution of `cargo metadata`: error: failed to parse manifest at `/Users/xf9x/xxx/Cargo.toml`

Caused by:
  Invalid character `.` in package name: `../xxx`
```

build fails.

ascjones:
This should be straighforward: this check just needs to be enhanced: https://github.com/paritytech/cargo-contract/blob/master/src/cmd/new.rs#L27

ascjones:
Closed in #192

PR #192, opened 2021-02-21
no periods in new contract names

trace-andreason:
to address issue #34

### DIFF
--- a/src/cmd/new.rs
+++ b/src/cmd/new.rs
@@ -27,6 +27,10 @@ pub(crate) fn execute<P>(name: &str, dir: Option<P>) -> Result<Option<String>>
 where
     P: AsRef<Path>,
 {
+    if name.contains('.') {
+        anyhow::bail!("Contract names cannot contain periods");
+    }
+
     if name.contains('-') {
         anyhow::bail!("Contract names cannot contain hyphens");
     }
@@ -109,6 +113,19 @@ mod tests {
             assert_eq!(
                 result.err().unwrap().to_string(),
                 "Contract names cannot contain hyphens"
+            );
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn rejects_name_with_period() {
+        with_tmp_dir(|path| {
+            let result = execute("../xxx", Some(path));
+            assert!(result.is_err(), "Should fail");
+            assert_eq!(
+                result.err().unwrap().to_string(),
+                "Contract names cannot contain periods"
             );
             Ok(())
         })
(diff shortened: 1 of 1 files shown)


src/cmd/new.rs
```diff
@@ -27,6 +27,10 @@ pub(crate) fn execute<P>(name: &str, dir: Option<P>) -> Result<Option<String>>
 where
     P: AsRef<Path>,
 {
+    if name.contains('.') {
```
ascjones, line 30: Rather than checking for individual invalid characters, we should rather only allow valid characters e.g. alphanumeric, underscores.

trace-andreason:
@ascjones good call, made that change and removed these invalid character checks.

src/cmd/new.rs
```diff
@@ -27,8 +27,8 @@ pub(crate) fn execute<P>(name: &str, dir: Option<P>) -> Result<Option<String>>
 where
     P: AsRef<Path>,
 {
-    if name.contains('-') {
-        anyhow::bail!("Contract names cannot contain hyphens");
+    if !name.replace('_', "").chars().all(char::is_alphanumeric) {
```
ascjones, line 30:
```suggestion
    if !name.chars().all(|c| c.is_alphanumeric() || c == '_') {
```

I think something like this is more readable (note I haven't checked whether that compiles or not)

trace-andreason: Good idea, probably a tad more efficient too. Made the change and it does indeed compile

ascjones (approved):
LGTM. @Robbepop can you just confirmt the name vaildation correct here - we should only allow alphanumeric or underscore?

src/cmd/new.rs
```diff
@@ -27,8 +27,8 @@ pub(crate) fn execute<P>(name: &str, dir: Option<P>) -> Result<Option<String>>
 where
     P: AsRef<Path>,
 {
-    if name.contains('-') {
-        anyhow::bail!("Contract names cannot contain hyphens");
+    if !name.chars().all(|c| c.is_alphanumeric() || c == '_') {
```
Robbepop, line 30: this allows `_` as identifier which is invalid. however, I think this isn't too bad?

ascjones: Well spotted. Is that even a valid crate name?

Robbepop: I just checked and `cargo` does not immediately put a stop to having `_` as crate name but my filesystem definitely does. xD

Robbepop: Note sure we really need to guard against this. What is your opinion?

ascjones: Seems unnecessary

Issue #198, opened 2021-02-22
The delegator example is successful only on the first compilation.


ii-ii-ii:
Under examples/delegator, execute the command
`. /build-all.sh`
If the contract has not been compiled before, it will compile normally.
If it has been built before, you will get an error.
<img width="1431" alt="image" src="https://user-images.githubusercontent.com/69485494/108653714-10744c00-752c-11eb-99ef-d3fa7e254abc.png">

After deleting the target folder, it will compile normally.

PR #199, opened 2021-02-24
Fix linker error when building complex contracts

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/198 and https://github.com/paritytech/ink/issues/701.

My understanding is that the linker error was a result of conflicting build artifacts in the `target` directory. 

The root contract is build with the `ink-as-dependency` feature enabled for sub-contracts, whereas the sub-contracts are build without this flag. This results in the external function `__ink_generate_metadata` (which is only build when `ink-as-dependency` is disabled) not being available in the artifacts, but cargo still tries to reuse the artifacts. Hence the linker error.

I fixed it now by building sub-contracts into sub-folders of `target/ink/`. This way all artifacts are separate.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -14,16 +14,11 @@
 // You should have received a copy of the GNU General Public License
 // along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
 
-use crate::workspace::ManifestPath;
 use crate::Verbosity;
 
 use anyhow::{Context, Result};
 use rustc_version::Channel;
-use std::{
-    ffi::OsStr,
-    path::{Path, PathBuf},
-    process::Command,
-};
+use std::{ffi::OsStr, path::Path, process::Command};
 
 /// Check whether the current rust channel is valid: `nightly` is recommended.
 pub fn assert_channel() -> Result<()> {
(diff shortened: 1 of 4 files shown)


ascjones:
Ahhh I was wondering what I was getting at in the original issue #60 

> Need to auto remove the ink-as-dependency feature in the root workspace manifest for the sub contracts

cmichi:
> Ahhh I was wondering what I was getting at in the original issue #60
> 
> > Need to auto remove the ink-as-dependency feature in the root workspace manifest for the sub contracts

Yeah me too when I implemented it :grimacing:.

ascjones:
Well if I don't even understand my own cryptic comments I don't expect you to :sweat:

src/util.rs
```diff
@@ -92,6 +98,15 @@ pub(crate) fn base_name(path: &Path) -> &str {
         .expect("must be valid utf-8")
 }
 
+/// Returns the absolute path to the manifest directory.
+pub(crate) fn absolute_path(manifest_path: &ManifestPath) -> Result<PathBuf, std::io::Error> {
```
ascjones, line 102: This should be a method on `ManifestPath`

ascjones (approved):
LGTM, we'll need to do a `0.9.1` release with this

Issue #169, opened 2021-02-04
Give contract authors control over optimization passes


cmichi:
## Motivation

In `src/cmd/build.rs` we currently pass the number of optimization passes `3` to `binaryen` by default. However, it is desirable to give contract authors control over this setting. Hence we want to expose it on the cli.

## ToDo

- [ ] `cargo contract build` should have an additional `--optimization-passes` parameter which makes this variable customizable.

clearloop:
Hi @cmichi , we have a modification in #131 related to the opt passes, if there is any plan on the wasm optimization in `cargo-contract`, please ping us ( @patractlabs )

ascjones:
In addition we could allow this to be configurable via a custom section in a contract's `Cargo.toml`.

PR #216, opened 2021-03-07
Adding "optimization-passes" option

trace-andreason:
Addressing issue #169

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -29,8 +29,8 @@ use crate::{
     crate_metadata::CrateMetadata,
     maybe_println, util, validate_wasm,
     workspace::{ManifestPath, Profile, Workspace},
-    BuildArtifacts, BuildResult, OptimizationResult, UnstableFlags, UnstableOptions, Verbosity,
-    VerbosityFlags,
+    BuildArtifacts, BuildResult, OptimizationFlags, OptimizationPasses, OptimizationResult,
+    UnstableFlags, UnstableOptions, Verbosity, VerbosityFlags,
 };
 use anyhow::{Context, Result};
 use colored::Colorize;
@@ -66,8 +66,8 @@ pub struct BuildCommand {
     verbosity: VerbosityFlags,
     #[structopt(flatten)]
     unstable_options: UnstableOptions,
-    #[structopt(long = "optimization-passes", default_value = "3")]
-    optimization_passes: u32,
+    #[structopt(flatten)]
+    optimization_passes: OptimizationFlags,
 }
 
 impl BuildCommand {
@@ -76,13 +76,15 @@ impl BuildCommand {
         let unstable_flags: UnstableFlags =
             TryFrom::<&UnstableOptions>::try_from(&self.unstable_options)?;
         let verbosity = TryFrom::<&VerbosityFlags>::try_from(&self.verbosity)?;
+        let optimization_passes =
+            TryFrom::<&OptimizationFlags>::try_from(&self.optimization_passes)?;
         execute(
             &manifest_path,
             verbosity,
             true,
             self.build_artifact,
             unstable_flags,
-            self.optimization_passes,
+            optimization_passes,
         )
     }
 }
@@ -105,7 +107,7 @@ impl CheckCommand {
         let unstable_flags: UnstableFlags =
             TryFrom::<&UnstableOptions>::try_from(&self.unstable_options)?;
         let verbosity: Verbosity = TryFrom::<&VerbosityFlags>::try_from(&self.verbosity)?;
-        let optimization_passes = 0;
+        let optimization_passes = OptimizationPasses::Zero;
         execute(
             &manifest_path,
             verbosity,
@@ -137,7 +139,6 @@ fn build_cargo_project(
     build_artifact: BuildArtifacts,
     verbosity: Verbosity,
     unstable_flags: UnstableFlags,
-    _optimization_passes: u32,
 ) -> Result<()> {
     util::assert_channel()?;
 
@@ -276,7 +277,7 @@ fn post_process_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
 /// optimizations (or bugs?) between Rust and Wasm.
 fn optimize_wasm(
     crate_metadata: &CrateMetadata,
-    optimization_passes: u32,
+    optimization_passes: OptimizationPasses,
 ) -> Result<OptimizationResult> {
     let mut dest_optimized = crate_metadata.dest_wasm.clone();
     dest_optimized.set_file_name(format!("{}-opt.wasm", crate_metadata.package_name));
@@ -307,7 +308,7 @@ fn optimize_wasm(
 fn do_optimization(
     dest_wasm: &OsStr,
     dest_optimized: &OsStr,
-    optimization_level: u32,
+    optimization_level: OptimizationPasses,
 ) -> Result<()> {
     let mut dest_wasm_file = File::open(dest_wasm)?;
     let mut dest_wasm_file_content = Vec::new();
@@ -315,9 +316,9 @@ fn do_optimization(
 
     let codegen_config = binaryen::CodegenConfig {
         // number of optimization passes (spends potentially a lot of time optimizing)
-        optimization_level,
+        optimization_level: optimization_level.to_passes(),
         // the default
-        shrink_level: 1,
+        shrink_level: optimization_level.to_shrink(),
         // the default
         debug_info: false,
     };
@@ -342,7 +343,7 @@ fn do_optimization(
 fn do_optimization(
     dest_wasm: &OsStr,
     dest_optimized: &OsStr,
-    optimization_level: u32,
+    optimization_level: OptimizationPasses,
 ) -> Result<()> {
     // check `wasm-opt` is installed
     if which::which("wasm-opt").is_err() {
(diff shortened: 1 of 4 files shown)


src/cmd/build.rs
```diff
@@ -66,6 +66,8 @@ pub struct BuildCommand {
     verbosity: VerbosityFlags,
     #[structopt(flatten)]
     unstable_options: UnstableOptions,
+    #[structopt(long = "optimization-passes", default_value = "3")]
+    optimization_passes: u32,
```
ascjones, line 70: I'd like to see this restricted to the valid values, perhaps by introducing an enum type. Also if you look at `wasm-opt --help` there are a couple of non integer values possible:

```
 -O0                                          execute no optimization passes 
   -O1                                          execute -O1 optimization passes
                                                (quick&useful opts, useful for 
                                                iteration builds) 
   -O2                                          execute -O2 optimization passes
                                                (most opts, generally gets most
                                                perf) 
   -O3                                          execute -O3 optimization passes
                                                (spends potentially a lot of 
                                                time optimizing) 
   -O4                                          execute -O4 optimization passes
                                                (also flatten the IR, which can
                                                take a lot more time and memory,
                                                but is useful on more nested / 
                                                complex / less-optimized input)
   -Os                                          execute default optimization 
                                                passes, focusing on code size 
   -Oz                                          execute default optimization 
                                                passes, super-focusing on code 
                                                size 
```

trace-andreason: @ascjones ahhh, I had assumed it was limited to integers as it was typed that way in the `do_optimization` function. I'll give the enum a try.

trace-andreason: oh, if we wanted to go the enum route, it might create some problems with the `binaryen-as-dependency` optionsfor installation which seems to only take a `u32` for optimization passes: https://docs.rs/binaryen/0.12.0/src/binaryen/lib.rs.html#21-28

ascjones: You can always convert from the enum to the `u32` at the call site. Would be worth finding out if `5 -> s` and `6 -> z`. At the very least we should validate the input value so it can't just be any `u32`.

trace-andreason: did a little more digging on the binaryen crate:
```
/// Codegen configuration.
#[derive(Default)]
pub struct CodegenConfig {
    /// 0, 1, 2 correspond to -O0, -Os, -Oz
    pub shrink_level: u32,
    /// 0, 1, 2 correspond to -O0, -O1, -O2, etc.
    pub optimization_level: u32,
    /// If set, the names section is emitted.
    pub debug_info: bool,
}
```
I made the shrink options default to 3 optimization passes in the case of binaryen as a dependency

ascjones: Awesome, I'd like @cmichi to take a look at this too since he's more familiar with the `wasm-opt` integration

src/main.rs
```diff
@@ -93,6 +93,83 @@ impl ExtrinsicOpts {
...
+            OptimizationPasses::One => "1",
+            OptimizationPasses::Two => "2",
+            OptimizationPasses::Three => "3",
+            OptimizationPasses::Four => "4",
+            OptimizationPasses::S => "s",
+            OptimizationPasses::Z => "z",
+        }
+    }
+
+    /// to_passes returns number of optimization passes to do
```
cmichi, line 150:
```suggestion
    /// Returns number of optimization passes to do
```

src/main.rs
```diff
@@ -93,6 +93,83 @@ impl ExtrinsicOpts {
...
+            OptimizationPasses::Zero => 0,
+            OptimizationPasses::One => 1,
+            OptimizationPasses::Two => 2,
+            OptimizationPasses::Three => 3,
+            OptimizationPasses::Four => 4,
+            _ => 3, // Default to three for shrink settings
+        }
+    }
+
+    /// to_shrink returns amount of shrinkage to do
```
cmichi, line 162:
```suggestion
    /// Returns amount of shrinkage to do
```

src/main.rs
```diff
@@ -93,6 +93,83 @@ impl ExtrinsicOpts {
     }
 }
 
+#[derive(Clone, Debug, StructOpt)]
+pub struct OptimizationFlags {
+    // number of optimization passes
+    #[structopt(long = "optimization-passes", default_value = "3")]
+    optimization_passes: String,
```
cmichi, line 100: I think this function should also contain 
```
    // How much to focus on shrinking the code size
    #[structopt(long = "shrink-level", default_value = "…")]
    shrink_level: String,
```

You seem to derive the `shrink-level` from the `optimization-passes` currently, but in `wasm-opt` it' an independent cli flag.

src/main.rs
```diff
@@ -93,6 +93,83 @@ impl ExtrinsicOpts {
...
+            _ => anyhow::bail!(
+                "Unknown optimization passes option {}",
+                value.optimization_passes
+            ),
+        }
+    }
+}
+
+impl OptimizationPasses {
+    /// Returns string representation of OptimizationPasses
```
cmichi, line 137:
```suggestion
    /// Returns the string representation of `OptimizationPasses`
```

src/cmd/build.rs
cmichi: Right now this is the cli help output:
```
$ cargo run -- contract build --help
…
        --optimization-passes <optimization-passes>    
             [default: 3]
```
What I would like to see is either 
* a note with possible params and a sentence noting that they are simply forwarded to `wasm-opt` or
* a list of all possible values with each ones description (like the list which @ascjones posted above).

src/main.rs
```diff
@@ -93,6 +93,83 @@ impl ExtrinsicOpts {
...
+            OptimizationPasses::Two => "2",
+            OptimizationPasses::Three => "3",
+            OptimizationPasses::Four => "4",
+            OptimizationPasses::S => "s",
+            OptimizationPasses::Z => "z",
+        }
+    }
+
+    /// to_passes returns number of optimization passes to do
+    pub(crate) fn to_passes(&self) -> u32 {
```
cmichi, line 151: Right now `cargo run` results in
```
warning: associated function is never used: `to_passes`
   --> src/main.rs:151:19
    |                                                                          
151 |     pub(crate) fn to_passes(&self) -> u32 {
    |                   ^^^^^^^^^
    |                                                                                                                                                          
    = note: `#[warn(dead_code)]` on by default
                                                                                                                                                               
warning: associated function is never used: `to_shrink`        
   --> src/main.rs:163:19                                                      
    |                                                                          
163 |     pub(crate) fn to_shrink(&self) -> u32 {
    |                   ^^^^^^^^^                                              
                                       
warning: 2 warnings emitted
```

This is because those two functions are only invoked when `binaryen` is used as a Rust dependency, instead of using a binary available on the system. You can fix it by making those two functions only available in the case of a Rust dependency.

```suggestion
    #[cfg(feature = "binaryen-as-dependency")]
    pub(crate) fn to_passes(&self) -> u32 {
```

src/main.rs
```diff
@@ -93,6 +93,83 @@ impl ExtrinsicOpts {
...
+            OptimizationPasses::One => 1,
+            OptimizationPasses::Two => 2,
+            OptimizationPasses::Three => 3,
+            OptimizationPasses::Four => 4,
+            _ => 3, // Default to three for shrink settings
+        }
+    }
+
+    /// to_shrink returns amount of shrinkage to do
+    pub(crate) fn to_shrink(&self) -> u32 {
```
cmichi, line 163:
```suggestion
    #[cfg(feature = "binaryen-as-dependency")]
    pub(crate) fn to_shrink(&self) -> u32 {
```

cmichi (changes requested):

src/cmd/build.rs
trace-andreason: @cmichi added. Here is what the new cli help output looks like:

```
        --optimization-passes <optimization-passes>
            number of optimization passes, passed as an argument to wasm-opt
            
            - `0`: execute no optimization passes
            
            - `1`: execute 1 optimization pass (quick & useful opts, useful for iteration builds)
            
            - `2`, execute 2 optimization passes (most opts, generally gets most perf)
            
            - `3`, execute 3 optimization passes (spends potentially a lot of time optimizing)
            
            - `4`, execute 4 optimization passes (also flatten the IR, which can take a lot more time and memory but is
            useful on more nested / complex / less-optimized input)
            
            - `s`, execute default optimization passes, focusing on code size
            
            - `z`, execute default optimization passes, super-focusing on code size
            
            - [default: 3]
```

src/cmd/build.rs
```diff
@@ -299,17 +308,17 @@ fn optimize_wasm(crate_metadata: &CrateMetadata) -> Result<OptimizationResult> {
...
     dest_optimized: &OsStr,
-    optimization_level: u32,
+    optimization_level: OptimizationPasses,
 ) -> Result<()> {
     let mut dest_wasm_file = File::open(dest_wasm)?;
     let mut dest_wasm_file_content = Vec::new();
     dest_wasm_file.read_to_end(&mut dest_wasm_file_content)?;
 
     let codegen_config = binaryen::CodegenConfig {
         // number of optimization passes (spends potentially a lot of time optimizing)
```
cmichi, line 318:
```suggestion
        // Number of optimization passes (spends potentially a lot of time optimizing)
```

cmichi (approved):
Thanks @trace-andreason !

PR #223, opened 2021-03-10
Refactor build command

ascjones:
Refactor triggered by discussion around https://github.com/paritytech/cargo-contract/pull/180/files#r591270056

- Refactor the build command in order to consolidate the logic around `BuildArtifacts` to a single place.
- No longer executes the build from within `metadata.rs`
- Replaces the `MetadataCommand` with free functions since it is no longer a top level command (this explains much of the diff in `metadata.rs`, but the code is mostly unchanged apart from the above mentioned change).


cmichi (approved):
LGTM, thanks!

There are some conflicts with master since the latest merge.

PR #231, opened 2021-03-23
Support `optimization-passes` in the release profile

cmichi:
Final piece for closing https://github.com/paritytech/cargo-contract/issues/169.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -608,7 +608,14 @@ mod tests_ci_only {
             // then
             // we have to truncate here to account for a possible small delta
             // in the floating point numbers
-            assert!(optimization.optimized_size.trunc() == optimization.original_size.trunc());
+            let optimized_size = optimization.optimized_size.trunc();
+            let original_size = optimization.original_size.trunc();
+            assert!(
+                optimized_size == original_size,
+                "The optimized size {:?} differs from the original size {:?}",
+                optimized_size,
+                original_size
+            );
 
             Ok(())
         })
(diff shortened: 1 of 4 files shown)


src/cmd/build.rs
```diff
@@ -352,13 +401,17 @@ fn do_optimization(
...
+    );
     let output = Command::new("wasm-opt")
         .arg(dest_wasm)
-        .arg(format!("-O{}", optimization_level.to_str()))
+        .arg(format!("-O{}", optimization_level))
         .arg("-o")
         .arg(dest_optimized)
         // the memory in our module is imported, `wasm-opt` needs to be told that
-        // the memory is initialized to zeros, otherwise it won't run the
+        // the memory is initialized to zeroes, otherwise it won't run the
```
ascjones, line 414: I believe both zeros and zeroes are acceptable.

ascjones (commented):
Looks good. Is adding a custom key to a cargo profile the correct way to do it? Should we add our own e.g. `cargo-contract` custom section for this?

src/workspace/manifest.rs
```diff
@@ -160,6 +162,26 @@ impl Manifest {
...
+        self.get_profile_release_table_mut()
+            .ok()?
+            .get("optimization-passes")
+            .map(|val| val.to_string())
+            .map(Into::into)
+    }
+
+    /// Extract optimization-passes from `[profile.release]`
+    #[cfg(test)]
+    pub fn set_profile_release_optimization_passes(
```
ascjones, line 176: Could just add this inline to the tests since `get_profile_release_table_mut` is public

ascjones (approved):
LGTM

PR #236, opened 2021-03-25
Add missing derive

cmichi:
`master` currently fails because of the missing derive. Not sure why this wasn't catched in the CI run before the merge.

I also noticed that one of the tests which I added doesn't actually `manifest.write()`.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -584,17 +584,21 @@ mod tests_ci_only {
     }
 
     #[test]
-    fn cli_optimization_passes_must_take_precedence_over_profile() {
+    fn optimization_passes_from_cli_must_take_precedence_over_profile() {
         with_tmp_dir(|path| {
             // given
             cmd::new::execute("new_project", Some(path)).expect("new project creation failed");
             let cargo_toml_path = path.join("new_project").join("Cargo.toml");
             let manifest_path =
                 ManifestPath::new(&cargo_toml_path).expect("manifest path creation failed");
+
             // we write "4" as the optimization passes into the release profile
-            assert!(Manifest::new(manifest_path.clone())?
+            let mut manifest = Manifest::new(manifest_path.clone())?;
+            assert!(manifest
                 .set_profile_optimization_passes(String::from("4").into())
                 .is_ok());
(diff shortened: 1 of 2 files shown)


src/cmd/build.rs
```diff
@@ -626,4 +637,39 @@ mod tests_ci_only {
...
+                .optimization_result
+                .expect("no optimization result available");
+
+            // then
+            assert!(
+                optimization.optimized_size < optimization.original_size * 0.5,
+                "The optimized size {:?} DOES NOT differ enough from the original size {:?}",
+                optimization.optimized_size,
+                optimization.original_size
+            );
```
Robbepop, line 670: This 0.5 magic number looks like a hack tbh ... is there no better solution to testify?

src/cmd/build.rs
```diff
@@ -513,6 +513,22 @@ mod tests_ci_only {
...
+        let manifest_path =
+            ManifestPath::new(cargo_toml_path).expect("manifest path creation failed");
+        let mut manifest = Manifest::new(manifest_path.clone()).expect("manifest creation failed");
+        manifest
+            .set_profile_optimization_passes(passes)
+            .expect("setting `optimization-passes` in profile failed");
+        manifest
+            .write(&manifest_path)
+            .expect("writing manifest failed");
+    }
```
Robbepop, line 531: Can you add a comment that briefly explains what this is doing?

src/main.rs
```diff
@@ -99,7 +99,7 @@ impl ExtrinsicOpts {
 }
 
 #[derive(Clone, Copy, Debug)]
-#[cfg_attr(test, derive(PartialEq))]
+#[cfg_attr(any(test, feature = "binaryen-as-dependency"), derive(PartialEq))]
```
ascjones, line 102: Why do we only derive `PartialEq` under these features?

cmichi: Only needed if either of these are enabled.

ascjones: Yeah sure but why be stingy and not give it to everyone?

ascjones (approved):
LGTM

Issue #232, opened 2021-03-23
Detect mismatches of `parity-scale-codec` in contract vs. ink!


cmichi:
Escalated from https://github.com/paritytech/ink-docs/pull/16.

A version mismatch results in a confusing `the trait "WrapperTypeEncode" is not implemented for …` error. We regularly get questions about this in the public channel. It's really not obvious what the source of error is and we should detect it.

PR #237, opened 2021-03-25
Detect version mismatches of `parity-scale-codec`

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/232.

CI currently fails because `master` fails (see https://github.com/paritytech/cargo-contract/pull/236 for the fix for that).

@Robbepop @ascjones Are you aware of any other dependencies of a contract which must be compatible with the ink! version of that dependency? `scale-info`?

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -439,10 +439,12 @@ fn assert_compatible_ink_dependencies(
     manifest_path: &ManifestPath,
     verbosity: Verbosity,
 ) -> Result<()> {
-    let args = ["-i=parity-scale-codec", "--duplicates"];
-    util::invoke_cargo("tree", &args, manifest_path.directory(), verbosity)
-        .map(|_| ())
-        .map_err(|_| anyhow::anyhow!("Mismatching versions of `parity-scale-codec` were found!"))
+    for dependency in ["parity-scale-codec", "scale-info"].iter() {
+        let args = ["-i", dependency, "--duplicates"];
+        let _ = util::invoke_cargo("tree", &args, manifest_path.directory(), verbosity)
+            .map_err(|_| anyhow::anyhow!("Mismatching versions of `{}` were found!", dependency))?;
+    }
+    Ok(())
 }
 
 /// Executes build of the smart-contract which produces a wasm binary that is ready for deploying.
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -433,8 +433,13 @@ fn do_optimization(
 /// Asserts that the contract's dependencies are compatible to the ones used in ink!.
 ///
 /// This function utilizes `cargo tree`, which takes semver into consideration.
-/// Hence this function only returns an `Err` if it is a proper mismatch in the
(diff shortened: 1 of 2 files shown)


ascjones:
Yes the versions of `scale-info` should match

src/cmd/build.rs
```diff
@@ -430,6 +430,23 @@ fn do_optimization(
     Ok(())
 }
 
+/// Asserts that the contract's dependencies are compatible to the ones used in ink!.
+///
+/// This function utilizes `cargo tree`, which takes semver into consideration.
+/// Hence this function only returns an `Err` if it is a proper mismatch in the
+/// major version, and not if the versions mismatch on the minor/patch version.
```
ascjones, line 437: Hm at the moment `scale-info` is < `1.0` so breaking changes might come in minor version bumps.

cmichi: Well then the versioning for `scale-info` is broken. Semver is pretty clear about this:

>MAJOR version when you make incompatible API changes,
>MINOR version when you add functionality in a backwards compatible manner, and

cmichi: Come to think about it, I know another tool which then doesn't use semver properly 🙈

cmichi: I mean I could hardcode a version mismatch for `scale-info` if the minor version is not the same? That's pretty hacky, but I guess the versioning scheme for `scale-info` won't change anytime soon.

ascjones: > Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable

https://semver.org/#spec-item-4

ascjones: We'll go for `1.0` soon, hopefully in time for `ink! 3.0.0` proper

cmichi: Just noticed that `cargo-tree` behaves correcly with `0.y.z`:
```
$ cargo tree --duplicates -i scale-info
error: There are multiple `scale-info` packages in your project, and the specification `scale-info` is ambiguous.
  scale-info:0.3.0
  scale-info:0.6.0
```
So it already detects version mismatches for `0.`.

cmichi:
That's the current output:
```
$ cargo run -- contract build --manifest-path /tmp/foo/Cargo.toml 
error: There are multiple `scale-info` packages in your project, and the specification `scale-info` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  scale-info:0.3.0
  scale-info:0.6.0
ERROR: Mismatching versions of `scale-info` were found!
Please ensure that your contract and your ink! dependency use a compatible version of this package.
```
Mismatches of `parith-scale-codec` and `scale-info` are detected.

ascjones:
Looks good but CI failing with compilation error.

ascjones:
CI fix merged to master so just needs merging in here

ascjones (approved):
Nice! This should save us a lot of time answering this common problem.

PR #242, opened 2021-03-30
Detect compatible `wasm-opt` versions and improve error messages

cmichi:
The improved error message:
```
ERROR: wasm-opt not found!
We use this tool to optimize the size of your contract's Wasm binary.

wasm-opt is part of the binaryen package. You can find detailed
installation instructions on https://github.com/WebAssembly/binaryen#tools.

There are also ready-to-install packages for many platforms:
* Debian/Ubuntu: https://tracker.debian.org/pkg/binaryen
* Homebrew: https://formulae.brew.sh/formula/binaryen
* Arch Linux: https://archlinux.org/packages/community/x86_64/binaryen
* Windows: binary releases are available at https://github.com/WebAssembly/binaryen/releases
```
I'm aware that linking packages here can be seen as controversial, but due to the number of questions we get in public channels regarding this I think we should consider it.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -413,10 +413,7 @@ fn do_optimization(
         .as_ref()
         .expect("we just checked if which returned an err; qed")
         .as_path();
-    log::info!(
-        "Path to wasm-opt executable: {}",
-        wasm_opt_path.display()
-    );
+    log::info!("Path to wasm-opt executable: {}", wasm_opt_path.display());
 
     let _ = check_wasm_opt_version_compatibility(wasm_opt_path)?;
 
@@ -434,7 +431,13 @@ fn do_optimization(
         // memory-packing pre-pass.
         .arg("--zero-filled-memory")
         .output()
-        .map_err(|err| anyhow::anyhow!("Executing {} failed with {:?}", wasm_opt_path.display(), err))?;
+        .map_err(|err| {
+            anyhow::anyhow!(
+                "Executing {} failed with {:?}",
+                wasm_opt_path.display(),
+                err
+            )
+        })?;
 
     if !output.status.success() {
         let err = str::from_utf8(&output.stderr)
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -429,7 +429,7 @@ fn do_optimization(
         // memory-packing pre-pass.
         .arg("--zero-filled-memory")
         .output()
-        .map_err(|err| anyhow::anyhow!("Executing {:?} failed with {:?}", wasm_opt_path, err))?;
+        .map_err(|err| anyhow::anyhow!("Executing {} failed with {:?}", wasm_opt_path.display(), err))?;
 
     if !output.status.success() {
         let err = str::from_utf8(&output.stderr)
@@ -456,7 +456,7 @@ fn check_wasm_opt_version_compatibility(wasm_opt_path: &Path) -> Result<()> {
         .map_err(|err| {
             anyhow::anyhow!(
                 "Executing `{:?} --version` failed with {:?}",
-                wasm_opt_path,
+                wasm_opt_path.display(),
                 err
             )
         })?;
@@ -845,7 +845,7 @@ mod tests_ci_only {
             {
                 let mut file = std::fs::File::create(&path).unwrap();
                 file.write_all(
-                    b"#!/bin/sh\necho \"wasm-opt version 13 (version_13-79-gc12cc3f50)\"",
+                    b"#!/bin/sh\necho \"wasm-opt version 98 (version_13-79-gc12cc3f50)\"",
                 )
                 .expect("writing wasm-opt-mocked failed");
             }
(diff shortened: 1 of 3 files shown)


src/cmd/build.rs
```diff
@@ -414,22 +427,84 @@ fn do_optimization(
...
+        .next()
+        .ok_or_else(|| anyhow::anyhow!("Cannot get version information for {:?}", version_stdout))?
+        .parse()
+        .map_err(|err| {
+            anyhow::anyhow!(
+                "Parsing version number failed with {:?} for {:?}",
+                err,
+                version_stdout
+            )
+        })?;
```
ascjones, line 496: Hate to say it but a regex :scream: might be appropriate here

ascjones:
> I'm aware that linking packages here can be seen as controversial, but due to the number of questions we get in public channels regarding this I think we should consider it.

Worth it to decrease the support burden IMO.

src/cmd/build.rs
```diff
@@ -391,21 +392,34 @@ fn do_optimization(
...
+            * Arch Linux: https://archlinux.org/packages/community/x86_64/binaryen\n\
+            * Windows: binary releases are available at https://github.com/WebAssembly/binaryen/releases"
+            .to_string()
+            .bright_yellow()
         );
     }
+    let wasm_opt_path = which
+        .as_ref()
+        .expect("we just checked if which returned an err; qed")
+        .as_path();
```
ascjones, line 415: I assume this is the absolute path now, so probably worth doing a `log::info` with that path.

src/cmd/build.rs
```diff
@@ -391,21 +392,34 @@ fn do_optimization(
...
-    if which::which("wasm-opt").is_err() {
+    let which = which::which("wasm-opt");
+    if which.is_err() {
         anyhow::bail!(
-            "{}",
-            "wasm-opt is not installed. Install this tool on your system in order to \n\
-             reduce the size of your contract's Wasm binary. \n\
-             See https://github.com/WebAssembly/binaryen#tools"
-                .bright_yellow()
+            "wasm-opt not found!\n\
```
ascjones, line 398: We might add that we expect it to be installed on the `PATH`, to be absolutely clear.

src/cmd/build.rs
```diff
@@ -414,22 +428,87 @@ fn do_optimization(
...
+/// Currently this must be a version >= 99.
+#[cfg(any(test, not(feature = "binaryen-as-dependency")))]
+fn check_wasm_opt_version_compatibility(wasm_opt_path: &Path) -> Result<()> {
+    let cmd = Command::new(wasm_opt_path)
+        .arg("--version")
+        .output()
+        .map_err(|err| {
+            anyhow::anyhow!(
+                "Executing `{:?} --version` failed with {:?}",
+                wasm_opt_path,
```
ascjones, line 459:
```suggestion
                wasm_opt_path.display(),
```

src/cmd/build.rs
```diff
@@ -414,22 +428,87 @@ fn do_optimization(
         // the memory is initialized to zeroes, otherwise it won't run the
         // memory-packing pre-pass.
         .arg("--zero-filled-memory")
-        .output()?;
+        .output()
+        .map_err(|err| anyhow::anyhow!("Executing {:?} failed with {:?}", wasm_opt_path, err))?;
```
ascjones, line 432:
```suggestion
        .map_err(|err| anyhow::anyhow!("Executing {} failed with {:?}", wasm_opt_path.display(), err))?;
```

src/cmd/build.rs
```diff
@@ -757,4 +836,58 @@ mod tests_ci_only {
...
+
+    #[test]
+    fn incompatible_wasm_opt_version_must_be_detected() {
+        with_tmp_dir(|path| {
+            // given
+            let path = path.join("wasm-opt-mocked");
+            {
+                let mut file = std::fs::File::create(&path).unwrap();
+                file.write_all(
+                    b"#!/bin/sh\necho \"wasm-opt version 13 (version_13-79-gc12cc3f50)\"",
```
ascjones, line 848:
```suggestion
                    b"#!/bin/sh\necho \"wasm-opt version 98 (version_13-79-gc12cc3f50)\"",
```

src/cmd/build.rs
```diff
@@ -757,4 +836,58 @@ mod tests_ci_only {
...
+                .expect("setting permissions failed");
+
+            // when
+            let res = check_wasm_opt_version_compatibility(&path);
+
+            // then
+            assert!(res.is_err());
+            assert_eq!(
+                format!("{:?}", res),
+                "Err(Your wasm-opt version is 13, but we require a version >= 99.)"
```
ascjones, line 862:
```suggestion
                "Err(Your wasm-opt version is 98, but we require a version >= 99.)"
```

src/cmd/build.rs
ascjones: Nice idea!

ascjones (approved):
LGTM

PR #243, opened 2021-03-30
Check wasm optimized output file exists

ascjones:
Possible improvement for the source of this error message here:

image

Suspect that the installed version of `wasm-opt` runs but doesn't produce the output file as expected.

I haven't reproduced it because my installation of `wasm-opt` is valid and produces a optimized file, however this is a useful check to make anyway, since we have no guarantees that `do_optimization` will write the optimized wasm fle.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -283,7 +283,9 @@ fn optimize_wasm(
     )?;
 
     if !dest_optimized.exists() {
-        return Err(anyhow::anyhow!("Optimization failed, optimized wasm output file not found."))
+        return Err(anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


src/cmd/build.rs
```diff
@@ -282,6 +282,12 @@ fn optimize_wasm(
         optimization_passes,
     )?;
 
+    if !dest_optimized.exists() {
+        return Err(anyhow::anyhow!(
+            "Optimization failed, optimized wasm output file not found."
```
cmichi, line 287: Maybe worth adding `dest_optimized` to the error message?

cmichi (approved):
LGTM!

Issue #247, opened 2021-04-02
wasm-opt version error on contract build


rrybarczyk:
This error has been replicated on both Arch Linux (by installing `binaryen` via `$ yay binaryen`) and OSX (by installing `binaryen` via `$ brew install binaryen`).

After installing `binaryen` with the preferred installation method,  running `$ cargo +nightly contract build` fails with the following error:

```
ERROR: Unable to extract version information from wasm-opt version 100.
Your wasm-opt version is most probably too old. Make sure you use a version >= 99.
```

However, the `wasm-opt` version installed is the latest:
```
$ wasm-opt --version
> wasm-opt version 100
```

I also tried building the tagged `wasm-opt` version 99, which confusingly installed `wasm-opt` version 98. When trying to build with this version, it expectedly does not work.

The work around to this is by using the unpreferred method of building `binaryen` as a dependency when installing cargo-contract:

```
cargo install --force --features binaryen-as-dependency cargo-contract
```

netrikare:
It looks like wasm-opt version checking regex assumes additional string after version number. Solution:
change: 
`let re = Regex::new(r"wasm-opt version (\d+)\s+").unwrap();`
to:          
`let re = Regex::new(r"wasm-opt version (\d+)\s*+").unwrap();`

in [src/cmd/build.rs](https://github.com/paritytech/cargo-contract/blob/cacf2d54f79b8184520dafc790302710d7f087d9/src/cmd/build.rs#L496)

cmichi:
Thanks @rrybarczyk and @netrikare! Sorry about the late response, just saw it this morning, easter collided with this ticket :see_no_evil:.

We just issued a new release which fixes this issue and will make some adaptions to our CI setup to catch stuff like this earlier.

You can update to the latest version via
```
cargo install cargo-contract --version ^0.11 --force --locked
```

PR #248, opened 2021-04-06
Fix `wasm-opt --version` parsing

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/247.

The issue was that binaryen includes more extensive version information in the `--version` output [only if built from source](https://github.com/WebAssembly/binaryen/blob/main/CMakeLists.txt#L20-L35). If installed via a package manager the version output can be more compact.

I've also escalated this into the ticket https://github.com/paritytech/ci_cd/issues/121, to install `binaryen` from a package manager in our CI.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -496,7 +496,7 @@ fn check_wasm_opt_version_compatibility(wasm_opt_path: &Path) -> Result<()> {
     let re = Regex::new(r"wasm-opt version (\d+)").unwrap();
     let captures = re.captures(version_stdout).ok_or_else(|| {
         anyhow::anyhow!(
-            "Unable to extract version information from {}.\n\
+            "Unable to extract version information from \"{}\".\n\
             Your wasm-opt version is most probably too old. Make sure you use a version >= 99.",
             version_stdout
         )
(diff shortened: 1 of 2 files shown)


src/cmd/build.rs
```diff
@@ -493,30 +493,37 @@ fn check_wasm_opt_version_compatibility(wasm_opt_path: &Path) -> Result<()> {
     let version_stdout = str::from_utf8(&cmd.stdout)
         .expect("Cannot convert stdout output of wasm-opt to string")
         .trim();
-    let re = Regex::new(r"wasm-opt version (\d+)\s+").unwrap();
+    let re = Regex::new(r"wasm-opt version (\d+)").unwrap();
```
ascjones, line 496: I didn't spot it in the original PR but this should be an `expect("invalid regex")` or such.

src/cmd/build.rs
```diff
@@ -493,30 +493,37 @@ fn check_wasm_opt_version_compatibility(wasm_opt_path: &Path) -> Result<()> {
...
     let captures = re.captures(version_stdout).ok_or_else(|| {
         anyhow::anyhow!(
-            "Unable to extract version information from {}.\n\
+            "Unable to extract version information from \"{}\".\n\
```
ascjones, line 499: Could us single quotes to avoid the escaped double quotes

ascjones (approved):

ascjones:
This might need a `0.10.1` release

src/cmd/build.rs
cmichi: Argh! Maybe we should consider adding a clippy rule for `unwrap()`.

PR #250, opened 2021-04-06
Bump scale codec version

austinabell:
for https://github.com/paritytech/ink/pull/754

There are a bunch of outdated references that already existed (parity-scale-codec 1.3.6 for example), is there a reason for this?


ascjones (approved):

ascjones:
> There are a bunch of outdated references that already existed (parity-scale-codec 1.3.6 for example),

Must be because of old substrate dependencies, but they should be gated behind the `extrinsics` feature.

PR #251, opened 2021-04-07
Remove support for `binaryen-as-dependency`

cmichi:
It caused only problems and complicated things.

I removed it from our CI as well: https://github.com/paritytech/scripts/pull/285. We now install `binaryen` as a system package in our docker image (this was not possible a couple weeks ago).

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -14,18 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
 
-use regex::Regex;
-use std::{convert::TryFrom, ffi::OsStr, fs::metadata, path::PathBuf};
-
-#[cfg(feature = "binaryen-as-dependency")]
-use std::{
-    fs::File,
-    io::{Read, Write},
-};
-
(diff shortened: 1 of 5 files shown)


src/cmd/build.rs
```diff
@@ -337,62 +334,13 @@ fn optimize_wasm(
...
-
 /// Optimizes the Wasm supplied as `crate_metadata.dest_wasm` using
 /// the `wasm-opt` binary.
 ///
 /// The supplied `optimization_level` denotes the number of optimization passes,
 /// resulting in potentially a lot of time spent optimizing.
 ///
 /// If successful, the optimized wasm is written to `dest_optimized`.
-#[cfg(not(feature = "binaryen-as-dependency"))]
 fn do_optimization(
```
ascjones, line 344: Might be worth moving `do_optimization` back into `optimize_wasm` now that there are not 2 versions of it.

cmichi: I would prefer keeping it compartmentalized since `do_optimization` is already quite long and IMHO the code is more readable this way.

ascjones (approved):
LGTM

ascjones:
If you see this in time, please also add this to the `Unreleased` section of the `CHANGELOG`

Issue #36, opened 2020-02-11
Running on Windows


ascjones:
Haven't tested this on Windows yet. Needs trying out and any issues fixing. 

Also worth exploring adding a CI task.

PR #261, opened 2021-04-19
Add CI for Windows

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/36.

Took inspiration [from here](https://github.com/paritytech/libsecp256k1/blob/master/.github/workflows/rust.yml).

The Windows run currently takes ~9 minutes.

While implementing this, I found the bug which caused some people asking in our channels why they got this error when building a contract under Windows:
```
[3/5] Optimizing wasm file
ERROR: The workspace root package should be a workspace member
```
It was due to a comparison of a non-canonicalized path with a canonicalized one.
```
workspace_root (non-canonicalized): "D:\\a\\cargo-contract\\cargo-contract\\foobar"
package_path (canonicalized):       "\\\\?\\D:\\a\\cargo-contract\\cargo-contract\\foobar"
comparison: Some("D:\\a\\cargo-contract\\cargo-contract\\foobar") == Some("\\\\?\\D:\\a\\cargo-contract\\cargo-contract\\foobar")
```
The `\\\\?` prefix appears only under Windows ‒ there a path with this prefix is the correct canonical path, this allows Windows-specific "long" paths ([more info](https://stackoverflow.com/a/41233992)).

I'm not sure if there might be more places in our code which could cause issues with this path discrepancy, FWIW tests are green.

I can do a new release once we have merged this.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,4 +1,4 @@
-name:                 continuous-intergration/windows
+name: continuous-intergration/windows
 
 on:
   pull_request:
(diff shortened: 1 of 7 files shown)


.github/workflows/windows.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name:          Install sudo for windows #https://github.com/actions/virtual-environments/issues/572
+        run: choco install sudo
+
+      - name:          Set cache_hash ENV and prepare cache dir
+        run:           |
+          echo "cache_hash=${{ runner.os }}-${{ steps.toolchain.outputs.rustc_hash }}-${{ matrix.compiler }}-${{ hashFiles('**/Cargo.toml') }}" >> "$GITHUB_ENV"
+          sudo chmod -R a+w $HOME/.cargo
+        shell:         bash
+
+      - name: Cache cargo registry
```
ascjones, line 35: Indentations are not matching with some other steps which are "aligned"

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fixed `ERROR: The workspace root package should be a workspace member` when building a contract
```
ascjones, line 10: :tada:

ascjones (approved):
LGTM, CI stuff needs a double check from @TriplEight though.

.github/workflows/windows.yml
cmichi: Argh, I just applied Intellij's auto-formatting and pushed the result.

TriplEight (approved):
LGTM.
You'll drastically improve caching efficiency if you'll check some of your duplicated dependencies with different versions and lead them to a common denominator.

cmichi:
> You'll drastically improve caching efficiency if you'll check some of your duplicated dependencies with different versions and lead them to a common denominator.

You mean the duplicates that show up if `cargo tree -d` is run? I recently looked through it, but it's third party dependencies which use different versions of the same lib.

TriplEight:
Yeah, kind of. Then there's nothing we can do :(

Issue #234, opened 2021-03-24
Flaky CI on `master`


cmichi:
The test `cmd::metadata::tests::generate_metadata` succeeds sometimes and fails at other times.

These are runs for the same commit:

* Success: https://gitlab.parity.io/parity/cargo-contract/-/jobs/864147
* Failure: https://gitlab.parity.io/parity/cargo-contract/-/jobs/864124

A re-run usually succeeds at some point. 

It fails at an assertion where the hash of the generated Wasm on the filesystem is compared against the hash written in the metadata `source.hash` field.

ascjones:
Closed by #263

PR #263, opened 2021-04-19
Fix flaky CI test

cmichi:
It's hunting season for https://github.com/paritytech/cargo-contract/issues/234 🏹!

I found it!

The issue was that we use
```
CARGO_TARGET_DIR: "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
```
in our CI configuration. This made all tests build to the same `target` directory. Since we have a number of tests which build a contract called `new_project` this made our `tmp_dir`'s obsolete and resulted in overlapping contracts build to e.g. `/ci-cache/cargo-contract/targets/263/test/ink/new_project.contract`.

The way I fixed it now is by appending a hash of a contract's `manifest_path` to the `target_directory` (for tests this path contains the `tmp_dir` path). This results in the paths now being e.g. `/ci-cache/cargo-contract/targets/263/test/0xd82f746d702f636172676f2d636f6e74726163742e746573742e5479444c634b2f6e65775f70726f6a6563742f436172676f2e746f6d6c/ink/new_project.contract`.

I'm not super-happy with that solution, since it makes caching obsolete for the contract build in those particular tests (since the hash is unique each time a test is run, due to `tmp_dir` being included in it).

I have a better idea, but it's a bit more involved: 
* We could use the full test path (e.g. `cmd::metadata::tests::generate_metadata`) for the hash instead.
* This would be achieved by writing a proc. macro (could also be called `#[test]`), which wraps around the Rust `#[test]` macro.
* The macro would then extract the `module_path!` and `function_name!` and put those into a thread-local variable.
* This thread-local variable would be used for hashing.

The disadvantage is that we would require something like `use crate::test;` for the macro to become available for tests. But we would then again have proper caching.

But: The CI hasn't really slowed down with the current fix in this PR ‒ a run of the `test` stage still takes 4-5 minutes. So maybe the current fix is already good enough. Though it fills up the cache and I'm not sure when it is cleared.

Wdyt?

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -127,7 +127,8 @@ pub mod tests {
     ///
     /// We typically use `with_tmp_dir` to generate temporary folders to build contracts
     /// in. But for caching purposes our CI uses `CARGO_TARGET_DIR` to overwrite the
-    /// target directory of any contract build.
+    /// target directory of any contract build -- it is set to a fixed cache directory
+    /// instead.
     /// This poses a problem since we still want to ensure that each test builds to its
     /// own, unique target directory -- without interfering with the target directory of
     /// other tests. In the past this has been a problem when a test tried to create a
@@ -137,6 +138,9 @@ pub mod tests {
     /// The fix we decided on is to append a unique number to each contract name which
     /// is created. This `COUNTER` provides a global counter which is accessed by each test
     /// (in each thread) to get the current `COUNTER` number and increase it afterwards.
+    ///
+    /// We decided to go for this counter instead of hashing (with e.g. the temp dir) to
+    /// prevent an infinite number of contract artifacts being created in the cache directory.
     static COUNTER: AtomicU32 = AtomicU32::new(0);
 
     /// Creates a new contract into a temporary directory. The contract's
--- a/src/util.rs
+++ b/src/util.rs
@@ -123,7 +123,20 @@ pub mod tests {
         f(tmp_dir.path()).expect("Error executing test with tmp dir")
     }
 
-    /// Counter to generate unique project names in `with_new_contract_project`.
+    /// Global counter to generate unique contract names in `with_new_contract_project`.
+    ///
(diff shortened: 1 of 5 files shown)


ascjones:
An alternative solution would be to add an *additional* `with_new_contract_project` function which auto generates a unique name for the contract project and maybe runs `new::exec` expecting success, and passing the resulting `ManifestPath` to the caller. 

The unique name could either be a hash of the tmp dir as you have done or you could even use an atomic counter and append that to the name e.g. `new_project1`, `new_project2` which would mean unique names within a CI run, but would get overwritten on new runs rather than creating endless new artifacts.

src/util.rs
```diff
@@ -118,4 +122,26 @@ pub mod tests {
...
+
+    /// Creates a new contract into a temporary directory. The contract's
+    /// `ManifestPath` is passed into `f`.
+    pub fn with_new_contract_project<F>(f: F)
+    where
+        F: FnOnce(ManifestPath) -> anyhow::Result<()>,
+    {
+        with_tmp_dir(|tmp_dir| {
+            let unique_name = format!("new_project_{}", COUNTER.load(Ordering::SeqCst));
+            COUNTER.fetch_add(1, Ordering::SeqCst);
```
ascjones, line 137: Could just use the result from `fetch_add`, instead of `COUNTER.load` above

src/util.rs
```diff
@@ -118,4 +122,26 @@ pub mod tests {
         // catch test panics in order to clean up temp dir which will be very large
         f(tmp_dir.path()).expect("Error executing test with tmp dir")
     }
+
+    /// Counter to generate unique project names in `with_new_contract_project`.
```
ascjones, line 126: It's worth mentioning why we use this method for generating unique names.

ascjones (approved):
LGTM

Issue #272, opened 2021-05-12
wasm-opt version is 91, but we require a version >= 99 (Not available on Ubuntu 20.04 LTS??)


orro3790:
Hello,

I followed the steps in the [INK tutorial](https://substrate.dev/substrate-contracts-workshop/#/0/building-your-contract), and right away I'm getting the following error when trying to execute:

`cargo +nightly contract build`

> ERROR: Your wasm-opt version is 91, but we require a version >= 99.

Is the tutorial outdated, or is this error on my part? :S

Running Ubuntu 20.04 LTS (Focal Fossa).. is it possible to upg to >= 99?

`apt-get install binaryen `  ---> installs v. 91, can't seem to find way to upg. :\

cmichi:
@orro3790 There is only an outdated version of the `binaryen` package available in your package manager. The best way forward is to download a recent binary release directly [from here](https://github.com/WebAssembly/binaryen/releases) and then make sure that the `wasm-opt` file from that release is in your `PATH`.

It's good that you created an issue for this, I'll add the above as a note when `cargo-contract` errors because of an outdated `binaryen`.

orro3790:
Thanks! This might be a really dumb question, but I've downloaded the latest version of binaryen as per your link, and moved the files to their respective usr/bin, usr/include, and usr/lib64 folders. But how exactly do I make sure that the wasm-opt file from that release is in my PATH?

My unix chops need work, but I know if I'm struggling with following the tutorial due to this problem, there are likely many other newbies that will face the same problem.

I really appreciate your help cmichi! Your response was so quick too!! <3

cmichi:
@orro3790  You could do something like `export PATH=/path/to/binaryen-version_101/bin/:$PATH` before executing the cargo +nightly contract build.

If you execute `wasm-opt --version` you should get a version number >= 99 then.

orro3790:
@cmichi You're the man!! Works like a charm! I hope this thread helps my fellow newbies as much as it helped me!

nodi-andy:
if someone encounters same problem, see here: https://docs.porta.network/developer-guide/smart-contracts/install-wasm-opt-version-101

PR #274, opened 2021-05-12
Suggest `binaryen` installation from GitHub on outdated version

cmichi:
This came up more often recently. Latest in https://github.com/paritytech/cargo-contract/issues/272, but also in the public channel.


ascjones (approved):
LGTM but there is a test which checks that string failing. Should probably change the test to not match on the string exactly,

Issue #221, opened 2021-03-09
Compilation fails when lib.name and package.name are different


ii-ii-ii:
When I set lib.name to be different from package.name, I get this error.
```
ERROR: Loading original wasm file 'erc20/contracts/target/ink/wasm32-unknown-unknown/release/erc20.wasm'

Caused by:
    Can't read from the file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
An unexpected error occurred:
```

I want to know if the compiled wasm file name is [package.name].wasm or [lib.name].wasm.

ascjones:
I believe the compiled wasm file name is <lib.name>.wasm, but we expect it to be <package.name>.wasm. This is a bug we should fix, in the meantime they will have to be the same.

ii-ii-ii:
> I believe the compiled wasm file name is <lib.name>.wasm, but we expect it to be <package.name>.wasm. This is a bug we should fix, in the meantime they will have to be the same.

Thanks for your reply. One more question, if package.name is erc20-fixed and lib.name is erc20_fixed, it still compiles successfully, but the wasm file name is erc20_fixed.wasm. After fixing this bug, what kind of filename will I get? erc20_fixed.wasm or erc20-fixed.wasm?

ascjones:
It will be the same as in the lib name (`erc20_fixed`), as it is now.

PR #277, opened 2021-05-21
Always use `lib.name` for Wasm output file name

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/221.

### DIFF
--- a/src/crate_metadata.rs
+++ b/src/crate_metadata.rs
@@ -28,7 +28,7 @@ use url::Url;
 pub struct CrateMetadata {
     pub manifest_path: ManifestPath,
     pub cargo_meta: cargo_metadata::Metadata,
-    pub package_name: String,
+    pub contract_artifact_name: String,
     pub root_package: Package,
     pub original_wasm: PathBuf,
     pub dest_wasm: PathBuf,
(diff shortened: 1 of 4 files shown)


src/crate_metadata.rs
```diff
@@ -58,12 +65,12 @@ impl CrateMetadata {
...
         original_wasm.push("wasm32-unknown-unknown");
         original_wasm.push("release");
-        original_wasm.push(package_name.clone());
+        original_wasm.push(lib_name.clone());
         original_wasm.set_extension("wasm");
 
         // {target_dir}/{package_name}.wasm
         let mut dest_wasm = target_directory.clone();
-        dest_wasm.push(package_name.clone());
+        dest_wasm.push(lib_name.clone());
```
ascjones, line 73: Note that `package_name` will still be used for the `contract` bundle, which could be confusing when viewing the `target` dir if they are different. 

We could just replace the `package_name` field `CrateMetadata` with something like `contract_artifact_name`, and populate with the `lib` name.

cmichi: Good point, I updated the PR.

ascjones (approved):
LGTM

Issue #280, opened 2021-05-26
Implement `cargo contract test`


cmichi:
- [ ] Spawn a child process which executes `cargo +nightly test` for the contract
- [ ] Return the piped output of the child process
- [ ] Implement appropriate error handling
- [ ] Support `--manifest-path`

cmichi:
Closed via https://github.com/paritytech/cargo-contract/pull/283.

PR #283, opened 2021-05-27
impl: test sub cmd with support for --manifest-path (#280)

chiefbiiko:
drafts `cargo contract test [FLAGS] [--manifest-path]`

in this first impl stdout is uncolored

let me know if you care much for the colors

### DIFF
--- a/src/cmd/test.rs
+++ b/src/cmd/test.rs
@@ -1,0 +1,94 @@
+// Copyright 2018-2021 Parity Technologies (UK) Ltd.
+// This file is part of cargo-contract.
+//
+// cargo-contract is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// cargo-contract is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::{maybe_println, util, workspace::ManifestPath, Verbosity, VerbosityFlags};
+use anyhow::Result;
+use colored::Colorize;
+use std::{convert::TryFrom, path::PathBuf};
+use structopt::StructOpt;
+
+/// Executes smart-contract tests off-chain by delegating to `cargo test`.
+#[derive(Debug, StructOpt)]
+#[structopt(name = "test")]
+pub struct TestCommand {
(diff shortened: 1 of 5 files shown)


Robbepop (commented):
What precisely is the win of having `cargo contract test` over just using `cargo test` as it can be done right now?

cmichi:
@Robbepop At the moment `cargo contract --help` shows the sub-command as:
```
test     Test the smart contract off-chain
```
If one invokes `cargo contract test` currently this results in an `unimplemented!` panic though, which is why people were irritated and started creating issues about it in this repository.

So we could either remove the sub-command or just hand over to `cargo test`. After talking to @ascjones we concluded that it's probably best to just spawn a sub-process with `cargo test` until we have further ideas what to do with the sub-command. 

What's your opinion?

Robbepop:
> @Robbepop At the moment `cargo contract --help` shows the sub-command as:
> 
> ```
> test     Test the smart contract off-chain
> ```
> 
> If one invokes `cargo contract test` currently this results in an `unimplemented!` panic though, which is why people were irritated and started creating issues about it in this repository.
> 
> So we could either remove the sub-command or just hand over to `cargo test`. After talking to @ascjones we concluded that it's probably best to just spawn a sub-process with `cargo test` until we have further ideas what to do with the sub-command (#280).
> 
> What's your opinion?

I currently do not see a major reason but it might help against confusion. So yeah let's do it the proposed way.

src/cmd/test.rs
```diff
@@ -0,0 +1,88 @@
...
+#[cfg(test)]
+mod tests_ci_only {
+    use crate::{util::tests::with_new_contract_project, Verbosity};
+
+    #[test]
+    fn passing_tests_yield_stdout() {
+        with_new_contract_project(|manifest_path| {
+            let res = super::execute(&manifest_path, Verbosity::Default).expect("test failed");
+
+            assert!(res.stdout.len() > 0);
```
cmichi, line 83: Could you add another assertions which checks for something like `test result: ok. 2 passed; 0 failed; 0 ignored`?

This way we can then be sure that there was no other unrelated failure the cause of `stdout.len() > 0`.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.13.0] - 2021-05-27
+
+### Added
+- Convenient offchain testing through `cargo contract test`
```
cmichi, line 12:
```suggestion
- Convenient off-chain testing through `cargo contract test`
```

cmichi (changes requested):
Thanks for the PR, it looks good!

I've got just some minor things. We use the writing `off-chain` everywhere, hence the suggestion of keeping it consistent.

chiefbiiko:
thanks for the review
i have included your suggestions now

cmichi (approved):
Thanks for the PR!

Issue #284, opened 2021-05-27
Enable `ink-debug` feature for enabling debug messages


ascjones:
In https://github.com/paritytech/ink/pull/792 we introduce the `ink-debug` feature, which needs to be enabled in order for `debug_print!` messages to be sent to the supervisor and printed to the console/sent back to the calling client. 

So we should allow the enabling/disabling of this feature by introducing the `--debug` and `--release` builds or some other way to control it. It may not assume that the contract has that feature added to its `Cargo.toml`, so could add as necessary during the build.

Perhaps in addition we should allow manual control of features passed to the underlying `build` command in case the user wants full control.

PR #298, opened 2021-07-08
Implement `BuildMode`

cmichi:
Closes https://github.com/paritytech/cargo-contract/issues/284.

Kinda blocked until we have a release of ink! `3.0.0-rc4`, since the `ink_env/ink-debug` feature is newer than the currently released `3.0.0-rc3`. The tests are failing because they use the published ink! crates.

This PR adds a new CLI flag `--release`, which mirrors the behavior of `cargo` ‒ compiling the artifact in release mode is not done by default.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -48,24 +48,14 @@ pub struct BuildCommand {
     /// Path to the Cargo.toml of the contract to build
     #[structopt(long, parse(from_os_str))]
     manifest_path: Option<PathBuf>,
-    /// Which mode to build the contract in.
+    /// By default the contract is compiled with debug functionality
+    /// included. This enables the contract to output debug messages,
+    /// but increases the contract size and the amount of gas used.
     ///
-    /// - `debug`: The contract will be compiled with debug functionality
-    ///   included. This enables the contract to output debug messages.
-    ///
-    ///   This increases the contract size and the used gas. A production
-    ///   contract should never be build in debug mode!
-    ///
-    /// - `release`: No debug functionality is compiled into the contract.
-    ///
-    /// - The default value is `debug`.
-    #[structopt(
-        long = "mode",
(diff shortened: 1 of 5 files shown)


ascjones (approved):
LGTM, good to go once `ink!` is released

src/cmd/build.rs
```diff
@@ -183,14 +199,18 @@ fn exec_cargo_for_wasm_target(
...
     let cargo_build = |manifest_path: &ManifestPath| {
         let target_dir = &crate_metadata.target_directory;
-        let args = [
+        let target_dir = format!("--target-dir={}", target_dir.to_string_lossy());
+        let mut args = vec![
             "--target=wasm32-unknown-unknown",
             "-Zbuild-std",
             "-Zbuild-std-features=panic_immediate_abort",
             "--no-default-features",
             "--release",
```
HCastano, line 208: Shouldn't this only be added depending on the `BuildMode` now?

src/cmd/build.rs
```diff
@@ -396,6 +416,7 @@ fn do_optimization(
                 err
             )
         })?;
+    // TODO
```
HCastano, line 419: Is this a leftover, or are you missing some functionality here?

src/cmd/build.rs
ascjones: I think we'd still want to build the wasm in release mode for optimal size, after all these blobs are still deployed to the chain. That said, I've not tested what the difference is building without `--release`.

src/cmd/build.rs
cmichi: Good catch, thx! I removed it and added a commit which makes our CI report this in the future (as we do in ink!).

HCastano (approved):
Looks good!

Now we just need `rc4` to be pushed so the CI can pass

PR #302, opened 2021-07-19
Add `--keep-symbols` flag

athei:
In order to properly analyze a contract binary we need to keep the wasm name section alive so that we can see what is going on. Using the non post processed wasm file is of no use because we are interested in the fully optimized binary when doing an analysis. To that end I added a new `--keep-symbols` flag to the `build` command which prevents `wasm-opt` from removing the debug section.

I needed to make some other changes to make this work, too. Most notable I removed the call to `pwasm_utils::optimize` which unconditionally removes all custom sections. We don't need this anyways because `wasm-opt` does the same optimizations and more. It sufficed to remove all exports from the module (but `call` and `deploy`) to have the same result without `pwasm_utils::optimize`. I argue that having one less pass involved is a good thing.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -288,14 +288,20 @@ fn strip_exports(module: &mut Module) {
     }
 }
 
+/// Load and parse a wasm file from disk.
+fn load_module<P: AsRef<Path>>(path: P) -> Result<Module> {
+    let path = path.as_ref();
+    parity_wasm::deserialize_file(path).context(format!(
+        "Loading of wasm module at '{}' failed",
+        path.display(),
+    ))
+}
+
 /// Performs required post-processing steps on the wasm artifact.
 fn post_process_wasm(crate_metadata: &CrateMetadata) -> Result<()> {
     // Deserialize wasm module from a file.
     let mut module =
-        parity_wasm::deserialize_file(&crate_metadata.original_wasm).context(format!(
-            "Loading original wasm file '{}'",
-            crate_metadata.original_wasm.display()
-        ))?;
+        load_module(&crate_metadata.original_wasm).context("Loading of original wasm failed")?;
 
     strip_exports(&mut module);
     ensure_maximum_memory_pages(&mut module, MAX_MEMORY_PAGES)?;
@@ -621,7 +627,7 @@ pub(crate) fn execute(
 mod tests_ci_only {
     use super::{assert_compatible_ink_dependencies, check_wasm_opt_version_compatibility};
     use crate::{
-        cmd::BuildCommand,
+        cmd::{build::load_module, BuildCommand},
         util::tests::{with_new_contract_project, with_tmp_dir},
         workspace::Manifest,
         BuildArtifacts, ManifestPath, OptimizationPasses, UnstableFlags, UnstableOptions,
@@ -650,6 +656,13 @@ mod tests_ci_only {
         manifest
             .write(&manifest_path)
             .expect("writing manifest failed");
+    }
+
+    fn has_debug_symbols<P: AsRef<Path>>(p: P) -> bool {
+        load_module(p)
+            .unwrap()
+            .custom_sections()
+            .any(|e| e.name() == "name")
     }
 
     /// Creates an executable `wasm-opt-mocked` file which outputs
@@ -703,6 +716,10 @@ mod tests_ci_only {
 
             // our optimized contract template should always be below 3k.
             assert!(optimized_size < 3.0);
+
+            // we specified that debug symbols should be removed
+            // original code should have some but the optimized version should have them removed
+            assert!(!has_debug_symbols(&res.dest_wasm.unwrap()));
 
             Ok(())
         })
@@ -957,6 +974,26 @@ mod tests_ci_only {
                     .file_name(),
                 Some(OsStr::new("some_lib_name.wasm"))
             );
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn keep_debug_symbols() {
+        with_new_contract_project(|manifest_path| {
+            let res = super::execute(
+                &manifest_path,
+                Verbosity::Default,
+                BuildArtifacts::CodeOnly,
+                UnstableFlags::default(),
+                OptimizationPasses::default(),
+                true,
+            )
+            .expect("build failed");
+
+            // we specified that debug symbols should be kept
+            assert!(has_debug_symbols(&res.dest_wasm.unwrap()));
 
             Ok(())
         })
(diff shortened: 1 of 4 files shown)


src/cmd/build.rs
```diff
@@ -86,8 +86,13 @@ pub struct BuildCommand {
     /// - It is possible to define the number of optimization passes in the
     ///   `[package.metadata.contract]` of your `Cargo.toml` as e.g. `optimization-passes = "3"`.
     ///   The CLI argument always takes precedence over the profile value.
-    #[structopt(long = "optimization-passes")]
+    #[structopt(long)]
     optimization_passes: Option<OptimizationPasses>,
+    /// Do not remove symbols (wasm name section) when optimizing.
+    ///
+    /// This is useful if one want to analyze or debug the optimized binary.
```
cmichi, line 93:
```suggestion
    /// Do not remove symbols (Wasm name section) when optimizing.
    ///
    /// This is useful if one wants to analyze or debug the optimized binary.
```

src/cmd/build.rs
```diff
@@ -260,13 +267,25 @@ fn ensure_maximum_memory_pages(module: &mut Module, maximum_allowed_pages: u32)
 /// Strips all custom sections.
 ///
 /// Presently all custom sections are not required so they can be stripped safely.
+/// The name section is already stripped by wasm-opt.
```
cmichi, line 270:
```suggestion
/// The name section is already stripped by `wasm-opt`.
```

src/cmd/build.rs
```diff
@@ -260,13 +267,25 @@ fn ensure_maximum_memory_pages(module: &mut Module, maximum_allowed_pages: u32)
...
+    module.sections_mut().retain(|section| match section {
+        Section::Reloc(_) => false,
+        Section::Custom(custom) if custom.name() != "name" => false,
+        _ => true,
+    })
+}
+
+/// A contract should export nothing but the "call" and "deploy" functions.
+///
+/// Any elements referenced by these exports become orphaned and are removed by wasm-opt.
```
cmichi, line 281:
```suggestion
/// Any elements referenced by these exports become orphaned and are removed by `wasm-opt`.
```

athei:
There is no `--debug` flag and hence it cannot be the default. The default for `build` is to build for production. `check` is for development.

cmichi:
> There is no `--debug` flag and hence it cannot be the default. The default for `build` is to build for production. `check` is for development.

I think what Robin means that in a recent PR we decided against a `cargo contract build --debug` flag and instead now mirror `cargo` behavior of building for debug by default. We instead added a `cargo contract build --release` flag which leaves debug info out.

By debug info we currently only have the `ink-debug` flag, which enables printing debug statements and hence some overhead needs to be compiled into the contract to support that.

I think the misunderstanding here is that you are probably referring to the fact that we always build contracts as `--release` and also still run `wasm-opt` on them. This is required so that contracts can  even be deployed.

So the distinction I think Robin wants to make is that we have a debug and a release mode for `cargo-contract` and that you could just leave the symbols in by default and only omit them on `cargo contract build --release`.

athei:
But nothing of those things is a reality on `master`, right?

cmichi:
> But nothing of those things is a reality on `master`, right? Because I don't see a `--release` flag anywhere.

https://github.com/paritytech/cargo-contract/pull/298

It's approved, but not yet merged because we need rc4 to be released first.

athei:
Thanks. I checked the PR. I disagree. `--keep-symbols` is orthogonal to `--debug`. I want to analyze the optimized binary as it would be deployed on chain (that might or might not include a build without the ink! debug features).

Robbepop:
If this is the case then please rename the flag to `--keep-debug-symbols` so that it is clear by name that this is for debug inspection.

atenjin:
we have create a pr in post https://github.com/paritytech/cargo-contract/pull/131
but we close it for no response .... 
maybe --debug is a good choose for many cases, not only in this situation. we also need those information which contains debug symble in name section

src/cmd/build.rs
```diff
@@ -260,31 +267,43 @@ fn ensure_maximum_memory_pages(module: &mut Module, maximum_allowed_pages: u32)
...
+    module.sections_mut().retain(|section| match section {
+        Section::Reloc(_) => false,
+        Section::Custom(custom) if custom.name() != "name" => false,
+        _ => true,
+    })
+}
+
+/// A contract should export nothing but the "call" and "deploy" functions.
+///
+/// Any elements referenced by these exports become orphaned and are removed by `wasm-opt`.
```
ascjones, line 281:
```suggestion
/// Any elements not referenced by these exports become orphaned and are removed by `wasm-opt`.
```

src/cmd/build.rs
```diff
@@ -771,24 +807,21 @@ mod tests_ci_only {
...
+            // The size does not exactly match the original size even without optimization
+            // passed because there is still some post processing happening.
+            let size_diff = optimization.original_size - optimization.optimized_size;
             assert!(
-                optimized_size < original_size,
-                "The optimized size DOES NOT {:?} differ from the original size {:?}",
-                optimized_size,
-                original_size
+                size_diff > (optimization.original_size / 2.0),
+                "The optimized size savings are to small: {}",
```
cmichi, line 823:
```suggestion
                "The optimized size savings are too small: {}",
```

src/cmd/build.rs
```diff
@@ -731,26 +771,22 @@ mod tests_ci_only {
 
                 // we choose zero optimization passes as the "cli" parameter
                 optimization_passes: Some(OptimizationPasses::Zero),
+                keep_debug_symbols: false,
             };
 
-            // when
```
cmichi, line 736: Why did you remove the `when` and `then`? The `given` is still there and [I think that it's generally a good structure for tests](https://martinfowler.com/bliki/GivenWhenThen.html).

ascjones (approved):
LGTM

src/cmd/build.rs
athei: I think it got lost somehow. I went back and forth on this part until I had it working. I will restore it.

cmichi (approved):
Thanks! `cargo fmt` fails right now, but besides that it LGTM.

athei:
> Thanks! `cargo fmt` fails right now, but besides that it LGTM.

Yeah but because of the code. Just some intermittent CI failures. Saw this multiple times here:
```
Error during execution of `cargo metadata`: error: failed to download `env_logger v0.9.0`
```

athei:
Merging this. The CI is broken (some caching issue, again). I ran `cargo fmt` locally.

PR #303, opened 2021-07-19
Fix tests on macOS

athei:
One of the tests fails to pass on macOS. The reason is that the temporary pass used is a symlink and one side of the equality used path canonicalization while the other does not.

This PR fixes that by adding canonicalization to the test only `with_tmp_dir` function.


HCastano (approved):
Tested it locally and it looks good.

We should consider adding some macOS runners to the CI to catch stuff like this (see https://github.com/paritytech/cargo-contract/issues/304).

Issue #110, opened 2020-11-18
Experiment with some wasm-opt options


Robbepop:
We currently are not using `wasm-opt`'s `--strip-debug --vacuum --dce` options. They might be useful to us to reduce the resulting `.wasm` files. Experimentation needed.

cmichi:
In that regard experimenting with [these](https://github.com/paritytech/cargo-contract/blob/master/src/cmd/build.rs#L187-L190) binaryen options could also have some benefits:

Currently:
```
    let codegen_config = binaryen::CodegenConfig {
        // execute -O3 optimization passes (spends potentially a lot of time optimizing)
        optimization_level: 3,
        // the default
        shrink_level: 1,
        // the default
        debug_info: false,
    };
```

HCastano:
I've been playing around a bit with some of the optimizations options using some (but not
all) of our example contracts.

For reference, the original sizes with `ink!` on [`hc-bump-allocator`](https://github.com/paritytech/ink/tree/8e84f341960c43676ad0fc22df554b5f49b80d0e) and `cargo-contract` on [`master`](https://github.com/paritytech/cargo-contract/tree/d79d07baae4ea5b84c3bde12cb697ed58a2613d1)
are in the table below.

| Contract | Original | Optimized |
--- | --- | --- |
| erc20          | 51.8K | 30.5K |
| trait-erc20    | 52.3K | 31.0K |
| erc1155        | 65.2K | 45.3K |
| erc721         | 68.3K | 41.1K |
| dns            | 44.2K | 25.3K |
| multisig_plain | 86.2K | 53.6K |


1. Setting `cargo-contract` to `--optimization-passes s`
This saves us ~0.2-0.3K.

2. Setting `cargo-contract` to `--optimization-passes z`
Same as (1).

3. Adding `--vaccum` option to `wasm-opt`
This increases the size by ~0.2-0.3K.

4. Adding `--strip-debug` option to `wasm-opt`
Same as (3).

5. Adding `--dae-optimizing` option to `wasm-opt`
Same as (3).

I think right now changing the opt level default from `-O3` to at least `-Os` is the most
sensible thing to do.

athei:
There are **a lot** of optimization options. I have no idea which are on by default and which are not. Would be nice to find that out. I guess the source code should reveal this information.

For example, `--strip-debug` is on by default. To keep the debug symbols `-g` is needed.

cmichi:
[...]

PR #305, opened 2021-07-19
Change default optimizations pass to focus on code size

HCastano:
This PR changes the default optimization option passed to `wasm-opt` from `-O3` to ~~`-Os`~~`-Oz`.
This saves us around 0.2-0.3K according to [these rough benchmarks](https://github.com/paritytech/cargo-contract/issues/110#issuecomment-882844120).

~~The default wasn't set to `-Oz` as I saw no noticeable size improvements, and it just
takes more time to run. However, you could also argue that since contracts are in general
quite small the extra time for `-Oz` is negligible so we should enable it anyways.~~

~~While I could be convinced of this, I'm not sure what hardware other people build on, and
the negative impact the change to `-Oz` could have for them.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Change default optimizations pass to focus on code size - [#305](https://github.com/paritytech/cargo-contract/pull/305)
+
 ## [0.12.1] - 2021-04-25
 
 ### Added
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -81,7 +81,7 @@ pub struct BuildCommand {
     ///
     /// - `z`, execute default optimization passes, super-focusing on code size
     ///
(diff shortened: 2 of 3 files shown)


athei:
We should invest everything in making contracts as small as possible. Any contract above 50KB wouldn't be viable to run on a parachain anyways. So the optimization time should be negligible.

athei (approved):

cmichi (changes requested):
Thank you!

Could you add a `### Changed` subsection to the `Unreleased` section of our `CHANGELOG.md`: https://github.com/paritytech/cargo-contract/blob/master/CHANGELOG.md ?

HCastano:
> Thank you!
> 
> Could you add a `### Changed` subsection to the `Unreleased` section of our `CHANGELOG.md`: https://github.com/paritytech/cargo-contract/blob/master/CHANGELOG.md ?

I like that a `CHANGELOG` is kept. However, there's gotta be a way to automate this, no? Maybe @s3krit can share some of his automation secrets 👀

cmichi (approved):

PR #321, opened 2021-08-03
Fix failing test for Windows `master` CI

cmichi:


src/cmd/build.rs
```diff
@@ -201,7 +201,7 @@ fn exec_cargo_for_wasm_target(
     // Currently will override user defined RUSTFLAGS from .cargo/config. See https://github.com/paritytech/cargo-contract/issues/98.
     std::env::set_var(
         "RUSTFLAGS",
-        "-C link-arg=-z -C link-arg=stack-size=65536 -C link-arg=--import-memory",
+        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory",
```
cmichi, line 204: For some reason (which I don't fully understand) there is an issue with the Windows CI currently failing on `master` if the `-z` is passed separately of the `stack-size` argument, which actually belongs to it. Since both versions should be equal I would change it to this.

src/cmd/build.rs
```diff
@@ -940,8 +940,13 @@ mod tests_ci_only {
...
             assert!(res.is_err());
-            assert!(format!("{:?}", res)
-                .starts_with("Err(Your wasm-opt version is 98, but we require a version >= 99."));
+            assert!(
+                format!("{:?}", res).starts_with(
+                    "Err(Your wasm-opt version is 98, but we require a version >= 99."
+                ),
+                "Expected a different output, found {:?}",
+                res
+            );
```
cmichi, line 949: This test is flaky sometimes, if the CI is re-run it succeeds. I added output of the `res` to have some more info when if fails spuriously again.

ascjones (approved):

PR #322, opened 2021-08-04
CI: force cache pre-population

TriplEight:
Closes: https://github.com/paritytech/ci_cd/issues/166

Let's try your suggestion @cmichi , but I'm pretty sure that the logic there is that it happens only if there's no such a cache dir on the host.

Last time I was revisiting a similar issue I came to the conclusion that this happens because cargo automatically removes some old caches.


cmichi (approved):
🤞

PR #323, opened 2021-08-04
Mention the contracts Docker image in the project README

### LINKED ISSUE #184: Provide docker image for cargo-contract
For substrate runtime, parity uses srtool(https://github.com/paritytech/srtool) to compile wasm. And Solang as well, providing a docker image which contrains newest Solang. (https://github.com/hyperledger-labs/solang/blob/main/Dockerfile)

And we think cargo-contract could also provide such docker image. This docker image could resolve the following things:
1. The wasm compilation target for same ink contract may be different in different computer, that is way substrate runtime needs srtool to generate deterministic result.
2. For windows, many users do not have `cmake` by default, thus, windows users need to parpare `cmake` before install. If using docker, windows users do not need to install cmake.
3. For our 3-rd parties, like RedSpot(https://redspot.patract.io/), if we want to unify different environment, an existed docker image is a good choice.

Thus, we help cargo-contract provides a docker image by default.

HCastano:
Closes #184.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ More relevant links:
 * Step 3: `cargo install --force cargo-contract`
 
 ### Docker Image
+
 If you'd prefer to use Docker instead we have a Docker image [available on the Docker Hub](https://hub.docker.com/r/paritytech/contracts-ci-linux)
 which contains the dependencies needed to build `cargo-contract`, as well as
 a pre-built `cargo-contract` binary which you can just execute.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,12 +45,16 @@ More relevant links:
...
   There's only an old version in your distributions package manager? Just use a 
   [binary release](https://github.com/WebAssembly/binaryen/releases).
-  
+
 * Step 3: `cargo install --force cargo-contract`
 
+### Docker Image
+If you'd prefer to use Docker instead we have a Docker image [available on the Docker Hub](https://hub.docker.com/r/paritytech/contracts-ci-linux)
+which contains the dependencies needed to build `cargo-contract`, as well as
+`cargo-contract` itself.
```
cmichi, line 57:
```suggestion
a pre-built `cargo-contract` binary which you can just execute.
```

cmichi (approved):

PR #325, opened 2021-08-09
Use new dependency resolver for template contract

HCastano:
By default, Cargo's dependency resolver leaks `std` features from `dev-dependencies` into
`no_std` builds. This results in confusing errors messages such as the one in
https://github.com/paritytech/ink/issues/889. 

To avoid issues like this we can use the new(ish) dependency resolver, which doesn't leak
features from `dev-dependencies` into the regular build. You can read more about it [here](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
-- Use new dependency resolver for template contract [#325](https://github.com/paritytech/cargo-contract/pull/325)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Use new dependency resolver for template contract [#325](https://github.com/paritytech/cargo-contract/pull/325)
```
cmichi, line 10:
```suggestion
- Use new dependency resolver for template contract - [#325](https://github.com/paritytech/cargo-contract/pull/325)
```

cmichi (approved):

PR #326, opened 2021-08-12
Do not strip out panic messages in debug builds

athei:
We should keep the panic messages for debug builds because they convey valuable information. Until `Result` based error handling is implemented that is even the only way of reacting to runtime errors (as opposed to programming errors for which panics are usually used).


cmichi (approved):
CI is currently failing, but just because the size of the Wasm blob increased in debug mode now (which is expected). Can you adapt the test?

athei:
Building that test in release mode now. That should fix it.

cmichi (approved):
I'll do a release soonish.

PR #343, opened 2021-09-08
Add information about CI to README

alvicsam:
Modified an image tag in README and also added information how to reproduce CI locally. We have several ci docker images with ready environment which can be used to build/test different projects. (Closes https://github.com/paritytech/ci_cd/issues/113)


HCastano (changes requested):
I don't really see the point of this aside from the example Docker usage. The CI stage info is unnecessarily detailed, especially for the README. Why do you think the CI docs  are worth adding?

alvicsam:
I thought that it might be useful for a developer to have a document describes how to reproduce CI locally. The CI stage info has a lot of details but it's hidden under collapsible section. If you think that this is too much for a readme I can suggest moving it to a separate document and put a link here.

TriplEight:
Yeah, everything in **preparation** is too broad, let's just publish it on https://github.com/paritytech/scripts/blob/master/README.md and link from here.

HCastano (approved):
Linking to the `scripts` repo is more appropriate. 

I also just tried the Docker steps and everything works, so thanks for the PR!

PR #347, opened 2021-09-22
Populate `metadataVersion` field with `ink_metadata` crate version

ascjones:
Instead of being hardcoded in the crate, the `metadata_version` field is populated with the version of the `ink_metadata` crate of the contract.

This will allow consumers (e.g. `polkadot.js` @jacogr) to be able to differentiate between different versions of the `ink` metadata.


src/cmd/build.rs
```diff
@@ -436,7 +436,6 @@ fn do_optimization(
             We use this tool to optimize the size of your contract's Wasm binary.\n\n\
             wasm-opt is part of the binaryen package. You can find detailed\n\
             installation instructions on https://github.com/WebAssembly/binaryen#tools.\n\n\
-
```
cmichi, line 439: Thanks, that warning was driving me nuts, but making a PR for it alone would have been overkill.

ascjones: Small PRs are fine IMO

cmichi (approved):
All good, thanks for creating the PR that fast!

I would hold off with publishing and merging though, until the UI supports the new format as well. I just built `flipper` using your branch here and the UI currently does not accept the contract bundle with the changed `metadataVersion` format.

cmichi:
One of the `cargo test --features test-ci-only` is failing. And could you add an entry to the unreleased section of our release notes?

jacogr:
Catch-22. before this is available the API (and the any UIs that use it), cannot quite support it. 

Or rather, if there is a sample .contract that populates this field correctly, at least the API work can be wrapped up and tested.

cmichi:
>Or rather, if there is a sample .contract that populates this field correctly, at least the API work can be wrapped up and tested.

The first version with scale-info 1.0 will contain `"metadataVersion":"3.0.0-rc6"`, I modified the contract from https://github.com/polkadot-js/api/issues/3983 to contain this value instead.

Here's the updated contract bundle: [flipper.contract.zip](https://github.com/paritytech/cargo-contract/files/7212234/flipper.contract.zip).

jacogr:
Ok. But let us be realistic - the metadata format has nothing to do with the contacts version. 

Since 0.1 there has been no need for a bump. 

In is the same way the Substarte metadata version (v14) is independent of the runerime version. Here the two are the same.

Format != functionality

ascjones:
`ink_metadata` is a separate crate, and although at the moment it's version is synced with the other `ink!` crates, I was assuming that once `3.0.0` is cut we can make it a free floating version? E.g. `ink_lang` might get to version `4` or `5` and still produce metadata version `3`. 

An alternative to using the `ink_metadata` crate version is to do the same like `frame_metadata` and encapsulate the metadata in a top level enum, and just keep the metadata versions is sync with the other crates.

`enum Metadata { V1(MetadataV1), V2(MetadataV2) }`

@Robbepop what are your thoughts on this?

Robbepop:
> `ink_metadata` is a separate crate, and although at the moment it's version is synced with the other `ink!` crates, I was assuming that once `3.0.0` is cut we can make it a free floating version? E.g. `ink_lang` might get to version `4` or `5` and still produce metadata version `3`.
> 
> An alternative to using the `ink_metadata` crate version is to do the same like `frame_metadata` and encapsulate the metadata in a top level enum, and just keep the metadata versions is sync with the other crates.
> 
> `enum Metadata { V1(MetadataV1), V2(MetadataV2) }`
> 
> @Robbepop what are your thoughts on this?

I fully agree with @jacogr .
With respect to your question in particular concerning `ink_metadata` I'd dislike seeing a divergence with other `ink_` crate versions although I definitely get your point.

ascjones:
I also agree, the difference being I was assuming that we could use the metadata crate version to represent the metadata version. But if the `ink_metadata` crate must always be in sync with the others then that won't work obviously.

Robbepop:
> But if the ink_metadata crate must always be in sync with the others then that won't work obviously.

The `ink_metadata` crate version does not need to be in sync with other ink! crate versions, however, I have the feeling that out-of-sync versions would lead to more confusions than it would help.
Maybe `cargo-contract` could transparently print out the currently used metadata version during compilation?

ascjones:
> The ink_metadata crate version does not need to be in sync with other ink! crate versions, however, I have the feeling that out-of-sync versions would lead to more confusions than it would help.

One could argue that adding another layer of versioning is also confusing, since the `ink_metadata` crate version would be decoupled from actual changes in that crate, therefore no longer strictly semver.

ascjones:
Closing in favour of https://github.com/paritytech/cargo-contract/pull/342 and https://github.com/paritytech/ink/pull/845

PR #356, opened 2021-11-08
Offline flag

achimcc:
This PR adds a `--offline` flag which can be used to build contracts without network acces with `cargo contract build --offline`.

The motivation for providing this extra flag is to be able to compile Smart contracts in [ink! Playground](https://github.com/paritytech/ink-playground) within a safe docker environment without direct network access.

Ussing the `--offline` flag for `cargo contract build`, both, cargo build and also the generation of cargo contract metadata become possible without network connection, using cached dependencies only.


ascjones (approved):
LGTM

Issue #98, opened 2020-11-04
Build ignores RUSTFLAGS


athei:
`cargo contract build` ignores flags passed in through the `RUSTFLAGS` environment variable. Variables passed in through the cargo config are also ignored. The reason is likely that they are overridden here:
https://github.com/paritytech/cargo-contract/blob/f8f048fbace9c10bb414fb4da0df91df772a12d8/src/cmd/build.rs#L57-L61

They should be merged with the pre-existing options where pre-existing options take precedence in case of conflicts.

Likely a similar change is necessary for metadata generation.

ltfschoen:
can we close this issues now that [this PR](https://github.com/paritytech/cargo-contract/pull/1124) has been merged that uses `CARGO_ENCODED_RUSTFLAGS` instead of `RUSTFLAGS` in `contract-build` crate?

athei:
I don't know if `cargo` merged those two or if one takes precedence in the presence of both. If the former is true we can close this.

ascjones:
Indeed we would need to confirm which is true, the docs are not enlightening. Even if the former is true the issue will still exist for `CARGO_ENCODED_RUSTFLAGS`.

PR #358, opened 2021-11-10
Use `-Clinker-plugin-lto` flag during Wasm build

xgreenx:
`-Clinker-plugin-lto` reduced the size of my contract
From:
Original wasm size: 68.6K, Optimized: 31.9K

To:
Original wasm size: 67.2K, Optimized: 31.0K

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -249,7 +249,7 @@ fn exec_cargo_for_wasm_target(
     // Currently will override user defined RUSTFLAGS from .cargo/config. See https://github.com/paritytech/cargo-contract/issues/98.
     std::env::set_var(
         "RUSTFLAGS",
-        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory -Clto -Clinker-plugin-lto",
+        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory -Clinker-plugin-lto",
     );
 
     let cargo_build = |manifest_path: &ManifestPath| {
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -249,7 +249,7 @@ fn exec_cargo_for_wasm_target(
     // Currently will override user defined RUSTFLAGS from .cargo/config. See https://github.com/paritytech/cargo-contract/issues/98.
     std::env::set_var(
         "RUSTFLAGS",
-        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory",
+        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory -Clto -Clinker-plugin-lto",
     );
 
     let cargo_build = |manifest_path: &ManifestPath| {


src/cmd/build.rs
```diff
@@ -249,7 +249,7 @@ fn exec_cargo_for_wasm_target(
     // Currently will override user defined RUSTFLAGS from .cargo/config. See https://github.com/paritytech/cargo-contract/issues/98.
     std::env::set_var(
         "RUSTFLAGS",
-        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory",
+        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory -Clto -Clinker-plugin-lto",
```
ascjones, line 252: I believe the `-Clto` will override the release profile which we automatically append/merge to the contract's `Cargo.toml`. See https://github.com/paritytech/cargo-contract/blob/a8c8589798c05468b41e8125b6c5c3c1194178b6/src/workspace/profile.rs#L22.

I think we might still need to pass `Clinker-plugin-lto` directly as an argument, but AFAIU it seems dependent upon `-Clto` which is configurable from the `[profile.release]`

The philosophy we have followed so far is that we should provide sensible defaults for these values, but also allow the user to override them, because the same hardcoded flags might now always produce the optimal contract size for all contracts.

ascjones (commented):
See comment

src/cmd/build.rs
xgreenx: What do you think if we will add it automatically [here](https://github.com/paritytech/cargo-contract/blob/a8c8589798c05468b41e8125b6c5c3c1194178b6/src/workspace/profile.rs#L53) if lto is on?=)

ascjones: `Clinker-plugin-lto` is not a profile setting https://doc.rust-lang.org/cargo/reference/profiles.html.

What effect does just having `Clto` on its own have on the size?

Anyway if this flag is providing such big gains, and doesn't harm any of the other `Lto` options then we can possibly hardcode it.

xgreenx: For example [that contract](https://github.com/Supercolony-net/openbrush-contracts/blob/main/examples/psp22/lib.rs) takes space:
Without `-Clto` and `-Clinker-plugin-lto`: Original wasm size: 100.5K, Optimized: 47.9K
With `-Clto`: Original wasm size: 68.6K, Optimized: 31.9K
With `-Clto` and `-Clinker-plugin-lto`: Original wasm size: 67.2K, Optimized: 31.0K

I tried it on contracts from [that folder](https://github.com/Supercolony-net/openbrush-contracts/tree/main/examples) and it causes a positive effect for all of them(but all of them has a heavy logic, except `reentrancy`. For `reentrancy ` it causes a low impact there).

src/cmd/build.rs
```diff
@@ -249,7 +249,7 @@ fn exec_cargo_for_wasm_target(
     // Currently will override user defined RUSTFLAGS from .cargo/config. See https://github.com/paritytech/cargo-contract/issues/98.
     std::env::set_var(
         "RUSTFLAGS",
-        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory",
+        "-C link-arg=-zstack-size=65536 -C link-arg=--import-memory -Clinker-plugin-lto",
```
ascjones, line 252: If this just does nothing with other `lto` settings then should be fine.

ascjones (commented):
Would be good to give the user control over these flags too see: https://github.com/paritytech/cargo-contract/issues/98

src/cmd/build.rs
xgreenx: I did measurements for `reentrancy`. Also, I removed explicit `-Clto` because it causes a compilation error if it is disabled or [not aplieable](https://github.com/rust-lang/cargo/issues/2301). I tested compiler will use `-Clinker-plugin-lto` if he can.

flipper:
Original wasm size: 38.9K, Optimized: 12.5K
Original wasm size: 24.4K, Optimized: 4.5K
Original wasm size: 24.1K, Optimized: 4.2K

flip_on_me:
Original wasm size: 33.8K, Optimized: 10.2K
Original wasm size: 21.5K, Optimized: 2.6K
Original wasm size: 21.2K, Optimized: 2.3K

xgreenx:
> Would be good to give the user control over these flags too see: #98

Yea, I saw that issue(I tried to increase the stack size in one moment xDD). I think it is not part of this PR=)

xgreenx:
Do I need to do something to get this PR approved? Or it is okey?=)

ascjones:
Can you confirm that with different LTO settings (via `[profile.release]`) that it will not break?

xgreenx:
> Can you confirm that with different LTO settings (via `[profile.release]`) that it will not break?

Yes, the compiler ignores that flag if the `lto` is disabled=)
image
image

Without the `lto` the size of `Erc20` is higher=)

ascjones:
> Without the lto the size of Erc20 is higher=)

Of course, hence the default. But we still want to provide the flexibility where possible.

ascjones (approved):

ascjones:
Have you removed the events to get `9.1K`?

xgreenx:
> Have you removed the events to get `9.1K`?

Nope, it is only with https://github.com/paritytech/ink/pull/1017 =)

xgreenx:
https://github.com/paritytech/cargo-contract/pull/364 done=)

xgreenx:
> Have you removed the events to get `9.1K`?

With https://github.com/paritytech/ink/pull/1016 and https://github.com/paritytech/ink/pull/1017 it takes Original wasm size: 31.1K, Optimized: 8.8K(8828)

PR #359, opened 2021-11-12
Enable building contract with metadata from another contract

cmichi:
Needed for https://github.com/paritytech/ink/issues/739.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+-  Added possibility to build a contract with the metadata of another contract - [#359](https://github.com/paritytech/cargo-contract/pull/359)
+
 ## [0.15.0] - 2021-10-18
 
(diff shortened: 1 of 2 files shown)


xgreenx:
I only want to clarify, that this feature is not needed for upgradability=)
Because of the idea of an upgradable contract, that you can change the code at any time, so old ABI can be incorrect, and better to not apply ABI at all for contracts like `Proxy`. 

I think it is better to have some util or command(in `cargo-contract`) that allows merging several contracts into one ABI file. After, the user can use this ABI file to do a call to the `Proxy` contract(so he should select the address of the proxy contract and then apply his ABI in UI).

xgreenx:
Processing metadata from child crates is important in terms of https://github.com/paritytech/ink/issues/809.
So I did an investigation and found the way, how you can collect metadata from crates.
https://github.com/xgreenx/ink-metadata

I created an example of how it can be implemented=) I think it should resolve our problem to generate metadata without a problem and correct=)

cmichi:
I'll close this issue as I used the flag mostly for testing and no longer think it makes sense in `master`.

PR #360, opened 2021-11-14
move to rust 2021

NukeManDan:


ascjones (commented):
This is not the only `Cargo.toml` file

NukeManDan:
`rust-version = "1.56.1"` introduced

For reference, I tried to get this to `cargo build` with 1.56.0 and it failed with
```
error: package `...` cannot be built because it requires rustc 1.56.1 or newer, while the currently active rustc version is 1.56.0
```
With `--ignore-rust-version` it _does_ build with 1.56.0, as expected.

ascjones (approved):

PR #368, opened 2021-11-22
Deserialize metadata

ascjones:
Extracted from #79. Would be required for any consumer to deserialize contract metadata into Rust types.

- Implement deserializing `cargo-contract` metadata
- Expose all fields as `pub` for decoded metadata.

### DIFF
--- a/metadata/lib.rs
+++ b/metadata/lib.rs
@@ -793,9 +793,9 @@ mod tests {
                 "types": []
             }
         }
-            .as_object()
-            .unwrap()
-            .clone();
+        .as_object()
+        .unwrap()
+        .clone();
 
         let metadata = ContractMetadata::new(source, contract, Some(user), abi_json);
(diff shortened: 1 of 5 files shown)


metadata/lib.rs
```diff
@@ -52,22 +52,27 @@
...
+};
 use url::Url;
 
 /// Smart contract metadata.
-#[derive(Clone, Debug, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct ContractMetadata {
-    source: Source,
-    contract: Contract,
+    pub source: Source,
```
cmichi, line 69: Could you add comments to the fields which are now `pub`?

cmichi:
@ascjones Forgot about this earlier, but could you update the changelog as well please? Then we can merge now-ish and include it in the release today.

cmichi (approved):

PR #376, opened 2021-11-26
Revert disabling overflow checks

cmichi:
Follow-up to https://github.com/paritytech/ink/pull/1049.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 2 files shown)


athei (approved):
I guess CI fails because it is against an old version of ink?

ascjones:
> I guess CI fails because it is against an old version of ink?

Yes, and because of this we might want to wait for `rc8` before merging this. Because as it is the template have `rc7` which still requires the overflow checks.

athei:
Ahh sure the template is against a release and not master. That makes sense. But I thought the next version isn't no RC but 3.0 proper? We could always make a quick rc right now? But I think it is still some manual labor to do so.

Issue #377, opened 2021-11-29
Remove `workspace` when building a contract


cmichi:
In https://github.com/paritytech/ink/pull/1054 @xgreenx found that if the `[workspace]` is removed from a contract's `Cargo.toml` the size decreases potentially significantly.

We should have `cargo-contract` remove the `[workspace]` section before building the contract.

xgreenx:
Adding of empty `[workspace]` ignores `[workspace]` from a parent. So it can be a solution=)
```rust
[workspace]
members = []
```

So I can handle that=)

PR #378, opened 2021-11-29
Added `workspace` section to override parent `workspace`

xgreenx:
It should close https://github.com/paritytech/cargo-contract/issues/377.
I tested it with sub contracts from the `delegator` example in ink!.

### DIFF
--- a/src/workspace/manifest.rs
+++ b/src/workspace/manifest.rs
@@ -283,6 +283,17 @@ impl Manifest {
         Ok(self)
     }
 
+    /// Set empty `[workspace]` section if it is not exist
+    ///
+    /// It allows ignore `workspace` from parent `Cargo.toml`.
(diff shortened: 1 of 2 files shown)


src/workspace/manifest.rs
```diff
@@ -283,6 +283,17 @@ impl Manifest {
         Ok(self)
     }
 
+    /// Set empty `[workspace]` section if it is not exist
+    ///
+    /// It allows ignore `workspace` from parent `Cargo.toml`.
+    /// It can reduce the size of the contract in some cases.
```
ascjones, line 289:
```suggestion
    /// Ignores the `workspace` from the parent `Cargo.toml`.
    /// This can reduce the size of the contract in some cases.
```

ascjones (approved):
LGTM.

PR #425, opened 2022-02-09
Update README.md: toolchain update tip

h4x3rotab:


README.md
```diff
@@ -51,6 +51,8 @@ More relevant links:
 
 * Step 3: `cargo install --force cargo-contract`.
 
+You can always update the toolchain to the latest version by running the Step 3.
```
HCastano, line 54:
```suggestion
You can always update the `cargo-contract` binary to the latest version by running the Step 3.
```

Small change in wording, but wouldn't want people to think that `binaryen` (for example) would also get updated as part of this

HCastano (approved):
Thanks!

PR #431, opened 2022-02-15
Add ink! linting MVP stage

cmichi:
This is an MVP, there are a couple shortcomings in there for which I want to create follow-ups:

* As it is now we unfortunately have to require people to install `dylint` manually. We should look into making it a proper dependency instead. However, this was not straight forward.
* We should render and publish a proper website displaying info about the lint (the one which right now is just in the doc comment). So similar to https://rust-lang.github.io/rust-clippy/master/.
* We should recognize nested calls to `ink_lang::codegen::initialize_contract`. So if it's not called directly in the constructor, but instead in a sub-function.


xgreenx:
> We should recognize if the call to ink_lang::codegen::initialize_contract is nested within the constructor. So if it's not called directly in the constructor, but instead in a sub-function.

Maybe better to require the user to always implement `SpreadAllocate` and we will use `ink_lang::codegen::initialize_contract` in the codegen?

If the contract implements `SpreadAllocate` we can create an instance of the contract. After we can pass that instance into the constructor and the user will do all the necessary stuff.

[Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=52cb2e5f10e7005478cd7a8a36ebca80) of the idea in the playground. 

```rust
impl Contract {
    #[ink(constrcutor)]
    fn new(mut self/* we will require `self` here*/, total_supply: u32) -> Self {
        self.total_supply = total_supply;
        self
    }
}
```

cmichi:
@xgreenx Thanks for your comment! So the thing is that we don't want to change the ink! constructor function signature, we would really like to stay as close to Rust syntax as possible.

Putting the `initialize_contract` call into the codegen would result in a changed function signature though. This has implications for e.g. the off-chain testing, so there some special API be needed so that it's still possible to e.g. write Rust unit tests for the contract.

Our thinking right now is that we will likely find a better way going forward and by requiring the `initalize_contract` in the meantime those contracts would be forward compatible. If we change the function signature now though, then those contracts can't be easily used with newer versions of ink! if we change the function signature back.

xgreenx:
Thanks for the response. I see you are right. I forgot about off-chain unit tests that the problem still exists for them.

> Our thinking right now is that we will likely find a better way going forward and by requiring the initalize_contract in the meantime those contracts would be forward compatible. If we change the function signature now though, then those contracts can't be easily used with newer versions of ink! if we change the function signature back.

Yesterday I tried to implement your idea regarding the new structure of the contract. [Here](https://github.com/paritytech/ink/issues/1134#issuecomment-1042351934) is results. Maybe you can check it in your free time=)

.github/workflows/macos.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     steps:
 
       - name: Checkout sources & submodules
-        uses: actions/checkout@master
+        uses: actions/checkout@v2.4.0
```
sergejparity, line 29:
```suggestion
        uses: actions/checkout@v2
```
It is recommended to use this tag as it following updates to the latest minor version

.github/workflows/windows.yml
```diff
@@ -32,27 +32,40 @@ jobs:
           pathInArchive: "binaryen-/bin/wasm-opt.exe"
 
       - name: Checkout sources & submodules
-        uses: actions/checkout@master
+        uses: actions/checkout@v2.4.0
```
sergejparity, line 35:
```suggestion
        uses: actions/checkout@v2
```
It is recommended to use this tag as it following updates to the latest minor version

.github/workflows/macos.yml
```diff
@@ -36,20 +36,33 @@ jobs:
 
       - name: Install toolchain
         id: toolchain
-        uses: actions-rs/toolchain@master
+        uses: actions-rs/toolchain@v1.0.7
```
sergejparity, line 39:
```suggestion
        uses: actions-rs/toolchain@v1
```
It is recommended to use this tag as it following updates to the latest minor version

.github/workflows/windows.yml
```diff
@@ -32,27 +32,40 @@ jobs:
...
         with:
           fetch-depth: 1
           submodules: recursive
 
       - name: Install toolchain
         id: toolchain
-        uses: actions-rs/toolchain@master
+        uses: actions-rs/toolchain@v1.0.7
```
sergejparity, line 42:
```suggestion
        uses: actions-rs/toolchain@v1
```
It is recommended to use this tag as it following updates to the latest minor version

sergejparity (approved):

build.rs
```diff
@@ -42,36 +42,147 @@ fn main() {
...
+    );
+    cmd.current_dir(ink_dylint_driver_dir.clone());
+
+    println!("Invoking cargo: {:?}", cmd);
+
+    let child = cmd
+        // capture the stdout to return from this function as bytes
+        .stdout(std::process::Stdio::piped())
+        .spawn()
+        .unwrap_or_else(|_| panic!("Error executing `{:?}`", cmd));
```
ascjones, line 149: Return `Err` instead of panicking here?

build.rs
```diff
@@ -42,36 +42,147 @@ fn main() {
...
+        });
+        std::process::exit(bool_to_exit_code(zipped_template));
+    }
+
+    #[cfg(not(feature = "cargo-clippy"))]
+    {
+        let zipped_lints = build_dylint_driver(manifest_dir, out_dir, dylint_driver_dst_file)
+            .map(|_| true)
+            .unwrap_or_else(|err| {
+                panic!("Failed building dylint driver: {:?}", err);
```
ascjones, line 97: If we are going to exit cleanly with an exit code below, then to be consistent it should either handle results or do `eprintln!` + `exit(1)`.

Or instead we can just `panic!` everywhere which might have the same result. Either way it should be consistent.

build.rs
```diff
@@ -42,36 +42,147 @@ fn main() {
...
+            });
+        std::process::exit(bool_to_exit_code(zipped_template && zipped_lints));
```
ascjones, line 99: I would just replace this whole block with a call to `build_dylint_driver` and then add an alternative with `#[cfg(feature = "cargo-clippy")]` since you already have that method decorated. It will make the control flow easier to follow

.gitlab-ci.yml
```diff
@@ -81,6 +87,26 @@ clippy:
...
+    # that would then make it impossible to upgrade the
+    # `ink_linting/rust-toolchain` file while still having
+    # this CI job succeed.
+    - rustup component add rustfmt clippy rust-src
+
+    - cargo check --verbose
+    - cargo fmt --verbose --all -- --check
+    - cargo clippy --verbose -- -D warnings;
+
+    - cargo test --verbose --workspace --all-features
```
ascjones, line 108: Is the `--workspace` necessary? Seems to run without it. Harmless anyway

ink_linting/ui/pass/mapping-two-constructors.rs
```diff
@@ -0,0 +1,60 @@
...
+    use ink_storage::{traits::SpreadAllocate, Mapping};
+
+    #[ink(storage)]
+    #[derive(SpreadAllocate)]
+    pub struct MyContract {
+        balances: Mapping<AccountId, Balance>,
+    }
+
+    impl MyContract {
+        /// The `initialize_contract` must be found, even if there is logic before it.
```
ascjones, line 32: Is this a copy of the "additional logic" UI test? There is only a single constructor here where there should be two according to the test name.

.gitlab-ci.yml
athei: Yes. Because we are not using a virtual manifest. All commands only run on the main package in thise case without `--workspace` or a `default-member` key in the manifest.

build.rs
```diff
@@ -42,36 +42,151 @@ fn main() {
...
+        .map(|_| true)
+        .unwrap_or_else(|err| {
+            eprintln!("Failed building dylint driver: {:?}", err);
+            std::process::exit(1);
+        });
+    std::process::exit(bool_to_exit_code(zipped_template && zipped_lints));
+}
+
+/// Returns the process exit code which corresponds to a boolean.
+fn bool_to_exit_code(val: bool) -> i32 {
```
ascjones, line 86: This function only called from the one place now so could be moved inline

build.rs
```diff
@@ -42,36 +42,151 @@ fn main() {
...
+                ink_dylint_driver_dir.display(),
+                dylint_driver_dst_file.display()
+            );
+            Ok(())
+        }
+        Err(e) => {
+            eprintln!("Error: {:?}", e);
+            Err(e)
+        }
+    }
```
ascjones, line 186: To avoid destructuring the result here, could add another high level method below main which does roughly:

```
fn zip_template_and_build_dylint_driver() -> Result<()> {
  zip_template()?;
  build_dylint_driver()?;
  println!("Some success message which used to be in the Ok arm");
  Ok(())
}
```
Then at the top level in `main()` you can do the `eprintln!` which will handle all errors, and only need to destructure once. The `println!`s in the `Ok` arm can just move inline to the methods.

src/cmd/build.rs
```diff
@@ -298,8 +295,88 @@ fn exec_cargo_for_wasm_target(
...
+            .to_string()
+            .bright_yellow());
+    }
+
+    if !execute_cmd(Command::new("dylint-link").arg("--version")) {
+        anyhow::bail!("dylint-link was not found!\n\
+            Make sure it is installed and the binary is in your PATH environment.\n\n\
+            You can install it by executing `cargo install dylint-link`."
+            .to_string()
+            .bright_yellow());
```
athei, line 378: I think this is useless here? We don't need `dylint-link` when running `cargo-contract` but when building it. This should go into `./build.rs` in order to give a proper error when this isn't installed during build.

ink_linting/src/mapping_initialized.rs
```diff
@@ -0,0 +1,345 @@
...
+            let constructor_span = constructors_without_initialize.get(0)
+                .expect("at least one faulty constructor must exist");
+            let snippet = snippet_opt(cx, *constructor_span)
+                .expect("snippet must exist");
+            span_lint_and_then(
+                cx,
+                MAPPING_INITIALIZED,
+                item.span,
+                &format!(
+                    "you have declared an `#[ink(storage)]` on `{}` without initializing it in the contract constructor.",
```
ascjones, line 194: Needs to be more specific, and perhaps without naming and shaming the user. If you look at the Rust compiler errors they are normally in the passive voice https://doc.rust-lang.org/error-index.html.

Here's a few ideas (though try and come up with something more readable!)

"missing contract initialization in constructor for #[ink(storage)] struct with a Mapping field."
"Storage struct with a Mapping field must be initialized on construction"

ink_linting/src/mapping_initialized.rs
```diff
@@ -0,0 +1,345 @@
...
+/// Visitor to determine if a `fn` contains a call to `ink_lang::utils::initialize_contract(…)`.
+///
+/// # Known Limitation
+///
+/// This function currently only finds call to `initialize_contract` which happen
+/// on the first level of the `fn` ‒ no nested calls are found! So if you would
+/// call `initialize_contract` within a sub-function of the ink! constructor
+/// this is not recognized!
+///
+/// Also the function call must be explicitly called `ink_lang::utils::initialize_contract(…)`,
```
ascjones, line 306: Is this true though? I modified the `mapping_one_constructor` to `use ink_lang::utils::initialize_contract;` and it still works. Also if you look at the regex it uses the non reexported name to match on

ascjones (commented):
First pass complete

.gitlab-ci.yml
```diff
@@ -81,6 +87,26 @@ clippy:
 
 #### stage:                        test (all features)
 
+test-dylint:
+  stage:                           test
+  <<:                              *docker-env
+  script:
```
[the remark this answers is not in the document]
cmichi: Yes exactly, rustc starts complaining about `incompatible metadata version found`.

cmichi:
@ascjones @athei I've addressed your comments and updated the PR. The error which Alex encountered happened when no `--manifest-path` was supplied, I've fixed that as well.

src/cmd/build.rs
```diff
@@ -298,8 +295,108 @@ fn exec_cargo_for_wasm_target(
...
+
+    let template = include_bytes!(concat!(env!("OUT_DIR"), "/ink-dylint-driver.zip"));
+    crate::util::unzip(template, tmp_dir.path().to_path_buf(), None)?;
+
+    let manifest_path = crate_metadata.manifest_path.cargo_arg()?;
+    let args = vec!["ink_linting", &manifest_path];
+    let tmp_dir_path = tmp_dir.path().as_os_str().to_string_lossy();
+    let env = vec![
+        ("DYLINT_LIBRARY_PATH", Some(tmp_dir_path.as_ref())),
+        ("DYLINT_DRIVER_PATH", Some(tmp_dir_path.as_ref())),
```
athei, line 322: When I remove this line the driver is cached and doesn't need to be rebuild on every check. Makes sense because you are compiling it to a new temporary directory each time.

I don't know where it is put when I remove this line, though.

src/cmd/build.rs
```diff
@@ -298,8 +295,108 @@ fn exec_cargo_for_wasm_target(
...
+    let tmp_dir = tempfile::Builder::new()
+        .prefix("cargo-contract-dylint_")
+        .tempdir()?;
+    log::debug!("Using temp workspace at '{}'", tmp_dir.path().display());
+
+    let template = include_bytes!(concat!(env!("OUT_DIR"), "/ink-dylint-driver.zip"));
+    crate::util::unzip(template, tmp_dir.path().to_path_buf(), None)?;
+
+    let manifest_path = crate_metadata.manifest_path.cargo_arg()?;
+    let args = vec!["ink_linting", &manifest_path];
```
athei, line 318: Let's be explicit.

```suggestion
    let args = vec!["--lib", "ink_linting", &manifest_path];
```

src/cmd/build.rs
```diff
@@ -298,8 +295,108 @@ fn exec_cargo_for_wasm_target(
...
+/// `cargo dylint` with it.
+fn exec_cargo_dylint(crate_metadata: &CrateMetadata, verbosity: Verbosity) -> Result<()> {
+    check_dylint_requirements(crate_metadata.manifest_path.directory())?;
+
+    let tmp_dir = tempfile::Builder::new()
+        .prefix("cargo-contract-dylint_")
+        .tempdir()?;
+    log::debug!("Using temp workspace at '{}'", tmp_dir.path().display());
+
+    let template = include_bytes!(concat!(env!("OUT_DIR"), "/ink-dylint-driver.zip"));
```
athei, line 314: Copy paste variable name?

build.rs
```diff
@@ -35,43 +35,160 @@ use substrate_build_script_utils::rerun_if_git_head_changed;
...
+fn zip_template(manifest_dir: &Path, out_dir: &Path) -> Result<()> {
+    let template_dir = manifest_dir.join("templates").join("new");
+    let template_dst_file = out_dir.join("template.zip");
     println!(
         "Creating template zip: template_dir '{}', destination archive '{}'",
         template_dir.display(),
-        dst_file.display()
+        template_dst_file.display()
     );
+    zip_dir(&template_dir, &template_dst_file, CompressionMethod::Stored).map(|_| {
```
ascjones, line 85: Not a fan of `map`ping into an io side effect, but now I am just nitpicking

ascjones (approved):
LGTM

src/cmd/build.rs
athei: I mean replacing `let template` by `let driver` or something.

athei:
I am wondering: We are writing our custom link but don't even run the stock clippy lints. Can we do that at the same time as we are running dylint so we don't have multiple passes?

xgreenx:
Is that possible to use `CARGO_TARGET_DIR` with `dylint` to not build it each time?
Also maybe make `dylint` configurable? For example, I don't want to build it during integration tests=)

PR #462, opened 2022-03-14
Fix issues with publishing

cmichi:
Ran into those when trying to publish the latest release.


cmichi:
I've removed the CI stage again, as the `../../../` relative path doesn't play well with our heavily modified `CARGO_…` env variables for the CI setup.

An alternative to the fixed `../../../` would be to require an env variable `INK_DYLINY_PATH`.

ascjones (approved):

PR #463, opened 2022-03-15
Include `ink_linting` folder in releases + Test publish/install in CI

cmichi:
I'm super unhappy about the `_Cargo.toml` solution, but can't think of anything better right now.

Note that we also use this hack for the crates in `templates/`.


.gitlab-ci.yml
```diff
@@ -131,6 +132,29 @@ test-new-project-template:
...
+# is more elaborate now and as part of it the `ink_linting` crate is built locally.
+# We introduced this CI job to make sure that publishing to a registry and installing
+# from it will work correctly.
+test-registry-publish-install:
+  stage:                           test
+  <<:                              *docker-env
+  before_script:
+    # Set up a local registry.
+    - mkdir -p ./estuary/crates/ ./estuary/indices/
+    - estuary --base-url=http://0.0.0.0:7878 --crate-dir ./estuary/crates/ --index-dir ./estuary/indices &
```
ascjones, line 145: So is `estuary` already part of docker image or installed somewhere else?

ascjones (approved):

PR #465, opened 2022-03-16
[DO NOT MERGE] gitspiegel trial

joao-paulo-parity:
Sorry for the unintended pings, I just want to see if gitspiegel is working properly now since @TriplEight reported the problem


joao-paulo-parity:
it is: https://gitlab.parity.io/parity/mirrors/cargo-contract/-/tree/465

Issue #467, opened 2022-03-18
Install `cargo-contract` failed


AsceticBear:

cmichi:
Oha, thank you for creating this issue!

I can't reproduce it locally and our CI also runs fine, I've created a branch with more detailed error reporting. Could you please try this and post the output?

```
cargo install --git https://github.com/paritytech/cargo-contract.git --branch cmichi-add-build-script-output --force --locked
```

AsceticBear:
This command works! Do you know the root cause here?

PR #469, opened 2022-03-18
Improve error reporting during installation + Release `1.0.1`

cmichi:
Follow-up to https://github.com/paritytech/cargo-contract/issues/467, where we don't have an idea what causes it, but the next time someone runs into it we would have better error reporting.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -96,10 +96,8 @@ fn zip_template_and_build_dylint_driver(manifest_dir: PathBuf, out_dir: PathBuf)
     let original_name = ink_dylint_driver_dir.join("_Cargo.toml");
     if !original_name.exists() {
         anyhow::bail!(
-            "'{:?}' does not exist, does the folder '{:?}'? {:?}",
+            "'{:?}' does not exist",
             original_name,
-            ink_dylint_driver_dir,
-            ink_dylint_driver_dir.exists()
         );
     }
 
--- a/build.rs
+++ b/build.rs
@@ -70,6 +70,14 @@ fn zip_template_and_build_dylint_driver(manifest_dir: PathBuf, out_dir: PathBuf)
     let dylint_driver_dst_file = out_dir.join("ink-dylint-driver.zip");
 
     let ink_dylint_driver_dir = manifest_dir.join("ink_linting");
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -85,24 +85,42 @@ fn zip_template_and_build_dylint_driver(manifest_dir: PathBuf, out_dir: PathBuf)
...
-        ink_dylint_driver_dir.join("Cargo.toml"),
-    )?;
-
-    let res = build_and_zip_dylint_driver(
-        ink_dylint_driver_dir.clone(),
-        out_dir,
-        dylint_driver_dst_file,
-    );
+    let original_name = ink_dylint_driver_dir.join("_Cargo.toml");
+    let original_name = original_name.canonicalize().map_err(|err| {
```
cmichi, line 89: I've also added path-canonicalization here now.

ascjones: Why not just canonicalize `ink_dylint_driver_dir` in the first place, then both `original_name` and `tmp_name` will be canonical.

build.rs
```diff
@@ -85,24 +93,39 @@ fn zip_template_and_build_dylint_driver(manifest_dir: PathBuf, out_dir: PathBuf)
...
-        out_dir,
-        dylint_driver_dst_file,
-    );
+    let original_name = ink_dylint_driver_dir.join("_Cargo.toml");
+    if !original_name.exists() {
+        anyhow::bail!(
+            "'{:?}' does not exist, does the folder '{:?}'? {:?}",
+            original_name,
+            ink_dylint_driver_dir,
+            ink_dylint_driver_dir.exists()
```
ascjones, line 102: I guess this will always be true because `canonicalize` would fail otherwise.

ascjones (approved):

build.rs
cmichi: https://doc.rust-lang.org/std/fs/fn.canonicalize.html#errors
>This function will return an error in the following situations, but is not limited to just these cases:
> * path does not exist.
> * A non-final component in path is not a directory.

So it could also be that the path exists, but something else doesn't work out.

ascjones: Right so `ink_dylint_driver_dir.exists()` will always be true if we get this far?

cmichi: Ahh right, I just now checked the context, thought you annotated another line.

PR #481, opened 2022-03-29
Add `decode` command for event, message and constructor data decoding

agryaznov:
As a response to user feature request (e.g. like in [this StackExchange question](https://substrate.stackexchange.com/questions/587/how-to-decode-smart-contract-call-and-event-data)):  
1. New command added which allows to decode *data* from contract events and messages
2. Corresponding *encoded* values printing added to debug log (where there was lack of such)


src/cmd/extrinsics/decode.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright 2018-2020 Parity Technologies (UK) Ltd.
```
ascjones, line 1: We might need to move things around a bit, since `decode` is not an extrinsic so does not belong in the `extrinsics` folder. However since it does require the `transcode` module, perhaps we could lift this command and that module up to the root.

src/cmd/decode.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, clap::ArgEnum)]
+enum DataType {
+    Event,
+    Message,
+}
+
+impl DecodeCommand {
+    pub fn run(&self) -> Result<()> {
```
ascjones, line 43: I'd like to see a basic test for decoding a message and an event. to prove that it works. As with your recently reported issues it is easy to introduce regressions.There are a few options here:

1. A full end to end test like https://github.com/paritytech/cargo-contract/blob/82157f46f01197f662715a3f0a81dd9d749686d3/src/cmd/build.rs#L1445 which invokes the process.
2. Invoking `run` directly with some setup https://github.com/paritytech/cargo-contract/blob/82157f46f01197f662715a3f0a81dd9d749686d3/src/cmd/build.rs#L1445
3. Factor out a function and pass in the contract metadata.

src/cmd/extrinsics/transcode/mod.rs
```diff
@@ -203,6 +202,37 @@ impl<'a> ContractMessageTranscoder<'a> {
         Ok(Value::Map(map))
     }
 
+    pub fn decode_contract_message(&self, data: &mut &[u8]) -> Result<Value> {
+        let mut msg_selector = [0u8; 4];
+        data.read(&mut msg_selector)?;
+        let msg_spec = self
+            .metadata
+            .spec()
+            .messages()
```
ascjones, line 211: Come to think of it, we should probably support constructor decoding too.

src/cmd/decode.rs
```diff
@@ -0,0 +1,163 @@
...
+                .arg("nightly")
+                .assert()
+                .success();
+
+            log::info!("Building contract in {}", project_dir.to_string_lossy());
+            cargo_contract(project_dir).arg("build").assert().success();
+
+            let msg_data: &str = "babebabe01";
+            let msg_decoded: &str =
+                r#"Ok(Map(Map { ident: Some("switch"), map: {String("value"): Bool(true)} }))"#;
```
ascjones, line 131: Ok so your test has highlighted an issue here :tada:  We should not be printing the debug `{:?}` output but rather the `Display` output

src/cmd/decode.rs
```diff
@@ -0,0 +1,163 @@
...
+        let decoded_data = match self.r#type {
+            DataType::Event => {
+                transcoder.decode_contract_event(&mut &decode_hex(&self.data).context(ERR_MSG)?[..])
+            }
+            DataType::Message => transcoder
+                .decode_contract_message(&mut &decode_hex(&self.data).context(ERR_MSG)?[..]),
+        };
+
+        println!(
+            "{:>width$} {:?}",
```
ascjones, line 57:
```suggestion
            "{:>width$} {}",
```

ascjones: As above.

src/cmd/decode.rs
```diff
@@ -0,0 +1,163 @@
...
+
+            log::info!("Building contract in {}", project_dir.to_string_lossy());
+            cargo_contract(project_dir).arg("build").assert().success();
+
+            let msg_data: &str = "babebabe01";
+            let msg_decoded: &str =
+                r#"Ok(Map(Map { ident: Some("switch"), map: {String("value"): Bool(true)} }))"#;
+
+            // then
+            // message data is being decoded properly
```
ascjones, line 134: Can we also add one for a constructor?

ascjones (commented):
Looks good, just a couple of nitpicks, and also we need to add `constructor` decoding

src/cmd/decode.rs
```diff
@@ -0,0 +1,163 @@
...
+        let contract = r#"
+	#![cfg_attr(not(feature = "std"), no_std)]
+
+	use ink_lang as ink;
+
+	#[ink::contract]
+	pub mod switcher {
+	    #[ink(event)]
+	    pub struct Switched {
+		new_value: bool,
```
ascjones, line 91: Indentation

src/cmd/decode.rs
```diff
@@ -0,0 +1,163 @@
...
+	    }
+
+	    #[ink(storage)]
+	    pub struct Switcher {
+		value: bool,
```
ascjones, line 96: Indentation

src/cmd/decode.rs
```diff
@@ -0,0 +1,163 @@
...
+	    impl Switcher {
+		#[ink(constructor)]
+		pub fn new(init_value: bool) -> Self {
+		    Self { value: init_value }
+		}
+
+		#[ink(message, selector = 0xBABEBABE)]
+		pub fn switch(&mut self, value: bool) {
+		    self.value = value;
+		}
```
ascjones, line 108: Indentation

src/cmd/decode.rs
agryaznov: Looks like the `.rustfmt.toml` needs to be fixed along with indentation in all project sources. I'll create a separate PR for this.

ascjones (approved):
LGTM

src/cmd/decode.rs
```diff
@@ -0,0 +1,179 @@
...
+mod tests {
+    use crate::util::tests::with_new_contract_project;
+    use std::path::Path;
+
+    /// Create a `cargo contract` command
+    fn cargo_contract(path: &Path) -> assert_cmd::Command {
+        let mut cmd = assert_cmd::Command::cargo_bin(env!("CARGO_PKG_NAME")).unwrap();
+        cmd.current_dir(path).arg("contract");
+        cmd
+    }
```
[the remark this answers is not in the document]
ascjones: Seems that `assert_cmd` [assumes](https://docs.rs/assert_cmd/latest/assert_cmd/cargo/index.html#limitations) it is an integration test (therefore the binary would already have been built). 

I've moved this test to integration test to see if it works now.

Issue #432, opened 2022-02-16
Extrinsics follow ups


ascjones:
Follow ups to #79. In no particular order, should be promoted to their own issues.

- Add support for remove_code extrinsic.
- Gas estimation: auto dry run first to use estimated gas as a value for the gas limit when running actual call.
- Add flag `--wait-for-finalization`. so that success/events are not displayed until the block is finalized
- Denominated units for Balances e.g. `100CAN`, `1MCAN`, instead of e.g. `10000000000000000000000000000`
- Signing: safer alternatives for value bearing chains than providing secret key on the command line. Possibly allow using encrypted file with private key, or integrating with external signers somehow.
- Dynamic `DispatchError::Module` errors, so that error details are displayed correctly across different chains https://github.com/paritytech/subxt/issues/443
- Bitvec encoding/decoding
- Improve error message when failing to encode contract message arg https://github.com/paritytech/cargo-contract/pull/79/commits/e40dc8051b37074afad69f18ccdc35e1de408dd8
-  Set up CI with `substrate-contracts-node` and unignore integration test: https://github.com/paritytech/cargo-contract/blob/60f65a2863be1266882bdd81861d836baeb06724/src/cmd/extrinsics/integration_tests.rs#L124

ascjones:
Extracted all to separate issues, or PRs already in progress

PR #484, opened 2022-03-31
Dry run gas limit estimation

ascjones:
To determine the gas limit to supply for instantiate and call commands, first invoke via rpc to determine `gas_required` for estimation.

By default `instantiate` and `call` commands will do a dry run first, and if `--gas` is not specified will use the `gas_required` from the RPC result as the `gas_limit` for the subsequent transaction. The user can skip the dry run with `--skip-dry-run`, in which case the `--gas` parameter is required.

The user will be prompted to confirm transaction submission with the suggested gas limit. This prompt can be skipped with the `--skip-confirm` flag, and the transaction will be submitted without user input.

image

From https://github.com/paritytech/cargo-contract/issues/432.

/cc @athei


athei:
I tested this. Works fine so far. But there is one regression: It does no longer print the debug buffer. Also, I think the formatting of the error is harder to read now. I think the output for the automatic dry run should be the same as for the manual one before.

Before:
```
Result ModuleError: Contracts::ContractTrapped: ["Contract trapped during execution."]
    Gas Consumed 185457130
    Gas Required 185457130
 Storage Deposit StorageDeposit::Charge(541630000000)
   Debug Message panicked at 'lol', /Users/alexander/Downloads/flipper/lib.rs:29:13
```

Now:
```
ERROR: Pre-submission dry-run failed: 'ModuleError: Contracts::ContractTrapped: ["Contract trapped during execution."]'. Use --skip-dry-run to skip this step.
```

ascjones:
image

athei:
Nice this issue is fixed. More:
* No automatic dry run is performed for `upload_code` (needed to inform user about `storage_deposit_limit` and subsequently set it for the extrinsic)
* Automatic dry-run for `instantiate` or `call` doesn't show storage deposit on success. I assume it doesn't set it for the extrinsic either. On error or for the manual dry-run the full information is shown.
* I can't confirm by just pressing enter. I need to type out `Y` which seems to be the default choice. It is also case sensitive.

ascjones:
>     * No automatic dry run is performed for `upload_code` (needed to inform user about `storage_deposit_limit` and subsequently set it for the extrinsic)
> 
>     * Automatic dry-run for `instantiate` or `call` doesn't show storage deposit on success. I assume it doesn't set it for the extrinsic either. On error or for the manual dry-run the full information is shown.

My intention was to add storage deposit as a nice and easy follow up, since it wasn't part of the (original) scope.

>     * I can't confirm by just pressing enter. I need to type out `Y` which seems to be the default choice. It is also case sensitive.

This was by design, I have seen it in other CLIs where confirming requires pressing two keys to make *Y* to avoid accidental presses, as in "are you really sure?" confirmation. However happy to make it less strict if you think that is too much

athei:
Ok then let's keep the deposit as follow up.

Regarding the confirmation: The only confusing thing is that `Y/n` suggests that the capital option is the default. I suggest making them both lowercase or both uppercase.

ascjones:
> Regarding the confirmation: The only confusing thing is that `Y/n` suggests that the capital option is the default. I suggest making them both lowercase or both uppercase.

Ok I have reverted to your previous suggestion, make it case insensitive and have **Y** as the default option. It seems I was wrong and the convention is as you suggest: https://askubuntu.com/questions/322600/do-you-want-to-continuey-n-why-the-upper-case.

image

athei (approved):

PR #519, opened 2022-04-22
extrinsics: fix `Vec<AccountId>` args

ascjones:
Where a message or a constructor accepts a `Vec<AccountId>` as an argument e.g [`multisig`](https://github.com/paritytech/ink/blob/0897dea3a8995207fd3ce7ac623a36f0f3eb0672/examples/multisig/lib.rs#L283), the custom decoder should be used so that SS58 format. e.g.

```bash
 contract instantiate \
      --args 2 [5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY,5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty] \
      --suri //Alice \
      --dry-run
```

### DIFF
--- a/src/cmd/extrinsics/instantiate.rs
+++ b/src/cmd/extrinsics/instantiate.rs
@@ -336,9 +336,10 @@ impl<'a> Exec<'a> {
             salt: self.args.salt.clone(),
         };
         let params = rpc_params![call_request];
-        let result: ContractInstantiateResult =
-            cli.request("contracts_instantiate", params).await
-                .context("contracts_instantiate RPC error")?;
+        let result: ContractInstantiateResult = cli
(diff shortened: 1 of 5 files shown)


cmichi:
What about `(AccountId, …)`?

ascjones:
> @ascjones What about `(AccountId, …)`?

You mean composite types containing an `AccountId`? Already handled: https://github.com/paritytech/cargo-contract/blob/cfd55bd66811f959e465bd7f1b6d5d60e2240fbc/src/cmd/extrinsics/transcode/transcoder.rs#L583

cmichi (approved):

PR #534, opened 2022-04-27
Fix `--dry-run` error deserialization and report error details

### LINKED ISSUE #533: `ModuleError` out of sync, fails to deserialize `--dry-run` error
Since https://github.com/paritytech/substrate/pull/10242, the [`error` field on `ModuleError`](https://github.com/paritytech/substrate/blob/01a3ad65d933cc58e79c04d114edb4d6c69b470a/primitives/runtime/src/lib.rs#L476) is now a `[u8; 4]`, and in the released version of `sp_runtime` we depend upon it is still a `u8`. 

So when we call `instantiate/call` with `--dry-run` and it triggers a `ContractTrapped` or any other internal `pallet_contracts` error, the returned error fails to deserialize because of the incompatible `ModuleError` types: it expects a `u8` but receives a `[u8; 4]`.

One solution would be to do a new `sp_runtime` release in substrate with the new struct, though that has the downside of not being backwards compatible. But we may not even want to be backwards compatible until contracts are actually on a production chain somewhere.

ascjones:
Closes #533.

This now parameterizes the RPC results with the generated `DispatchError` type from `subxt` so the types line up with the latest contracts node for deserializataion.

This also allows us to retrieve the `DispatchError` details from the metadata so we can map to a user friendly error:

image


src/cmd/extrinsics/call.rs
```diff
@@ -34,17 +36,25 @@ use jsonrpsee::{
...
 use subxt::{
     rpc::NumberOrHex,
     ClientBuilder,
     Config,
     DefaultConfig,
 };
 
-type ContractExecResult = pallet_contracts_primitives::ContractExecResult<Balance>;
+type ContractExecResult =
+    ContractResult<result::Result<ExecReturnValue, RuntimeDispatchError>, Balance>;
```
ascjones, line 57: This is the crux of the change, it is using the `subxt` generated `RuntimeDispatchError` instead of the default `sp_runtime::DispatchError`. This means it is using the same `DispatchError` type as would be returned in case of an error on calling the extrinsic.

src/cmd/extrinsics/runtime_api/mod.rs
```diff
@@ -15,6 +15,17 @@
 // along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
 
 #[subxt::subxt(
-    runtime_metadata_path = "src/cmd/extrinsics/runtime_api/contracts_runtime.scale"
+    runtime_metadata_path = "src/cmd/extrinsics/runtime_api/contracts_runtime.scale",
+    derive_for_type(type = "sp_runtime::DispatchError", derive = "::serde::Deserialize"),
```
ascjones, line 19: Adding these derives to these generated types is necessary so that they can be used to deserialize the RPC `--dry-run` error responses.

athei (approved):
LGTM

PR #560, opened 2022-05-17
Fix CI badge in README

rcny:
Just a small cosmetic change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 [![CI Status][a1]][a2]
 [![Latest Release][d1]][d2]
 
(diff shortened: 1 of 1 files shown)


cmichi:
@rcny I would like to rerun the failed CI job here: https://gitlab.parity.io/parity/mirrors/cargo-contract/-/jobs/1572737. But I can't see the typical "Retry" button. Could you check that I have sufficient permissions? Maybe some mishap of the migration to the new mirroring strategy?

cmichi (approved):

Issue #513, opened 2022-04-20
Interrupting installation leaves inconsistent state and if resumed fails


cmichi:
If `cargo install cargo-contract --force` is interrupted with Ctrl + C it leaves an inconsistent temp dir and the installation cannot be resumed due to `_Cargo.toml` no longer being available (because it's already named `Cargo.toml` at that point).

PR #571, opened 2022-05-19
Fix bug when `cargo-contract` installation had been interrupted

cmichi:
Fixes https://github.com/paritytech/cargo-contract/issues/513.

If the crate installation had been interrupted the file `_Cargo.toml` had already been renamed to `Cargo.toml`. If the installation was resumed the `_Cargo.toml` was no longer found. This resulted in the panic `Failed renaming _Cargo.toml to Cargo.toml: File not found.`.

I fixed it by copy (with overwrite) the file `_Cargo.toml` to `Cargo.toml` instead.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -111,9 +111,9 @@ fn zip_template_and_build_dylint_driver(
     }
 
     let tmp_name = ink_dylint_driver_dir.join("Cargo.toml");
-    std::fs::rename(&original_name, &tmp_name).map_err(|err| {
+    std::fs::copy(&original_name, &tmp_name).map_err(|err| {
         anyhow::anyhow!(
-            "Failed renaming '{:?}' to '{:?}': {:?}",
+            "Failed copying '{:?}' to '{:?}': {:?}",
             original_name,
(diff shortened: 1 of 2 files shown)


athei:
Without suggesting that you should do that here (but it wouldn't be too hard): A more elegant solution would use a guard type that does the cleanup on `Drop`. This will be safe against panics.

build.rs
```diff
@@ -64,6 +64,17 @@ fn main() {
     }
 }
 
+/// Holds the path to a file meant to be temporary.
+struct TmpFileGuard(PathBuf);
+
+impl Drop for TmpFileGuard {
+    // Once the struct instance is dropped we remove the file.
```
athei, line 71:
```suggestion
    /// Once the struct instance is dropped we remove the file.
```

athei (approved):
This is fine. But you could be a total champ and put the type into a private sub module and do the copy/rename in the constructor in order to have proper RAII. We don't need the copy and stick with rename in this case even.

PR #591, opened 2022-05-27
Manually convert Url::url to String in order to include the default port

### LINKED ISSUE #587: Unable to connect via the default port
`cargo-contract` uses `url` library to parse the command line argument to `url::Url` and then converts it back to `String` while passing to `subxt::ClientBuilder`. 

The whole `url` lib operates under the assumption that if the default port for a particular protocol is used, it is represented as `None` variant. For some reason it doesn't care to include that default port when converting `Url` to `String`:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=59e7111b1aa5c5542e1127c1f13b88ec 

So that whole `String` -> `Url` -> `String` conversion results in the port value being lost in case the default port was used. `subxt::ClientBuilder` rejects to use an address without a port.

h4nsu:
Fixes #587

### DIFF
--- a/src/cmd/extrinsics/mod.rs
+++ b/src/cmd/extrinsics/mod.rs
@@ -111,6 +111,18 @@ impl ExtrinsicOpts {
     pub fn verbosity(&self) -> Result<Verbosity> {
         TryFrom::try_from(&self.verbosity)
     }
+
+    /// Convert URL to String without omitting the default port
+    pub fn url_to_string(&self) -> String {
+        let mut res = self.url.to_string();
+        match (self.url.port(), self.url.port_or_known_default()) {
+            (None, Some(port)) => {
+                res.insert_str(res.len() - 1, &format!(":{}", port));
+                res
+            }
+            _ => res,
+        }
+    }
 }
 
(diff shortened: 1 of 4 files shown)


src/cmd/extrinsics/mod.rs
```diff
@@ -111,6 +111,18 @@ impl ExtrinsicOpts {
...
         TryFrom::try_from(&self.verbosity)
     }
+
+    /// Convert URL to String without omitting the default port
+    pub fn url_to_string(&self) -> String {
+        let mut res = self.url.to_string();
+        match (self.url.port(), self.url.port_or_known_default()) {
+            (None, Some(port)) => {
+                res.insert_str(res.len() - 1, &format!(":{}", port));
+                res
```
ascjones, line 121:
```suggestion
                format!("{res}:{port}"));
```

h4nsu: That doesn't work due to trailing `/`. Your solution produces `wss://your.address/:443` when what we need is `wss://your.address:443/`

ascjones (approved):
Thanks!

PR #594, opened 2022-05-30
Fix `instantiate_with_code` with already uploaded code

### LINKED ISSUE #589: ERROR: Failed to find CodeStored event
The error `Failed to find CodeStored event` is always emitted when a contract is instantiated using this command:

`cargo contract instantiate $WASM --args $CONTRACT_ARGS --constructor $CONSTRUCTOR --suri $SURI --value $ENDOWMENT --url '$ENDPOINT:$PORT'"`

The error is emitted from:

https://github.com/paritytech/cargo-contract/blob/2961428c231b8203935f9e26142b00015f5d19c2/src/cmd/extrinsics/instantiate.rs#L290-L295

It seems irrelevant whether the `CodeStored` event is emitted as the contract works ok.

ascjones:
When calling `instantiate_with_code` more than once (with different `salt` values), since the code has already been uploaded on the first call, subsequent calls will not store the code and not raise the `CodeStored` event. 

Fixes #589.

### DIFF
--- a/src/cmd/extrinsics/instantiate.rs
+++ b/src/cmd/extrinsics/instantiate.rs
@@ -251,7 +251,9 @@ impl<'a> Exec<'a> {
             Code::Upload(code) => {
                 let (code_hash, contract_account) =
                     self.instantiate_with_code(code).await?;
-                name_value_println!("Code hash", format!("{:?}", code_hash));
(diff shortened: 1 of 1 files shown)


HCastano (approved):

forgetso:
Thanks!

PR #597, opened 2022-06-03
Refactor out transcode as a separate library

fbielejec:
This PR refactors the bits responsible for encoding contract creator and message calls into a separate library.

It's intention is to facilitate building various tools for interacting with contracts.
One possible use case is creating automated end-to-end tests written in Rust, declarative DSL based tools for describing pipelines for deploying / updating non-trivial smart contract suites (think Truffle / Hardhat) etc.

For that reason the Licence is Apache (although I have cargo-culted the contract-metadata library in that respect, [GPL with ClasspathException](https://en.wikipedia.org/wiki/GPL_linking_exception#The_classpath_exception) would likely also be permissive enough - Although I'm Not A Lawyer).


fbielejec:
`test-registry-publish-install` check is failing but I do not have access to your cargo registry to fix that, so will need some help :open_hands:

transcode/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+itertools = "0.10.3"
+log = "0.4.17"
+nom = "7.1.1"
+nom-supreme = { version = "0.7.0", features = ["error"] }
+scale-info = { version = "2.1.1", default-features = false, features = ["derive"] }
+serde = { version = "1.0.137", default-features = false, features = ["derive"] }
+serde_json = "1.0.81"
+scale = { package = "parity-scale-codec", version = "3.0.0", features = ["derive"] }
+
+# dependencies for extrinsics (deploying and calling a contract)
```
ascjones, line 37: This comment seems redundant for this standalone crate

ascjones (approved):
LGTM!

transcode/Cargo.toml
fbielejec: removed

transcode/Cargo.toml
```diff
@@ -0,0 +1,51 @@
+[package]
+name = "contract-transcode"
+version = "0.6.0"
```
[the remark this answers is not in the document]
fbielejec: fixed - 0.1.0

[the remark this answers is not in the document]
ascjones: CI is failing because you need to update the path dependency version

ascjones:
https://crates.io/crates/contract-transcode

transcode/src/util.rs
```diff
@@ -0,0 +1,24 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
+
+/// Decode hex string with or without 0x prefix
+pub fn decode_hex(input: &str) -> Result<Vec<u8>, hex::FromHexError> {
+    if input.starts_with("0x") {
+        hex::decode(input.trim_start_matches("0x"))
```
gilescope, line 20: You don't need the if here. You can call this function regardless of if the string is prefixed with an 0x or not.

ascjones: Happy to approve a small PR which refactors this

PR #617, opened 2022-06-22
Bump `cargo_metadata` to `v0.15.0`

HCastano:
Looks like `v0.14.3` was yanked, which is causing problems if you try
and install `cargo-contract` from Git.


ascjones (approved):

ascjones (commented):
Dunno why all CIs are failing though

PR #633, opened 2022-07-06
Update linting nightly, fix UI tests

ascjones:


ascjones:
> Hey nice! How did you do the linting upgrade?
> 
> In the past I adapted our linting it to one of the examples in https://github.com/trailofbits/dylint/blob/master/examples/general/.

I used the commit hash from those examples and then fixed the compiler errors from there.

PR #635, opened 2022-07-06
[DO NOT MERGE] ci: test if cachepot works

TriplEight:
I've crafted a `contracts-ci-linux:cachepot` with `cargo install --git https://github.com/paritytech/cachepot --tag=0.1.0-rc.2 --features=redis`. To test if it's functional.

UPD: So far the problem is it doesn't refer to an existing Redis installation.
UPD1: Forgot to add all the `CACHEPOT_` variables to the project
image
Now it speaks to Redis

there's a lot of `Blocking waiting` in all the jobs

Pipeline with both rusty_cachier and cachepot https://gitlab.parity.io/parity/mirrors/cargo-contract/-/jobs/1664991 2m48s (cachepot doesn't add anything)
Pipeline without rusty_cachier https://gitlab.parity.io/parity/mirrors/cargo-contract/-/jobs/1666463 6m28s
Pipeline without rusty_cachier, with `CACHEPOT_LOG=trace` https://gitlab.parity.io/parity/mirrors/cargo-contract/-/jobs/1666632 7m21s
Pipeline with no caching https://gitlab.parity.io/parity/mirrors/cargo-contract/-/jobs/1666706 6m23s

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -44,8 +44,8 @@ workflow:
     - rusty-cachier environment check --gracefully
     - $(rusty-cachier environment inject)
     - rusty-cachier project mtime
-  # after_script:
-    # - rusty-cachier snapshot destroy
+  after_script:
+    - rusty-cachier snapshot destroy
 
 .docker-env:                       &docker-env
   image:
@@ -138,9 +138,12 @@ test-no-run-basic:
   <<:                              *docker-env
   <<:                              *collect-artifacts
   script:
+    - rusty-cachier snapshot create
     - cargo test --verbose --workspace --all-features --timings --no-run
     - mkdir -p ./artifacts/$CI_JOB_NAME
-    - cp target/cargo-timings/cargo-timing-*.html artifacts/$CI_JOB_NAME
+    - cp /cargo_target_dir/cargo-timings/cargo-timing-*.html artifacts/$CI_JOB_NAME
+    - rusty-cachier cache upload
+    # - cachepot -s
 
 test-no-run-cachier:
   stage:                           test
@@ -149,12 +152,13 @@ test-no-run-cachier:
   variables:
     RUSTY_CACHIER_SINGLE_BRANCH:     master
     RUSTY_CACHIER_DONT_OPERATE_ON_MAIN_BRANCH: "true"
+    # RUSTY_CACHIER_BYPASS:            "true"
   script:
     - rusty-cachier snapshot create
     - cargo test --verbose --workspace --all-features --timings --no-run
     - mkdir -p ./artifacts/$CI_JOB_NAME
     - cp /cargo_target_dir/cargo-timings/cargo-timing-*.html artifacts/$CI_JOB_NAME
-    - rusty-cachier snapshot destroy
+    - rusty-cachier cache upload
 
 test-no-run-cachepot:
   stage:                           test
@@ -184,8 +188,8 @@ test-no-run-both:
     - cargo test --verbose --workspace --all-features --timings --no-run
     - mkdir -p ./artifacts/$CI_JOB_NAME
     - cp /cargo_target_dir/cargo-timings/cargo-timing-*.html artifacts/$CI_JOB_NAME
+    - rusty-cachier cache upload
     - cachepot -s
-    - rusty-cachier snapshot destroy
 
 # test-new-project-template:
 #   stage:                           test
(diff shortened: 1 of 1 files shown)


Xanewok:
I'm trying to test this locally but I'm not sure if this is a good benchmark - without using any caching solution, I'm just trying to run
```
$ cargo test --workspace --all-features
```

locally, using latest nightly (`rustc 1.64.0-nightly (4d6d601c8 2022-07-26)`) and I'm getting *a lot* of rebuilds - I'm thinking that each test case tries to build some tool/driver (looking at the logs?) separately, stomping each other's artifacts and/or not sharing the work and we end up compiling the same thing over, and over.

Not only do I get rebuilds for each test case, the `cargo test` can't seem to re-use the local `target` anyway and does all the work again on subsequent `cargo test ...` runs.

How is the test suite supposed to work? Is this expected? Should we first look into sharing the artifacts in the test suite, before looking into introducing caching solutions here?

TriplEight:
@Xanewok 

> I'm getting a lot of rebuilds - I'm thinking that each test case tries to build some tool/driver (looking at the logs?) separately, stomping each other's artifacts and/or not sharing the work and we end up compiling the same thing over, and over.

This, I believe, is also one of the old cargo shortcomings -- when one test or crate requires one version of dependency and the other needs a different one, cargo recompiles it over the same artifacts. This is not for sure, so I suggest the html output of `cargo build --timings` can help us here. It looks like that https://meet.google.com/mio-sonp-ihb and there there should be repeated compilations of the same crates with different versions.
The problem is even worse when you run a job with cache, that cache firs runs into the artifacts which were created last.

TriplEight:
bot rebase

Xanewok:
I did some preliminary testing and here are the timings for a `cargo test ... --no-run` run:

| [Before cache population](https://gitlab.parity.io/parity/mirrors/cargo-contract/-/pipelines/207240) | b4b5bfa6fe06f7a9833372ccae1093d83916b500 |   | [After cache population](https://gitlab.parity.io/parity/mirrors/cargo-contract/-/pipelines/207246) | c5dde4ae3a1177fd6ec9f864e3184d45f3d9d99f |
|------------------------------------------------------------------------------------------------------|------------------------------------------|---|-----------------------------------------------------------------------------------------------------|------------------------------------------|
| Test-no-run-basic                                                                                    | 6m10s                                    |   | Test-no-run-basic                                                                                   | 6m11s                                    |
| Test-no-run-cachepot                                                                                 | 6m13s                                    |   | Test-no-run-cachepot                                                                                | 6m7s                                     |
| Test-no-run-cachier                                                                                  | 2m5s                                     |   | Test-no-run-cachier                                                                                 | 1m53s                                    |
| Test-no-run-both                                                                                     | 1m52s                                    |   | Test-no-run-both                                                                                    | 1m48s                                    |

Jobs were executed with `CACHEPOT_LOG=trace`, `RUST_LOG=trace` and with the `--timings` option.

I verified that cachepot recognizes cache misses and subsequent hits, although the speed-up is very negligible and almost within a margin of error. I can also confirm that this somewhat interacts okay-ish with the cachier and we get the cache hits as expected.

My educated guess is that the I/O overhead (whether that's from the console output or the artifact transfer itself) more or less trumps most of the CPU savings; will need to re-run this again but without the logging and compare the results.

rcny:
I've excluded `rusty-cachier` from the latest iterations of these test pipelines as `cargo build` parts of the `rusty-cachier` enabled jobs are completed almost instantly. The goal here is to check that cachepot works, not `rusty-cachier`. `rusty-cachier` with `cachepot` completes instantly, these measurements don't provide any meaningful feedback for our primary task.

I've settled on three jobs, `test-no-run-basic` & `test-no-run-cachepot` are intended to showcase the same as before, `test-no-run-cachepot-aligned` was added to check if `rusty-cachier`'s mtime, cargo_home & cargo_target_dir normalization facilities could change anything. `sparse-registry` unstable option was also added to get recent performance improvements for the crates.io index accessing.

Here what I've got. These are the average durations of ~5 of each mentioned job from [this](https://gitlab.parity.io/parity/mirrors/cargo-contract/-/pipelines/208682) pipeline:

image


Based on this data, I can observe that `test-no-run-cachepot` job is ~10% faster than `test-no-run-basic` on average in this setup. The jobs run in the isolated environment without any neighbors, in a serial fashion. There are no any meaningful performance improvements from the `rusty-cachier` preliminary commands, so I assume that cachepot doesn't care about mtime of source files (as opposed to Cargo). 

The host isn't overloaded, I don't observe any worrying `iowait`, LA or any relevant `iostat` output. As far as I'm concerned, the host isn't under any kind of a resource pressure while running the test jobs.

HCastano:
Feel free to re-open if you need this

Issue #643, opened 2022-07-12
Missing `CodeStored` Event in `cargo contract upload`


xermicus:
I think this is the same issue as #589 but for `upload`, I can just fix it in the same way as in #594?

Proposed fix in #644

PR #644, opened 2022-07-12
contract upload: emitting a warning instead of an error when the cont…

xermicus:
When uploading the same contract multiple times, no `CodeStored` event will be emitted anymore. So instead of failing the command, I propose to just issue a warning as hint for the user.

I think this very similar to: https://github.com/paritytech/cargo-contract/pull/594

Fixes https://github.com/paritytech/cargo-contract/issues/643


ascjones (approved):
LGTM

PR #663, opened 2022-07-28
Fix CI: Wrong path of binaryen

athei:
The download step of binaryen for Windows apparently uses the wrong path within the archive.


HCastano (approved):
Windows stage fails for a different reason now

Issue #525, opened 2022-04-23
Add information for reproducible builds to metadata


cmichi:
Needed to enable reproducible builds for https://github.com/paritytech/ink/issues/1226.

We need:
- rust version
- cargo contract version
- release vs. debug build
- wasm-opt version

## ToDo

- [ ] Emit the information to the metadata on `cargo contract build`
- [ ] Add `cargo contract verify [TARGET.contract]` which reads the information from the metadata, builds the current workspace according to it and compares the resulting `output.contract` file against the `TARGET`
- [ ] Check that no environment variables are passed through from the user to rustc
- [ ] Create follow-up issue for Solang to do the same.
- [ ] Create follow-up issue for ask! to do the same.

athei:
Don't we control the build through `cargo contract`. Meaning: We don't pass through any environment variables. If we do: We should filter everything out. Otherwise we would always miss something.

athei:
I updated the issue so that it says that we want to emit this data into the metadata. Reasoning is that this makes the life of everyone much easier without really preventing any use case: We **always** want to validate metadata **and** wasm anyways. Having a validated wasm with wrong metadata is just as dangerous as invalidated. Additionally, the information included are partly specific to ink! and should not leak to `pallet-contracts`.

The `verify` command in his first instantiation should only do a check whether the used nightly is the same as the one specified in the contract to be verified. On mismatch it should error out. Ultimately we want it to download the correct toolchain and use it.

HCastano:
I haven't been able to create reproducible build artifacts while working through #696
across the different machines I have. This is gonna be an update on my findings so far.

At the moment I'm building contracts manually using `cargo` instead of using
`cargo-contract` (but matching the settings used by `cargo-contract`). This reduces the
possible variability in the build introduced by other steps in the `cargo-contract` build
pipeline (e.g `wasm-opt`) (thanks for that suggestion Michi).

```bash
#!/bin/bash
set -eux

# Our default profile from https://github.com/paritytech/cargo-contract/blob/master/src/workspace/profile.rs#L30-L39
echo '[profile.release]
opt-level = "z"
lto = "fat"
codegen-units = 1
overflow-checks = true
panic = "abort"' >> Cargo.toml

# Our flags from https://github.com/paritytech/cargo-contract/blob/master/src/cmd/build.rs#L277-L302
export RUSTC_BOOTSTRAP=1
[...]

PR #680, opened 2022-08-08
Add Rust specific build info to metadata

HCastano:
Will close the metadata part of #525.

This is what the new metadata looks like:

```json
  "source": {
    "hash": "0xad8e1a3d7d1c44b4c7b3f715765ab8003e15aa2d8d1146a64cbc9b8f7e0a8df9",
    "language": "ink! 4.0.0-alpha.3",
    "compiler": "rustc 1.64.0",
    "build_info": {
      "build_mode": "Release",
      "cargo_contract_version": "2.0.0-alpha.4",
      "rustc_version": "stable-x86_64-unknown-linux-gnu",
      "wasm_opt_settings": {
        "keep_debug_symbols": false,
        "optimization_passes": "Z"
      }
    }
```

The stuff related to the `verify` command will be done in a follow-up.

Note: I've ended up removing the `wasm-opt` version from the `build_info` for now since there's
no easy way to get it from the new `wasm-opt` library. 

I'm going to assume that two installations of `cargo-contract` contain the same `wasm-opt` 
library (this only holds if both were installed with `--locked`). We should add it back in the future
in order to give users a better indication of what went wrong during the `verify` phase.


athei:
I don't think we should add the wasm-opt version. We should have a useable `wasm-opt` version as crate as early as next week. Then the version is described by the manifest alread.

HCastano:
> I don't think we should add the wasm-opt version. We should have a useable `wasm-opt` version as crate as early as next week. Then the version is described by the manifest alread.

Sure, but until that happens this is still needed. We can get rid of that field once that crate is published.

crates/cargo-contract/src/cmd/metadata.rs
```diff
@@ -73,6 +75,42 @@ struct ExtendedMetadataResult {
...
 }
 
+/// Information about the settings used to build a particular ink! contract.
+///
+/// Note that this should be an optional part of the metadata since it may not necessarily
+/// translate to other languages (e.g ask!, Solidity).
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+pub struct BuildInfo {
+    /// The stable version of `rustc` used to build the contract.
+    pub rustc_version: Version,
```
ascjones, line 85: Can/should we also add the toolchain which contains the architecture? e.g. `stable-x86_64-unknown-linux-gnu`. I remember there being some discussions about this affecting the wasm bytecode output, can't remember what the conclusions were.

crates/cargo-contract/src/cmd/build/mod.rs
```diff
@@ -650,7 +655,29 @@ pub(crate) fn execute(args: ExecuteArgs) -> Result<BuildResult> {
             &crate_metadata.contract_artifact_name,
         )?;
 
-        Ok(optimization_result)
+        let cargo_contract_version =
+            if let Ok(version) = Version::parse(env!("CARGO_PKG_VERSION")) {
```
ascjones, line 659: AFAIU `CARGO_PKG_VERSION` is only available at compile time (and also during `cargo run`). So would need to be captured in a `const` e.g. `const VERSION: &'static str = env!("CARGO_PKG_VERSION");`

crates/cargo-contract/src/cmd/build/mod.rs
```diff
@@ -650,7 +658,26 @@ pub(crate) fn execute(args: ExecuteArgs) -> Result<BuildResult> {
...
+            version
+        } else {
+            anyhow::bail!(
+                "Unable to parse version number for the currently running \
+                    `cargo-contract` binary."
+            );
+        };
+
+        let build_info = BuildInfo {
+            rustc_version: crate::util::rustc_toolchain()?,
```
ascjones, line 671: This field should be renamed `rust_toolchain` or so, as the actual version of `rustc`is in the `compiler` field/

crates/metadata/src/lib.rs
```diff
@@ -137,6 +139,11 @@ pub struct Source {
     /// with the metadata.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub wasm: Option<SourceWasm>,
+    /// Extra information about the environment in which the contract was built.
+    ///
+    /// Useful for producing deterministic builds.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub build_info: Option<Map<String, Value>>,
```
ascjones, line 146: Curious why we convert to loosely typed json `Map` here? Is that for alt. compilers e.g. `solang` to provide arbitrary data here?

ascjones (approved):

crates/metadata/src/lib.rs
HCastano: Yep, other compilers may want to include other info here. It's also useful for us if we decide that we need to add/remove info later on

PR #686, opened 2022-08-10
Bump `ink_*` crates to `v3.3.1`

HCastano:
Updates all the `ink_*` dependencies instead of just a couple (take that dependabot!).

Supersedes #684 and #685.

### DIFF
--- a/ink_linting/Cargo.lock
+++ b/ink_linting/Cargo.lock
@@ -57,24 +57,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0dde43e75fd43e8a1bf86103336bc699aa8d17ad1be60c76c0bdfd4828e19b78"
-dependencies = [
- "autocfg 1.1.0",
-]
-
-[[package]]
-name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
-
-[[package]]
-name = "base64"
-version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
 name = "bitflags"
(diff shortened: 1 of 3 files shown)


ink_linting/Cargo.lock
```diff
@@ -499,29 +454,47 @@ dependencies = [
...
+ "parity-scale-codec",
+ "rand",
+ "secp256k1",
+ "sha2",
+ "sha3",
+]
+
 [[package]]
 name = "ink_env"
-version = "3.0.0-rc8"
```
HCastano, line 510: ayy lmao

ink_linting/_Cargo.toml
```diff
@@ -28,11 +28,11 @@ regex = "1.5.4"
...
-ink_primitives = { git = "https://github.com/paritytech/ink", default-features = false }
-ink_metadata = { git = "https://github.com/paritytech/ink", default-features = false, features = ["derive"] }
-ink_env = { git = "https://github.com/paritytech/ink",  default-features = false }
-ink_storage = { git = "https://github.com/paritytech/ink",  default-features = false }
-ink_lang = { git = "https://github.com/paritytech/ink", default-features = false }
+ink_primitives = { version = "3.3", default-features = false }
+ink_metadata = { version = "3.3", default-features = false, features = ["derive"] }
+ink_env = { version = "3.3",  default-features = false }
+ink_storage = { version = "3.3",  default-features = false }
+ink_lang = { version = "3.3", default-features = false }
```
cmichi, line 35: What is your thinking here? This way the CI for `cargo-contract` will no longer alert us if we merge something into ink! `master` that breaks the linter.

HCastano: This crate was introduced before the ink! 3.0 release, so pulling Git deps looks like a remnant of that. 

More importantly though, the lockfile is using old RC deps, which is what ends up getting used in the [`test-dylint` CI step](https://gitlab.parity.io/parity/mirrors/cargo-contract/-/jobs/1732092), not ink! `master`.

cmichi: I would leave it to reference the ink! `master`, otherwise we might publish an ink! release that in the worst case breaks the last `cargo-contract` release. We would only notice after having published the ink! release.

For the lockfile: Hmm, the CI is not using it. I guess we committed it to have `--locked` working and forgot to update it.

athei: It seems wrong to test it in that way. If you want to test for that it should be a CI job on the ink! repository.

HCastano: I'm with Alex here. We'd see the linter breaking in the `examples-contract-build` step of the ink! CI anyways

PR #687, opened 2022-08-10
Wrap `wasm-opt` related functions in struct

HCastano:
This PR adds a `WasmOptHandler` struct, which wraps all the functions related to
`wasm-opt`. This lets us ensure that our `wasm-opt` binary is there and valid before we
even try and optimize a binary.

It also makes it easier to pull out certain information about the `wasm-opt` instance
we're using (such as the `wasm-opt` version, which I need for #680).


src/cmd/build.rs
```diff
@@ -250,6 +249,222 @@ impl CheckCommand {
...
+const WASM_OPT_INSTALLATION_SUGGESTION: &str =
+                "wasm-opt not found! Make sure the binary is in your PATH environment.\n\n\
+                We use this tool to optimize the size of your contract's Wasm binary.\n\n\
+                wasm-opt is part of the binaryen package. You can find detailed\n\
+                installation instructions on https://github.com/WebAssembly/binaryen#tools.\n\n\
+                There are ready-to-install packages for many platforms:\n\
+                * Debian/Ubuntu: apt-get install binaryen\n\
+                * Homebrew: brew install binaryen\n\
+                * Arch Linux: pacman -S binaryen\n\
+                * Windows: binary releases at https://github.com/WebAssembly/binaryen/releases";
```
cmichi, line 261: The indentation seems to mismatch the other indentations in the file.

src/cmd/build.rs
```diff
@@ -250,6 +249,222 @@ impl CheckCommand {
...
+/// A helpful struct for interacting with Binaryen's `wasm-opt` tool.
+struct WasmOptHandler {
+    /// The path to the `wasm-opt` binary.
+    wasm_opt_path: PathBuf,
+    /// The optimization level that should be used when optimizing the Wasm binary.
+    optimization_level: OptimizationPasses,
+    /// Whether or not to keep debugging information in the final Wasm binary.
+    keep_debug_symbols: bool,
+    /// The version number of the `wasm-opt` binary being executed.
+    _version: u32,
```
cmichi, line 272: Why prefix with `_`?

cmichi:
I think you should put `WasmOptHandler` in its own file, that's just the next consequence of abstracting away from what is happening in that handler and it will make the `build.rs` file less cluttered.

src/cmd/build.rs
HCastano: It's not used in this PR, but will be used in #680

cmichi (approved):
👍

src/wasm_opt.rs
```diff
@@ -0,0 +1,356 @@
...
+        let mut command = Command::new(self.wasm_opt_path.as_path());
+        command
+        .arg(dest_wasm.as_os_str())
+        .arg(format!("-O{}", self.optimization_level))
+        .arg("-o")
+        .arg(dest_optimized.as_os_str())
+        // the memory in our module is imported, `wasm-opt` needs to be told that
+        // the memory is initialized to zeroes, otherwise it won't run the
+        // memory-packing pre-pass.
+        .arg("--zero-filled-memory");
```
ascjones, line 115: Surely this whole block should be indented?

src/wasm_opt.rs
```diff
@@ -0,0 +1,356 @@
...
+        })?
+        .as_str()
+        .parse()
+        .map_err(|err| {
+            anyhow::anyhow!(
+                "Parsing version number failed with '{:?}' for '{:?}'",
+                err,
+                version_stdout
+            )
+        })?;
```
ascjones, line 233: Same here.

ascjones (approved):
LGTM, `rustfmt` still on a bender though

src/wasm_opt.rs
ascjones, line 115: Surely this whole block should be indented?

src/wasm_opt.rs
ascjones, line 233: Should also be indented

src/wasm_opt.rs
HCastano: RustFmt seems indifferent, but yeah I've indented it

PR #690, opened 2022-08-12
Fix windows dylint build

### LINKED ISSUE #649: Windows installation fails building `ink_linting`
Recently the windows CI has started to [fail ](https://github.com/paritytech/cargo-contract/runs/7322817986?check_suite_focus=true#step:8:691) with ` Error: environment variable not found: RUSTUP_TOOLCHAIN` from the `dylint-link` linker.

So the build.rs does in fact [remove this environment variable](https://github.com/paritytech/cargo-contract/blob/master/build.rs#L193), which is required so as not to override the required version for the child `cargo` invocation in the `ink_linting/rust-toolchain` file. However this has always worked previously and it also continues to work on Linux. My current working assumption is that there is a bug in the recent `1.25` release of `rustup` perhaps by not setting `RUSTUP_TOOLCHAIN` for the child process where it should. There seem to be a couple of other windows specific issues e.g. https://github.com/rust-lang/rustup/issues/3036, but not exactly the same issue we are having.

I have rolled back `rustup` to `1.24.3` on a local Windows machine and that seems to fix the issue, however there doesn't seem to be a way to control the version of `rustup` [installed via `actions-krs`](https://github.com/actions-rs/core/blob/master/src/commands/rustup.ts).

A couple of ideas of how to proceed:

1. Confirm issue is in `rustup` and find and fix it.
2. Remove the `cmd.env_remove("RUSTUP_TOOLCHAIN");` line in `build.rs` and instead add `+nightly-2022-06-30` (from the `rust-toolchain` file). This arg would take precedence over the current `RUST_TOOLCHAIN`. One issue with this is that `1.25` also has a [bug ](https://github.com/rust-lang/rustup/issues/3036) which currently prevents this.

ascjones:
Fixes #649 :crossed_fingers: 

Copied workaround from dylint source: https://github.com/trailofbits/dylint/blob/f5cb5b765573526bb08255a6c905c363ce461243/internal/src/cargo.rs#L52

Also see https://github.com/rust-lang/rustup/issues/3036#issuecomment-1193568646 and https://github.com/rust-lang/rustup/issues/3050

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -476,6 +476,7 @@ dependencies = [
  "contract-metadata",
  "contract-transcode",
  "current_platform",
+ "dirs",
  "heck",
  "hex",
  "impl-serde",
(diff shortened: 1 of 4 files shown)


cmichi (approved):
:medal_sports: :medal_military: :1st_place_medal:

athei (approved):

PR #696, opened 2022-08-18
Add `verify` command

HCastano:
Will close the `verify` subcommand part of #525.

Note that this command does not necessarily work across different operating systems
or CPU architectures (see the discussion in #525 for more details). As such it should
only be used in the context of a well known Docker container à la `srtool`. 

Current built on top of #680 since I want to make sure that we have all the correct info
in the metadata.

### DIFF
--- a/crates/cargo-contract/src/cmd/verify.rs
+++ b/crates/cargo-contract/src/cmd/verify.rs
@@ -66,22 +66,36 @@ impl VerifyCommand {
         file.read_to_string(&mut contents)?;
 
         let metadata: ContractMetadata = serde_json::from_str(&contents)?;
-        let build_info = metadata.source.build_info.as_ref().unwrap();
-        let build_info: BuildInfo =
-            serde_json::from_value(build_info.clone().into()).unwrap();
+        let build_info = if let Some(info) = metadata.source.build_info {
+            info
+        } else {
+            anyhow::bail!(
+                "\nThe metadata does not contain any build information which can be used to \
+                verify a contract."
+                .to_string()
+                .bright_yellow()
+            )
+        };
+
+        let build_info: BuildInfo = serde_json::from_value(build_info.clone().into())?;
+
+        tracing::debug!(
+            "Parsed the following build info from the metadata: {:?}",
+            &build_info,
+        );
 
         // 2. Call `cargo contract build` with the `BuildInfo` from the metadata.
         let expected_rustc_version = build_info.rustc_version;
         let rustc_version = rustc_version::version()
             .expect("`rustc` always has a version associated with it.");
 
-        anyhow::ensure!(
-            rustc_version == expected_rustc_version,
-            "You are trying to `verify` a contract using the `{rustc_version}` toolchain.\n\
+        let rustc_matches = rustc_version == expected_rustc_version;
+        let mismatched_rustc = format!(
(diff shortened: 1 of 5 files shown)


src/cmd/verify.rs
```diff
@@ -0,0 +1,135 @@
...
+impl VerifyCommand {
+    pub fn run(&self) -> Result<()> {
+        let manifest_path = ManifestPath::try_from(self.manifest_path.as_ref())?;
+
+        // 1. Read the given metadata, and pull out the `BuildInfo`
+        let mut file = File::open(&self.contract)?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+
+        let metadata: ContractMetadata = serde_json::from_str(&contents)?;
```
ascjones, line 61: Can also do `from_reader` to avoid the intermediate `String` e.g. https://github.com/paritytech/cargo-contract/blob/master/transcode/src/lib.rs#L171

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -0,0 +1,191 @@
...
+                &reference_wasm,
+                &built_wasm
+            );
+
+            anyhow::bail!(format!(
+                "\nFailed to verify the authenticity of {} contract againt the workspace \n\
+                found at {}.",
+                format!("`{}`", metadata.contract.name).bright_white(),
+                format!("{:?}", manifest_path.as_ref()).bright_white()).bright_red()
+            );
```
ascjones, line 179: It would be super useful to print out the `SourceCompiler` properties of both `reference_wasm` and `built_wasm` in case of non matching wasm. 

It's useful to know whether they are the same or different for figuring out why verification failed.

And while we are at it, any other properties we have available in `Source` e.g. `SourceLanguage` to catch different versions of `ink!`.

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -0,0 +1,191 @@
...
+                therefore we are unable to verify the contract."
+                .to_string()
+                .bright_yellow()
+            )
+        };
+
+        let fs_wasm = std::fs::read(built_wasm_path)?;
+        let built_wasm = SourceWasm::new(fs_wasm);
+
+        if reference_wasm != built_wasm {
```
ascjones, line 167: I believe we should also compare the contract ABI metadata itself? They should be identical too, and if not it might give clues as to what (if anything) is different.

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let build_result = execute(args)?;
+
+        // 3. Grab the built Wasm contract and compare it with the Wasm from the metadata.
+        let reference_wasm = if let Some(wasm) = metadata.source.wasm {
+            wasm
+        } else {
+            anyhow::bail!(
+                "\nThe metadata for the reference contract does not contain a Wasm binary,\n\
+                therefore we are unable to verify the contract."
```
ascjones, line 145: In this case couldn't we just grab the code hash and then compare that?

ascjones (commented):
Possibly a job for a follow up, but it would be really nice to enhance this by querying a node for the contract code deployed at an account and comparing it with the built contract.

SkymanOne:
Closing in favour of #1306

PR #697, opened 2022-08-19
Handle extra chars in wasm-opt version

ascjones:
https://gitlab.parity.io/parity/mirrors/smart-bench/-/jobs/1758132 fails with the string `"'\''wasm-opt' version 109 '(version_109)'\'"`


ascjones:
Duh I read the logs wrong :facepalm:

PR #703, opened 2022-08-23
Compile dylints when compiling the contract

athei:
This removes the brittle setup of compiling the dylints within our `build.rs` file. Instead, it patches the contracts manifest to run the dylint. The dylint  itself will live inside the ink! repository and is downloaded from there. The build is quick and cached within the `target` directory as all other artifacts.

## Future Work

We don't want to pull the lint from master but rather from a tag that corresponds to the used ink! version. However, we don't wait for this because that is only possible once the next ink! version is released.


src/workspace/manifest.rs
```diff
@@ -372,6 +372,37 @@ impl Manifest {
         Ok(self)
     }
 
+    pub fn with_dylint(&mut self) -> Result<&mut Self> {
+        let ink_dylint = {
+            let mut map = value::Table::new();
+            map.insert("git".into(), "https://github.com/paritytech/ink/".into());
+            map.insert("branch".into(), "at/linting".into());
```
ascjones, line 379: Reminder to change this before merging

src/cmd/build.rs
```diff
@@ -335,61 +333,47 @@ fn exec_cargo_for_wasm_target(
...
         //
-        // This is because the ink! dylint driver crate found in `dylint` uses a
-        // fixed Rust toolchain via the `ink_linting/rust-toolchain` file. By
-        // removing this env variable we avoid issues with different Rust toolchains
-        // interfering with each other.
-        ("CARGO_TARGET_DIR", None),
+        // This is because we build from a temporary directory (to patch the manifest) but still
+        // want the output to live at a fixed path. `cargo dylint` does not accept this information
+        // on the command line.
+        ("CARGO_TARGET_DIR", Some(target_dir.as_ref())),
```
ascjones, line 352: Brilliant this is what I was missing when I looked at doing this!

image

ascjones (approved):
:rocket:

src/workspace/manifest.rs
athei: Done

PR #709, opened 2022-08-25
Remove `nightly` override from decode test

HCastano:
I was confused as to why my local toolchain kept getting overriden from `stable` to
`nightly` after a test run. Turns out its from this decode test.


ascjones (approved):
I wonder what other things we can do to your environment from within a test :thinking: :imp:

HCastano:
> I wonder what other things we can do to your environment from within a test 🤔 👿

```rust
assert_cmd::Command::new("apt")
        .arg("install")
        .arg("sl")
        .assert()
        .success();
```

🚂

athei (approved):
Ahh damn that already bit me.

PR #721, opened 2022-09-01
Move to virtual manifest

athei:
Non virtual manifests are confusing:

* They clutter the root dir
* When not specifying `--workspace` only the main crate is build
* The concept of a main crate is a useless distinction


ascjones (approved):

metadata/LICENSE
```diff
@@ -1,203 +0,0 @@
-
```
HCastano, line 1: We still need to have this, don't we?

athei: Yes. Removed by mistake.

PR #722, opened 2022-09-01
Add `--output-json` for `call`, `instantiate` & `upload` commands

### LINKED ISSUE #682: Add `--output-json` for `call` and `instantiate` commands
Have you considered making the tool spit out some output with a schema, say JSON? This would help a lot when using the tool in various automated pipelines.

SkymanOne:
# Summary
As requested this PR closes #682, this PR adds support for printing out response from calling `cargo contract instantiate ...`. and `cargo contract call ...` outputs in JSON format by introducing  `--output-json` flag.

Previous human-readable format has been preserved.

# Examples
## Calling a contract with arg
```bash
cargo contract call --contract 5HLJobbpgd9EpJFySGDEanHkgh4nMaG6h1pFq9D2A5aTqibt --message inc --args 5 --suri //Alice --skip-confirm --output-json
{
  "verbosity": "Default",
  "estimated_gas": 74999922688,
  "events": [
    {
      "pallet": "Balances",
      "name": "Withdraw",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 86298156
          }
        }
      ]
    },
    {
      "pallet": "TransactionPayment",
      "name": "TransactionFeePaid",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "actual_fee",
          "value": {
            "UInt": 86298156
          }
        },
        {
          "name": "tip",
          "value": {
            "UInt": 0
          }
        }
      ]
    },
    {
      "pallet": "System",
      "name": "ExtrinsicSuccess",
      "fields": [
        {
          "name": "dispatch_info",
          "value": {
            "Map": {
              "weight": {
                "UInt": 7841193043
              },
              "class": {
                "Tuple": {
                  "ident": "Normal",
                  "values": []
                }
              },
              "pays_fee": {
                "Tuple": {
                  "ident": "Yes",
                  "values": []
                }
              }
            }
          }
        }
      ]
    }
  ]
}
```
## Calling a contract with `--dry-run` flag
```bash
cargo contract call --contract 5HLJobbpgd9EpJFySGDEanHkgh4nMaG6h1pFq9D2A5aTqibt --message inc --args 5 --suri //Alice --skip-confirm --dry-run --output-json
{
  "result": "Success!",
  "reverted": false,
  "data": "Unit",
  "gas_consumed": 7391961043,
  "gas_required": 74999922688,
  "storage_deposit": {
    "charge": "0x0"
  }
}
```
## Instantiating contract
```bash
{
  "contract": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9",
  "code_hash": "0xd73514bcae8754202a7cf32b2d4057ba9c95e7e82a3cb8f00b5ce1c64995ab70",
  "verbosity": "Default",
  "estimated_gas": 27371571412,
  "events": [
    {
      "pallet": "Balances",
      "name": "Withdraw",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 31189830465
          }
        }
      ]
    },
    {
      "pallet": "System",
      "name": "NewAccount",
      "fields": [
        {
          "name": "account",
          "value": {
            "Literal": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9"
          }
        }
      ]
    },
    {
      "pallet": "Balances",
      "name": "Endowed",
      "fields": [
        {
          "name": "account",
          "value": {
            "Literal": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9"
          }
        },
        {
          "name": "free_balance",
          "value": {
            "UInt": 100405000000
          }
        }
      ]
    },
    {
      "pallet": "Balances",
      "name": "Transfer",
      "fields": [
        {
          "name": "from",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "to",
          "value": {
            "Literal": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 100405000000
          }
        }
      ]
    },
    {
      "pallet": "Balances",
      "name": "Reserved",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 100405000000
          }
        }
      ]
    },
    {
      "pallet": "Balances",
      "name": "Reserved",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 536140000000
          }
        }
      ]
    },
    {
      "pallet": "Contracts",
      "name": "CodeStored",
      "fields": [
        {
          "name": "code_hash",
          "value": {
            "Hex": {
              "s": "d73514bcae8754202a7cf32b2d4057ba9c95e7e82a3cb8f00b5ce1c64995ab70",
              "bytes": [
                215,
                53,
                20,
                188,
                174,
                135,
                84,
                32,
                42,
                124,
                243,
                43,
                45,
                64,
                87,
                186,
                156,
                149,
                231,
                232,
                42,
                60,
                184,
                240,
                11,
                92,
                225,
                198,
                73,
                149,
                171,
                112
              ]
            }
          }
        }
      ]
    },
    {
      "pallet": "Contracts",
      "name": "Instantiated",
      "fields": [
        {
          "name": "deployer",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "contract",
          "value": {
            "Literal": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9"
          }
        }
      ]
    },
    {
      "pallet": "Balances",
      "name": "Transfer",
      "fields": [
        {
          "name": "from",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "to",
          "value": {
            "Literal": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 100020000000
          }
        }
      ]
    },
    {
      "pallet": "Balances",
      "name": "Reserved",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 100020000000
          }
        }
      ]
    },
    {
      "pallet": "Balances",
      "name": "Deposit",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "amount",
          "value": {
            "UInt": 27233000969
          }
        }
      ]
    },
    {
      "pallet": "TransactionPayment",
      "name": "TransactionFeePaid",
      "fields": [
        {
          "name": "who",
          "value": {
            "Literal": "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
          }
        },
        {
          "name": "actual_fee",
          "value": {
            "UInt": 3956829496
          }
        },
        {
          "name": "tip",
          "value": {
            "UInt": 0
          }
        }
      ]
    },
    {
      "pallet": "System",
      "name": "ExtrinsicSuccess",
      "fields": [
        {
          "name": "dispatch_info",
          "value": {
            "Map": {
              "weight": {
                "UInt": 3870509443
              },
              "class": {
                "Tuple": {
                  "ident": "Normal",
                  "values": []
                }
              },
              "pays_fee": {
                "Tuple": {
                  "ident": "Yes",
                  "values": []
                }
              }
            }
          }
        }
      ]
    }
  ]
}
```

## Instantiating a contract with dry-run
```bash
cargo contract instantiate --suri //Alice target/ink/incrementer.wasm --skip-dry-run --gas 27371571412 --args 0 --skip-confirm --dry-run --output-json
{
  "result": "Success!",
  "contract": "5EMUDqdfmSsyE7v3uksHtGRuEeQrgk7QteoJ8tXhqGANDCfG",
  "reverted": false,
  "data": "0x",
  "gas_consumed": 7360357127,
  "gas_required": 74999922688,
  "storage_deposit": {
    "charge": "0x2eaa42cc40"
  }
}
```
# Updates
- Added serialised `StorageDeposit` changes
- Generic serialisation of Module errors and Generic error
- Added an entry to the Changelog


SkymanOne:
Currently, JSON is only displayed when if the operation has been successful. Should JSON response be introduced for errors?

cmichi:
> Currently, JSON is only displayed when if the operation has been successful. Should JSON response be introduced for errors?

That would be ideal, then third-party tooling can be easier use `--output-json` for scripting. I would print the errors as JSON on `stderr`.

For context: The PR has conflicts with `master` since we intentionally decided to branch it off before the breaking metadata changes, which currently prevent these commands on `master` from being used against a Substrate node.

SkymanOne:
# Update
Error are now partially json-rendered. Only `instantiate` command support serialised error output, `call` command can not produce serialised errors because of [`Error` enum from `subxt` crate does not have `serde` derive](https://github.com/paritytech/subxt/issues/646)
# Example
```
cargo contract instantiate --suri //Charlie --code-hash 0xd73514bcae8754202a7cf32b2d4057ba9c95e7e82a3cb8f00b5ce1c64995ab70 --constructor new --args 0 --output-json
{
  "pallet": "Contracts",
  "error": "DuplicateContract",
  "docs": [
    "A contract with the same AccountId already exists."
  ]
}
```

SkymanOne:
# Update
Module and Generic errors are now serialised generically
## `ModuleError`
```json
{
  "module_error": {
    "pallet": "Contracts",
    "error": "DuplicateContract",
    "docs": [
      "A contract with the same AccountId already exists."
    ]
  }
}
```

## `GenericError`
```json
{
  "generic_error": {
    "error": "Rpc error: RPC call failed: ErrorObject { code: ServerError(1010), message: \"Invalid Transaction\", data: Some(RawValue(\"Inability to pay some fees (e.g. account balance too low)\")) }"
  }
}
```

crates/cargo-contract/src/cmd/extrinsics/mod.rs
```diff
@@ -238,6 +320,47 @@ fn error_details(error: &DispatchError, metadata: &subxt::Metadata) -> Result<St
...
+    pub error: String,
+    pub docs: Vec<String>,
+}
+
+#[derive(serde::Serialize)]
+pub struct GenericError {
+    error: String,
+}
+
+fn error_details_object(
```
ascjones, line 343: This should be a constructor on `ErrorVariant`, either `new` or somenthing like `from_dispatch_error`. Then, you could even get rid of `fn error_details` and use this instead. Just need to implement `Display` on `ErrorVariant` to give a string instead of json.

ascjones (commented):
WIP review

crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -103,16 +181,17 @@ pub fn display_events(
...
         }
+        events.push(event_entry);
     }
-    println!();
-    Ok(())
+
+    Ok(CallResult {
+        events,
+        verbosity,
+        output_type,
```
ascjones, line 194: This field looks not be used

crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -103,16 +181,17 @@ pub fn display_events(
...
+        ..Default::default()
```
ascjones, line 195: Prefer to explicitly initialise all fields.

crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -103,16 +181,17 @@ pub fn display_events(
...
+                let field = Field::new(field_name, decoded_field);
+                event_entry.fields.push(field);
             }
         }
+        events.push(event_entry);
     }
-    println!();
-    Ok(())
+
+    Ok(CallResult {
```
ascjones, line 191: This function could be made to a constructor for `CallResult`

crates/cargo-contract/src/cmd/extrinsics/events.rs
SkymanOne: I think using `Default::default()` allows in future override `Default` trait for types we want and specify the default value without changing manually editing multiple lines

ascjones: I disagree, if you add a new field which implements `Default` it will be easy to forget to initialize it properly. If you explicitly require each field to be initialized then the compiler will prevent this, if you just want the default value it's not much effort to add a `new_field: Default::default()`.

crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -19,100 +19,175 @@ use super::{
...
+#[derive(Default, serde::Serialize)]
+pub struct CallResult {
+    /// Instantiated contract hash
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub contract: Option<String>,
+    /// Instantiated code hash
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub code_hash: Option<String>,
+    /// The verbosity flags.
+    pub verbosity: Verbosity,
```
ascjones, line 79: I'm pretty sure we don't want to display the verbosity anyway, perhaps can be passed just to `display_events`

crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -19,100 +19,175 @@ use super::{
...
-    }
+/// Result of the contract call
+#[derive(Default, serde::Serialize)]
+pub struct CallResult {
+    /// Instantiated contract hash
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub contract: Option<String>,
+    /// Instantiated code hash
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub code_hash: Option<String>,
```
ascjones, line 77: A more strongly typed way to achieve this without having to set these optional fields, would be to defined a common "inner" struct:

```
#[derive(serde::Serialize)]
struct ExtrinsicResult {
  events; Vec<Event>
  ..
}
```
Then different top level structs with the varying details, flattening the inner struct:
```
#[derive(serde::Serialize)]
struct InstantiateWithCodeResult {
  contract: String,
  #[serde(flatten)]
  details: ExtrinsicResult,
}

#[derive(serde::Serialize)]
struct InstantiateResult {
  code_hash: String,
  #[serde(flatten)]
  details: ExtrinsicResult,
}
```

SkymanOne: `code_hash` is still an optional field because of this logic in `instantiate_with_code()`
```rust
// The CodeStored event is only raised if the contract has not already been uploaded.
let code_hash = result
    .find_first::<api::contracts::events::CodeStored>()?
    .map(|code_stored| code_stored.code_hash);
// <--snippet--> 
call_result.contract = Some(instantiated.contract.to_ss58check());
```

So, `contract` is always present in the struct, but `code_hash` field may be `None`. Is there any point defining common inner struct?

ascjones: You're right it is probably not necessary, the main issue is that we should avoid the pattern of having a `mut CallResult` and setting the values of the public fields.

The values should be set in the constructor

crates/cargo-contract/src/cmd/extrinsics/upload.rs
```diff
@@ -145,12 +149,11 @@ impl UploadCommand {
 
         let result = submit_extrinsic(client, &call, signer).await?;
 
-        display_events(
-            &result,
-            transcoder,
-            &client.metadata(),
-            &self.extrinsic_opts.verbosity()?,
-        )?;
+        let call_result =
```
ascjones, line 152: In case of an error returned from `submit_extrinsic`, we should also display the error as json, just like we do in the `call` extrinsic.

crates/cargo-contract/src/cmd/extrinsics/instantiate.rs
```diff
@@ -280,28 +285,46 @@ impl Exec {
...
+            CallResult::from_events(&result, &self.transcoder, &self.client.metadata())?;
+
+        call_result.estimated_gas = gas_limit;
 
         // The CodeStored event is only raised if the contract has not already been uploaded.
         let code_hash = result
             .find_first::<api::contracts::events::CodeStored>()?
             .map(|code_stored| code_stored.code_hash);
 
+        if let Some(code_hash) = code_hash {
```
ascjones, line 298: Same here: in case of an error returned from `submit_extrinsic`, we should also display the error as json, just like we do in the `call` extrinsic.

crates/cargo-contract/src/cmd/extrinsics/events.rs
ascjones: Ok I have made a refactoring PR to show how I might do it, along with some other stuff https://github.com/paritytech/cargo-contract/pull/730

Feel free to reject the PR or change it, it is just some ideas about refactorings I would do.

crates/cargo-contract/src/cmd/extrinsics/instantiate.rs
```diff
@@ -203,66 +215,59 @@ pub struct Exec {
 }
 
 impl Exec {
-    async fn exec(&self, code: Code, dry_run: bool) -> Result<()> {
+    async fn exec(&self, code: Code, dry_run: bool, is_json: bool) -> Result<()> {
```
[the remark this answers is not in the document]
ascjones: I have done that in https://github.com/paritytech/cargo-contract/pull/730

crates/cargo-contract/src/cmd/extrinsics/upload.rs
```diff
@@ -147,17 +147,22 @@ impl UploadCommand {
...
+                    DisplayEvents::from_events(&result, transcoder, &client.metadata())?;
+                println!(
+                    "{}",
+                    events.display_events(self.extrinsic_opts.verbosity()?)
+                );
+                let code_stored =
+                    result.find_first::<api::contracts::events::CodeStored>()?;
+                Ok(code_stored)
+            }
+            // TODO: add json output here too
```
ascjones, line 163: Please resolve this TODO

crates/cargo-contract/src/main.rs
```diff
@@ -486,14 +486,21 @@ fn main() {
...
+    };
     match exec(args.cmd) {
         Ok(()) => {}
         Err(err) => {
-            eprintln!(
-                "{} {}",
-                "ERROR:".bright_red().bold(),
-                format!("{:?}", err).bright_red()
-            );
+            if !suppress_err {
```
ascjones, line 497: I think we actually want to report the error as json rather than supressing it. At the moment for some errors we will get the pretty printed json at the call site, and all others will not be displayed at all.

crates/cargo-contract/src/cmd/extrinsics/upload.rs
SkymanOne: Not in the scope of this PR

ascjones: Even though the `upload` command is not explicitly mentioned, we should assume that the intent is to enable json output for all extrinsic commands and not just 2 out of the 3.

crates/cargo-contract/src/main.rs
SkymanOne: If don't do this, then the output result can be like this:
```json
{
  "module_error": {
    "pallet": "Contracts",
    "error": "DuplicateContract",
    "docs": [
      "A contract with the same AccountId already exists."
    ]
  }
}
{
  "error": "Pre-submission dry-run failed. Use --skip-dry-run to skip this step."
}
```

ascjones: Then we will need to rethink the approach. How about instead of printing the json error directly, we bubble it back up as an error?.

So instead of `eprintnln!("{}", json_error); Ok(())` you could do `(Err(anyhow!("{}", json_err))`. Then this would be printed correctly at the top level

crates/cargo-contract/src/cmd/extrinsics/upload.rs
ascjones: Thank you :pray:

crates/cargo-contract/src/main.rs
SkymanOne: I made the error propagate to the top level, all the way back to the `main()`, then I check if an error is an json format and print out, otherwise I wrap the message into a serializable object.

crates/cargo-contract/src/main.rs
```diff
@@ -486,15 +486,41 @@ fn main() {
...
     };
     match exec(args.cmd) {
         Ok(()) => {}
         Err(err) => {
-            if !suppress_err {
+            if is_json {
+                // error message can be either plain text or json string
+                // we need to check whether the error message is json object
+                let json_result =
+                    serde_json::from_str::<serde_json::Value>(&err.to_string());
```
ascjones, line 502: It's not great to serialize to json at the error site and then deserialize again at the top level. I think we can do better with a strongly typed approach see https://github.com/paritytech/cargo-contract/pull/736

PR #730, opened 2022-09-07
Suggested refactorings for PR #722 (`--output-json`)

ascjones:
Some suggested refactorings for #722, feel free to accept or reject as you please. **NOT TESTED**, it compiles only.

- Proper handling and display of `subxt::Error` returned from `submit_extrinsic`, consistent with the dry run
- Removes some duplication in `instantiate` for displaying the extrinsic result.
- Moves `is_json` to a field in `Instantiate::exec`
- Moves unrelated fields from `events.rs` and moves them to their local and relevant extrinsics
- Removes `estaimated_gas` field, it is an input not an output.


SkymanOne:
It is definitely an improvement over my changes. I tested run it to my machine. Thanks!

PR #740, opened 2022-09-13
Update contracts node metadata, define Weight type

ascjones:
The latest version of `substrate-contracts-node` imports the new `Weight` type which is now used for the `gas_limit` parameter in contract extrinsics. This PR updates the metadata to reflect this, as well as providing a local copy of the type for convenience (implementing `From<u64>`)


crates/cargo-contract/src/cmd/extrinsics/runtime_api/mod.rs
```diff
@@ -17,4 +17,22 @@
...
+pub struct Weight {
+    /// The weight of computational time used based on some reference hardware.
+    ref_time: u64,
+}
+
+impl From<u64> for Weight {
+    fn from(ref_time: u64) -> Self {
+        Self { ref_time }
+    }
+}
```
athei, line 38: I don't think we should do that given that we expect further fields will be added. I think adding a `from_ref_time` would be more appropriate.

athei (approved):

PR #750, opened 2022-09-19
denominated units for balances in events

### LINKED ISSUE #673: Denominated units for Balances
Extracted from #432.

Any balances should be displayed denominated as the currency of the chain e.g. 100CAN, 1MCAN, instead of e.g. 10000000000000000000000000000

## Details

By default all balances are represented by a `u128` which is the smallest unit of a chain's balance e.g. in Polkadot it is called a [`Planck`](https://wiki.polkadot.network/docs/redenomination).

So in this case an account with 1 DOT == 10_000_000_000 Plancks.

At the moment we would display the big number which is obviously not very user friendly.

So we should convert the raw units into the denominated token. You can query the[ `system_properties` ](https://github.com/paritytech/substrate/blob/e0ccd008fe8bfaf29357ea87561e60f3baaae08c/client/rpc-api/src/system/mod.rs#L52) rpc to fetch arbitrary JSON encoded properties defined in the chainspec. By convention it may contain the following useful properties:
```
"properties": {
    "tokenDecimals": 12,
    "tokenSymbol": "KSM"
  },
```
These can be used to convert the raw units into more human readable balances.

If those properties are not present we should consider defaulting to some sensible defaults. I suggest looking at how `polkadot.js` handles this for inspiration: https://github.com/polkadot-js/api/search?q=tokenDecimals

## Displaying it

The other thing to figure out is how to identify where such a `Balance` type is being used, and then to decode it into a custom type that can display the required conversion.

Suggest looking at how other custom display types are registered e.g. with `AccountId` and `Hash` types: https://github.com/paritytech/cargo-contract/blob/f903c0349ba97b0219e14efcd81f67f4c48733eb/transcode/src/lib.rs#L154-L156

The slightly tricky thing might be that `Balance` is typically a primitive `u128` type and not its own type, so it may be required to check the `typeName` field of the type metadata to see if it is `Balance` or `T::Balance`, since usually a type alias or associated type is used for this. Needs some investigation though.

SkymanOne:
Closes #673 

Example of denomination of balances when running a call on substrate contracts node:
```
       Event Balances ➜ Transfer
         from: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
         to: 5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9
         amount: 100.2nDOT
       Event Balances ➜ Reserved
         who: 5FUdyfmrdbk1H5F1E4HnweiQjENqwAWnPxH2Mkty6LeK5at9
         amount: 100.2nDOT
       Event Balances ➜ Deposit
         who: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
         amount: 188.457902μDOT
       Event TransactionPayment ➜ TransactionFeePaid
         who: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
         actual_fee: 4.23927238nDOT
         tip: 0DOT
```

# Update
Balances can also be inputed in a denominated format (i.e. `500MDOT` for 500 x 10^6 x 10^`<tokenDecimals>`DOT).
The token symbol must match the one specified in the running local chain. If the chain config does not have a token symbol and/or decimals. `UNIT` is used as a default symbol and 12 decimals are used for denomination

This includes:
- ` --value` for `call` and `instantiate` commands
-  `--storage_deposit_limit` for any command

Includes fixes addressed in #751

### DIFF
--- a/crates/cargo-contract/src/cmd/extrinsics/balance.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/balance.rs
@@ -41,16 +41,16 @@ use anyhow::{
 /// Represents different formats of a balance
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum BalanceVariant {
-    /// Default format: no symbol, no denomination
+    /// Default format: no symbol, no token_decimals
     Default(Balance),
-    /// Denominated format: symbol and denomination are present
+    /// Denominated format: symbol and token_decimals are present
     Denominated(DenominatedBalance),
 }
 
 #[derive(Debug, Clone)]
 pub struct TokenMetadata {
-    /// Number of denomination used for denomination
-    pub denomination: usize,
+    /// Number of token_decimals used for denomination
+    pub token_decimals: usize,
     /// Token symbol
     pub symbol: String,
 }
@@ -80,8 +80,8 @@ impl TokenMetadata {
 
         let default_decimals = json!(12);
         let default_units = json!("UNIT");
-        let denomination = sys_props
-            .get("tokenDecimal")
+        let token_decimals = sys_props
+            .get("tokenDecimals")
             .unwrap_or(&default_decimals)
             .as_u64()
             .context("error converting decimal to u64")?
@@ -92,7 +92,7 @@ impl TokenMetadata {
             .as_str()
             .context("error converting symbol to string")?;
         Ok(Self {
-            denomination,
+            token_decimals,
             symbol: symbol.to_string(),
         })
     }
@@ -164,7 +164,7 @@ impl BalanceVariant {
     ///  use anyhow::{Result, Ok};
     ///  let decimals = 6;
     ///  let tm = TokenMetadata {
-    ///        denomination: decimals,
+    ///        token_decimals: decimals,
     ///        symbol: String::from("DOT"),
     /// };
     /// let sample_den_balance = Balance::Denominated(DenominatedBalance {
@@ -181,7 +181,7 @@ impl BalanceVariant {
     ///  use anyhow::{Result, Ok};
     ///  let decimals = 6;
     ///  let tm = TokenMetadata {
-    ///        denomination: decimals,
+    ///        token_decimals: decimals,
     ///        symbol: String::from("DOT"),
     /// };
     /// let sample_den_balance = Balance::Denominated(DenominatedBalance {
@@ -197,7 +197,7 @@ impl BalanceVariant {
         match self {
             BalanceVariant::Default(balance) => Ok(*balance),
             BalanceVariant::Denominated(den_balance) => {
-                let zeros: usize = (token_metadata.denomination as isize
+                let zeros: usize = (token_metadata.token_decimals as isize
                     + match den_balance.unit {
                         UnitPrefix::Giga => 9,
                         UnitPrefix::Mega => 6,
@@ -228,9 +228,9 @@ impl BalanceVariant {
     ///
     /// I takes `value` of `Into<u128>` and [TokenMetadata]
     /// and calculates the value in an denominated format
-    /// by manipulating the denomination.
+    /// by manipulating the token_decimals.
     ///
-    /// If the number is divisible by 10^(`denomination` + `unit_zeros`),
+    /// If the number is divisible by 10^(`token_decimals` + `unit_zeros`),
     /// It sets the [UnitPrefix] and divides the `value` into `Decimal`
     ///
     /// If no [TokenMetadata] was present, than that means
@@ -242,7 +242,7 @@ impl BalanceVariant {
     ///  use anyhow::{Result, Ok};
     ///  let decimals = 10;
     ///  let tm = TokenMetadata {
-    ///        denomination: decimals,
+    ///        token_decimals: decimals,
     ///        symbol: String::from("DOT"),
     /// };
     /// let sample_den_balance = BalanceVariant::Denominated(DenominatedBalance {
@@ -271,13 +271,13 @@ impl BalanceVariant {
 
             let number_of_digits = n.to_string().len();
 
-            let giga_units_zeros = token_metadata.denomination + 9;
-            let mega_units_zeros = token_metadata.denomination + 6;
-            let kilo_units_zeros = token_metadata.denomination + 3;
-            let one_unit_zeros = token_metadata.denomination;
-            let milli_units_zeros = token_metadata.denomination.checked_sub(3);
-            let micro_units_zeros = token_metadata.denomination.checked_sub(6);
-            let nano_units_zeros = token_metadata.denomination.checked_sub(9);
+            let giga_units_zeros = token_metadata.token_decimals + 9;
+            let mega_units_zeros = token_metadata.token_decimals + 6;
+            let kilo_units_zeros = token_metadata.token_decimals + 3;
+            let one_unit_zeros = token_metadata.token_decimals;
+            let milli_units_zeros = token_metadata.token_decimals.checked_sub(3);
+            let micro_units_zeros = token_metadata.token_decimals.checked_sub(6);
+            let nano_units_zeros = token_metadata.token_decimals.checked_sub(9);
 
             let multiple: Decimal;
             let unit: UnitPrefix;
@@ -404,7 +404,7 @@ mod tests {
     #[test]
     fn balance_variant_denominated_success() {
         let tm = TokenMetadata {
-            denomination: 10,
+            token_decimals: 10,
             symbol: String::from("DOT"),
         };
         let bv = BalanceVariant::from_str("500MDOT").expect("successful parsing. qed");
@@ -418,7 +418,7 @@ mod tests {
     fn balance_variant_denominated_equal() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 500 * 1_000_000 * 10_000_000_000;
@@ -431,7 +431,7 @@ mod tests {
     fn balance_variant_denominated_equal_fraction() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005_000_000_000_000_000;
@@ -444,7 +444,7 @@ mod tests {
     fn balance_variant_denominated_equal_small_units() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005_000;
@@ -456,7 +456,7 @@ mod tests {
     fn smallest_value() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 1;
@@ -471,7 +471,7 @@ mod tests {
         // which results in value less than zero
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let bv =
@@ -484,7 +484,7 @@ mod tests {
     fn giga() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005_000_000_000_000_000_000;
@@ -497,7 +497,7 @@ mod tests {
     fn kilo() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005_000_000_000_000;
@@ -510,7 +510,7 @@ mod tests {
     fn unit() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005_000_000_000;
@@ -523,7 +523,7 @@ mod tests {
     fn milli() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005_000_000;
@@ -535,7 +535,7 @@ mod tests {
     fn micro() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005_000;
@@ -547,7 +547,7 @@ mod tests {
     fn nano() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_005;
@@ -560,7 +560,7 @@ mod tests {
     fn different_digits() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 5_235_456_210_000_000;
@@ -571,10 +571,10 @@ mod tests {
     }
 
     #[test]
-    fn non_standard_denomination() {
+    fn non_standard_token_decimals() {
         let decimals = 10;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let balance: Balance = 50_015_000_000_000;
@@ -587,7 +587,7 @@ mod tests {
     fn small_number_of_decimals_zero() {
         let decimals = 6;
         let tm = TokenMetadata {
-            denomination: decimals,
+            token_decimals: decimals,
             symbol: String::from("DOT"),
         };
         let bv = BalanceVariant::from_str("0.4μDOT").expect("successful parsing. qed");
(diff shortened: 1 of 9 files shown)


SkymanOne:
Currently blocked by [#654](https://github.com/paritytech/subxt/pull/654) in subxt repo

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,274 @@
...
+    ///
+    /// Throws Error if [BalanceVariant::Denominated(String)] is in an incorrect format.
+    pub fn denominate_balance(&self, token_metadata: &TokenMetadata) -> Result<Balance> {
+        match self {
+            BalanceVariant::Default(balance) => Ok(*balance),
+            BalanceVariant::Denominated(input) => {
+                if let Some(balance_str) =
+                    input.strip_suffix(&format!("k{}", token_metadata.symbol))
+                {
+                    let balance: f64 = balance_str.parse()?;
```
ascjones, line 85: We should avoid using floats for dealing with currency amounts, see https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency.

I have used https://docs.rs/rust_decimal/latest/rust_decimal/ before, that is worth a look.

ascjones (changes requested):

crates/cargo-contract/src/cmd/extrinsics/balance.rs
SkymanOne: Thanks! I used `Decimal` for `denominate_balance()`. I thought of using it also in `from<from<T: Into<u128>>>()`, but `Decimal::new()` only takes `i64` and `u32` according to [docs](https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#method.new), so that would imply using `TryInto` which would add unnecessary type conversions and `Result` as a return type

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,289 @@
...
+impl BalanceVariant {
+    /// Converts BalanceVariant into Balance.
+    ///
+    /// Throws Error if [BalanceVariant::Denominated(String)] is in an incorrect format.
+    pub fn denominate_balance(&self, token_metadata: &TokenMetadata) -> Result<Balance> {
+        match self {
+            BalanceVariant::Default(balance) => Ok(*balance),
+            BalanceVariant::Denominated(input) => {
+                if let Some(balance_str) =
+                    input.strip_suffix(&format!("k{}", token_metadata.symbol))
```
ascjones, line 91: Too much repetition in this method, making the calculations unclear. We should separate parsing from the calculations by doing it up front in a `FromStr` impl. You could flatten it out to something like:

```
#[derive(Debug, Clone)]
pub struct Balance {
    value: u128,
    unit: Option<UnitPrefix>,
    symbol: Option<String>,
}

#[derive(Debug, Clone)]
pub enum UnitPrefix {
    Giga,
    Mega,
    Kilo,
    One,
    Milli,
    Micro,
    Nano,
}
```

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,289 @@
...
+                    let balance = Decimal::from_str(input)?;
+                    Ok(<Balance>::try_from(
+                        balance * Decimal::from(token_metadata.denomination),
+                    )?)
+                }
+            }
+        }
+    }
+
+    /// Display token units in a denominated format.
```
ascjones, line 135: Please add some docs for how this method should behave, perhaps with some doctest examples.

Also we need some comprehensive unit tests for this, there aren't any at the moment. These are important because this method is controlling the value displayed to the user which *must* be accurate and with no loss of precision.

ascjones (commented):
The main thing is the code in `balance.rs` needs to be tightened up a bit, needs more tests since there are bugs there.

Also I think we are missing decoding `Balance` for contract events, for that the `decode_contract_event` would need to check for fields with the type name `Balance`: https://github.com/paritytech/cargo-contract/blob/05494e2bd62b48c6b79b99bd91c0657ee8656103/crates/transcode/src/lib.rs#L250

crates/cargo-contract/src/cmd/extrinsics/balance.rs
ascjones: One issue is that there could be some loss of precision with `rust_decimal`, it seems the [max value](https://github.com/paupino/rust-decimal/blob/master/src/decimal.rs#L201) is 96 bits https://github.com/paupino/rust-decimal/blob/master/src/decimal.rs#L464.

Thinking out loud we need:
1. Carefully protect against loss of precision when performing conversions in `denominate_balance`, although this might be accounted for already in overflow panics. If this is the case these should be tested and handled gracefully, because we would be relying on the underlying library panicking on overflow rather than, say wrapping around.
2. Research alternative libraries which would allow storing up to `u128` precision decimals?
3. Define a custom built `Decimal` type e.g. 
```
pub struct Decimal {
    value: u128,
    scale: i8,
}
```
And then provide some specialised methods to scale by exponents, which is the main thing we need.

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,289 @@
...
+        assert!(
+            parse_balance("500%").is_err(),
+            "expected to fail parsing incorrect balance"
+        );
+    }
+
+    #[test]
+    fn balance_variant_denominated_success() {
+        let tm = TokenMetadata {
+            denomination: 12,
```
ascjones, line 228: I know `denomination` is not used in the parsing but this is an invalid value for this, it seems that in other cases it is `1_000_000_000_000`

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,289 @@
...
+        };
+        let balance: Balance = 500 * 1_000_000 * denomination;
+        let bv = parse_balance("500MDOT").expect("successful parsing. qed");
+        let balance_parsed = bv.denominate_balance(&tm).expect("successful parsing. qed");
+        assert_eq!(balance, balance_parsed);
+    }
+
+    #[test]
+    fn balance_variant_denominated_equal_fraction() {
+        let denomination: u128 = format!("1{}", "0".repeat(12)).parse().unwrap();
```
ascjones, line 266: All tests for the `denominate_balance` are assuming `12` decimals. There should be some other values for this, there are some bugs this will show.

crates/cargo-contract/src/cmd/extrinsics/balance.rs
SkymanOne: I agree that we can structure denominated balance. However, I think there should be two struct `Balance` and `DenominatedBalance` because they are of different types: `DenominatedBalance` is represented as decimal, it has a a point like `45.234234MDOT` whereas `Balance` is of `u128` type

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -43,146 +41,292 @@ pub enum BalanceVariant {
...
             .context("error converting symbol to string")?;
-        let denomination: u128 = format!("1{}", "0".repeat(decimals as usize)).parse()?;
         Ok(Self {
             denomination,
             symbol: symbol.to_string(),
         })
     }
 }
 
+impl TryFrom<String> for DenominatedBalance {
```
ascjones, line 98: If you change this to  `FromStr` you can replace `parse_balance` with using plain `value_parser` (in clap `v4`)

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -43,146 +41,292 @@ pub enum BalanceVariant {
...
+            let mega_units_zeros = token_metadata.denomination + 6;
+            let kilo_units_zeros = token_metadata.denomination + 3;
+            let one_unit_zeros = token_metadata.denomination;
+            let milli_units_zeros = token_metadata.denomination - 3;
+            let micro_units_zeros = token_metadata.denomination - 6;
+            let nano_units_zeros = token_metadata.denomination - 9;
+
+            let multiple: Decimal;
+            let unit: UnitPrefix;
+            if number_of_digits > giga_units_zeros {
```
ascjones, line 266: It would be clearer to use range matching here, and remove duplication where possible (e.g.)

```
let (zeros, unit) = match number_of_digits {
  0..nano_units_zeros => (giga_units_zeros, UnitPrefix::Nano),
  ...
  giga_units_zeros.. => (giga_units_zeros, UnitPrefix::Giga)
}

// remove duplication of this 
let multiple = Decimal::from_str_exact(..._
```

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -43,146 +41,292 @@ pub enum BalanceVariant {
...
+                ))?;
+                unit = UnitPrefix::Mega;
+            } else if number_of_digits <= mega_units_zeros
+                && number_of_digits > kilo_units_zeros
+            {
+                multiple = Decimal::from_str_exact(&format!(
+                    "1{}",
+                    "0".repeat(kilo_units_zeros)
+                ))?;
+                unit = UnitPrefix::Mega;
```
ascjones, line 287: We need some tests for this method, this and other `Mega` looks to be wrong

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -43,146 +41,292 @@ pub enum BalanceVariant {
...
     /// Denominated format: symbol and denomination are present
-    Denominated(String),
+    Denominated(DenominatedBalance),
 }
 
 #[derive(Debug, Clone)]
 pub struct TokenMetadata {
     /// Number of denomination used for denomination
-    pub denomination: u128,
+    pub denomination: usize,
```
ascjones, line 50: This should probably be renamed to `token_decimals` now

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -43,146 +41,292 @@ pub enum BalanceVariant {
...
 impl TokenMetadata {
     /// Query [TokenMetadata] through the node's RPC
     pub async fn query(client: &Client) -> Result<Self> {
         let sys_props = client.rpc().system_properties().await?;
 
         let default_decimals = json!(12);
         let default_units = json!("UNIT");
-        let decimals = sys_props
+        let denomination = sys_props
             .get("tokenDecimal")
```
ascjones, line 81:
```suggestion
            .get("tokenDecimals")
```
See https://github.com/paritytech/substrate/blob/72b9c36d66e351af9d90cbf8c37e2cf90dd59582/client/chain-spec/res/chain_spec.json#L5k

ascjones (commented):
The handling of the decimal parsing is much better now :+1: 

The conversion from raw `u128` to `DenominatedBalance` needs some testing and tidying up.

One big thing that looks to be missing is the handling of `Balance` values when actually interacting with contracts. We should be able to use a denominated balance when invoking e.g. https://github.com/paritytech/ink/blob/master/examples/erc20/lib.rs#L130.

And also when displaying the resulting contract `Transfer` event: https://github.com/paritytech/ink/blob/master/examples/erc20/lib.rs#L27.

This needs to be done in https://github.com/paritytech/cargo-contract/blob/05494e2bd62b48c6b79b99bd91c0657ee8656103/crates/transcode/src/lib.rs#L250

crates/cargo-contract/src/cmd/extrinsics/balance.rs
SkymanOne:
```
exclusive range pattern syntax is experimental
see issue #37854 <https://github.com/rust-lang/rust/issues/37854> for more information
```

ascjones: Nonetheless you can remove some duplication there even with the conditionals

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,670 @@
...
+            let micro_units_zeros = token_metadata.token_decimals.checked_sub(6);
+            let nano_units_zeros = token_metadata.token_decimals.checked_sub(9);
+
+            let multiple: Decimal;
+            let unit: UnitPrefix;
+            if (giga_units_zeros + 1..).contains(&number_of_digits) {
+                multiple = Decimal::from_str_exact(&format!(
+                    "1{}",
+                    "0".repeat(giga_units_zeros)
+                ))?;
```
ascjones, line 287: This line is repeated in each conditional branch, prefer to return the no of zeros from the conditional and then do the `Decimal::from_str_exact` once at the end.

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,670 @@
...
+                    "0".repeat(kilo_units_zeros)
+                ))?;
+                unit = UnitPrefix::Kilo;
+            } else if (one_unit_zeros + 1..=kilo_units_zeros).contains(&number_of_digits)
+            {
+                multiple =
+                    Decimal::from_str_exact(&format!("1{}", "0".repeat(one_unit_zeros)))?;
+                unit = UnitPrefix::One;
+            } else if milli_units_zeros.is_some()
+                && (milli_units_zeros.unwrap() + 1..=one_unit_zeros)
```
ascjones, line 311: Instead of unwrapping at the usage site, maybe use an `expect` at the declaration site that we don't expect `token_deciamls` < 3. Although maybe we should  allow it just any high precision values will not be displayed as denominated

ascjones (commented):
`BalanceVariant::from` still needs some tidying up.

For some inspiration, have a look at [`formatBalance`](https://github.com/polkadot-js/common/blob/c92d5cfb20d92492bb6fdf5a6382d52e8b9f7b49/packages/util/src/format/formatBalance.ts#L88) that polkadot.js uses. It does so in a fairly concise manner. See e.g. [`calcSi`](https://github.com/polkadot-js/common/blob/e30bac1ed69d1d9880709541409f0761b5589e76/packages/util/src/format/si.ts#L49)

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -89,16 +91,31 @@ impl TokenMetadata {
...
 }
 
-impl TryFrom<String> for DenominatedBalance {
-    type Error = anyhow::Error;
+impl FromStr for BalanceVariant {
+    type Err = anyhow::Error;
+
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
+        let input = input.replace('_', "");
+        if input.contains('.') || input.ends_with(|ch: char| ch.is_alphabetic()) {
```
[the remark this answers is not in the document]
SkymanOne: Could you please elaborate where the ambiguity is caused here? I assumed if a user explicitly specifies a value with a decimal point, this already implies that denominated units are used. It would be odd if users normally enter `1_000_000.0`. Nonetheless, if user decided to add `.0` at the end, it will still work fine, the string will go through the parsing and result in `1_u128` when encoding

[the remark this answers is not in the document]
ascjones: The conditional `if input.contains('.') || input.ends_with(|ch: char| ch.is_alphabetic())` gives the impression that just a decimal point is enough to specify a denominated value.

However I just ran a test:
```
assert_eq!(
    BalanceVariant::from_str("1").unwrap(),
    BalanceVariant::from_str("1.0").unwrap()
)
```
And the second one fails because `DenominatedBalance::from_str` in fact does require a token symbol. So it is okay then, requires a test case perhaps.

The conditional here is a bit confusing because it is doing sort of a precheck that it *might* be Denominated balance and then trying to parse it. Alternative might be to do something like:

```
        Ok(match input.parse::<Balance>() {
            Ok(balance) => BalanceVariant::Default(balance),
            Err(_) => BalanceVariant::Denominated(DenominatedBalance::from_str(
                &input,
            )?)
        })
```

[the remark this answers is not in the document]
SkymanOne: Okay. That makes sense, I will rewrite and separate parsing logic

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,673 @@
...
+        let symbols = value
+            .trim_start_matches(|ch: char| ch.is_numeric() || ch == '.' || ch == ',');
+        let unit_char = symbols
+            .chars()
+            .next()
+            .context("no units or symbols present")?;
+        let unit: UnitPrefix = match unit_char {
+            'G' => UnitPrefix::Giga,
+            'M' => UnitPrefix::Mega,
+            'K' => UnitPrefix::Kilo,
```
ascjones, line 132:
```suggestion
            'k' => UnitPrefix::Kilo,
```
https://en.wikipedia.org/wiki/International_System_of_Units#Prefixes

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -0,0 +1,673 @@
...
+                let multiple =
+                    Decimal::from_str_exact(&format!("1{}", "0".repeat(zeros)))?;
+                let fract_scale = den_balance.value.fract().scale();
+                let mantissa_difference = zeros as isize - fract_scale as isize;
+                if mantissa_difference < 0 {
+                    return Err(anyhow!(
+                        "Given precision of a Balance value is higher than allowed"
+                    ))
+                }
+                let balance: Balance = (den_balance.value * multiple).try_into()?;
```
ascjones, line 223: Better to call `checked_mul` and handle an overflow cleanly?

crates/cargo-contract/src/cmd/extrinsics/balance.rs
```diff
@@ -220,7 +220,11 @@ impl BalanceVariant {
                         "Given precision of a Balance value is higher than allowed"
                     ))
                 }
-                let balance: Balance = (den_balance.value * multiple).try_into()?;
+                let balance: Balance = den_balance
+                    .value
+                    .checked_mul(multiple)
+                    .context("error while converting balance to raw format.")?
```
ascjones, line 226: Specifically it would be an overflow in this case

ascjones (approved):
LGTM

Issue #734, opened 2022-09-09
Cannot pass `--storage-deposit-limit ` to `cargo contract call`


pmikolajczyk41:
# Problem statement

Calling a contract with `--storage-deposit-limit` specified fails with a rather useless error:
```shell
> cargo contract call --url "$NODE" --suri "$BOB" \
    --skip-confirm \
    --contract "$CONTRACT_ADDRESS" \
    --gas "$GAS_LIMIT" \
    --message "add" \
    --storage-deposit-limit "$STORAGE_LIMIT"

ERROR: Rpc error: RPC call failed: ErrorObject { code: ServerError(1002), message: "Verification Error: Runtime error: Execution failed: Error calling api function: Failed to convert parameter `tx` from node to runtime of validate_transaction", data: Some(RawValue("RuntimeApi(\"Execution failed: Error calling api function: Failed to convert parameter `tx` from node to runtime of validate_transaction\")")) }

Caused by:
    0: RPC call failed: ErrorObject { code: ServerError(1002), message: "Verification Error: Runtime error: Execution failed: Error calling api function: Failed to convert parameter `tx` from node to runtime of validate_transaction", data: Some(RawValue("RuntimeApi(\"Execution failed: Error calling api function: Failed to convert parameter `tx` from node to runtime of validate_transaction\")")) }
    1: RPC call failed: ErrorObject { code: ServerError(1002), message: "Verification Error: Runtime error: Execution failed: Error calling api function: Failed to convert parameter `tx` from node to runtime of validate_transaction", data: Some(RawValue("RuntimeApi(\"Execution failed: Error calling api function: Failed to convert parameter `tx` from node to runtime of validate_transaction\")")) }
```

whereas skipping the option succeeds:
```shell
> cargo contract call --url "$NODE" --suri "$BOB" \
    --skip-confirm \
    --contract "$CONTRACT_ADDRESS" \
    --gas "$GAS_LIMIT" \
    --message "add"

<success logs>
```

_Note: `$BOB` has enough funds for both gas and storage deposit limits._

# Reproduction

## Version
```shell
> cargo contract --version
cargo-contract 1.5.0-unknown-x86_64-unknown-linux-gnu
```

## Contract
```rust
#[ink::contract]
mod contract {
    use ink_storage::{traits::SpreadAllocate, Mapping};

    #[ink(storage)]
    #[derive(SpreadAllocate)]
    pub struct Contract {
        pub map: Mapping<AccountId, ()>,
    }

[...]

PR #751, opened 2022-09-20
Fix storage deposit limit encoding

ascjones:
A bug in the `subxt` codegen meant that compact types as type parameters were not generated correctly. `storage_deposit_limit` is an `Option<Compact<u128>>` but was generated as a `Option<u128>` which resulted in incorrect encoding. https://github.com/paritytech/subxt/pull/651 fixes that issue and this PR uses the new `subxt` release and fixes the issue here.

Fixes https://github.com/paritytech/cargo-contract/issues/734

### DIFF
--- a/crates/cargo-contract/src/cmd/extrinsics/events.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/events.rs
@@ -111,11 +111,12 @@ impl DisplayEvents {
                     let field = Field::new(String::from("data"), contract_event);
                     event_entry.fields.push(field);
                 } else {
-                    let field_name = field.name().map(ToOwned::to_owned).unwrap_or_else(|| {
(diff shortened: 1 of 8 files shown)


crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -100,26 +100,27 @@ impl DisplayEvents {
 
             let event_data = &mut event.field_bytes();
             let mut unnamed_field_name = 0;
-            for (field, field_ty) in event_fields {
+            for field in event_fields {
```
ascjones, line 103: Note that changes here are only to fix compilation because of the updating to `subxt` `0.24`, and unrelated to the bugfix itself.

HCastano (approved):

PR #737, opened 2022-09-12
Skip linting if running building in --offline mode

ascjones:
Fixes #732.

Skips the linting step if building with `--offline`. This is because `cargo dylint` requires network access to build the linting drivers.

PR #753, opened 2022-09-21
Release `v2.0.0-alpha.3`

ascjones:
## [2.0.0-alpha.3] 

This release supports compatibility with the [`v4.0.0-alpha.3`](https://github.com/paritytech/ink/releases/tag/v4.0.0-alpha.3)
release of `ink!`. It is *not* backwards compatible with older versions of `ink!`.

### Added
- `--output-json` support for `call`, `instantiate` and `upload` commands - [#722](https://github.com/paritytech/cargo-contract/pull/722)

### Fixed
- Skip linting if running building in --offline mode -  [#737](https://github.com/paritytech/cargo-contract/pull/737)

### DIFF
--- a/crates/cargo-contract/src/workspace/manifest.rs
+++ b/crates/cargo-contract/src/workspace/manifest.rs
@@ -289,7 +289,7 @@ impl Manifest {
         let ink_dylint = {
             let mut map = value::Table::new();
             map.insert("git".into(), "https://github.com/paritytech/ink/".into());
-            map.insert("branch".into(), "master".into());
+            map.insert("tag".into(), "v4.0.0-alpha.3".into());
             map.insert("pattern".into(), "linting/".into());
             value::Value::Table(map)
         };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [2.0.0-alpha.3] - 2022-09-21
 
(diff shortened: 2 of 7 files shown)


HCastano (commented):
We need to also update them template for the new ink! release

crates/cargo-contract/src/workspace/manifest.rs
```diff
@@ -289,7 +289,7 @@ impl Manifest {
         let ink_dylint = {
             let mut map = value::Table::new();
             map.insert("git".into(), "https://github.com/paritytech/ink/".into());
-            map.insert("branch".into(), "master".into());
+            map.insert("tag".into(), "v4.0.0-alpha.3".into());
```
athei, line 292: What this needs to do is to read out the version of the used ink! version. Otherwise linting a v3 ink contract will also disable linting.

ascjones: Ok, then I will need to do several more things to make this backwards compatible for v3 `ink` contracts. Alternatively we could say that you have to use `cargo-contract` `1.x` for `v3` contracts

athei: Yeah I think this is fine. We are still changing too much to be backwards compatible.

ascjones: https://github.com/paritytech/cargo-contract/issues/754

athei (approved):

HCastano (approved):

ascjones:
https://crates.io/crates/cargo-contract/2.0.0-alpha.3

Issue #733, opened 2022-09-08
Integrating wasm-opt via wasm-opt-rs


brson:
I have been working on [Rust bindings to wasm-opt](https://github.com/brson/wasm-opt-rs), with a goal of integrating those bindings directly into cargo-contract, to ease the developer workflow. This work is part of [a w3f grant](https://github.com/w3f/Grants-Program/blob/master/applications/wasm-opt-for-rust.md).

Today cargo-contract developers must source wasm-opt from somewhere outside the rust ecosystem, typically from a system package manager, or binaries on the binaryen github.

With the above-linked wasm-opt-rs, we have two options that should provide a better experience to Rust developers:

1) suggest `cargo install wasm-opt --locked`
2) call wasm-opt programmatically from within cargo-contract so the developer doesn't need to care about it

I am opening this issue to discuss this prospect.

As a strawman proposal, so you can see the impact this would have on cargo-contract, I have prepared a branch of cargo-contract that implements both these strategies in two separate commits:

https://github.com/brson/cargo-contract/commits/wasm-opt-rs

Things to note about this patch

- it is intended to be easily revertable in multiple ways: the patch itself can be reverted, the cargo feature turned off, and the wasm API disabled dynamically; in case bugs are encountered and things need to be backed out
- it interprets the `std::process::Command` that cargo-contract is already constructing in order to run wasm-opt programmatically, so that only one code path is needed to set up either the CLI or the API. It is expected that some time in the future the CLI code path would be removed, and at that time cargo-contract can switch from the Command API to a more idiomatic builder API, also available in the crate.
- wasm-opt-rs depends on Rust 1.48+ and a c++17 compiler. It does not depend on cmake or make.

Please give this patch a look and let me know your thoughts about the proposed integration. I am happy to make all accommodations.

The API documentation for wasm-opt-rs is at https://docs.rs/wasm-opt/0.0.1-preview.3/wasm_opt/index.html, though this link is a bit out of date, and doesn't show the `Command`-based API used in the linked patch.

wasm-opt-rs is not ready for production today, but I expect it to be within 4-6 weeks, pending a [binaryen bugfix](https://github.com/WebAssembly/binaryen/issues/4938) and more integration testing.

I expect to receive a further maintenance grant from w3f to do light maintenance of wasm-opt-rs, updating both it and cargo-contract each time a new version of binaryen is released, so this crate should continue to be supported to a sufficient level for cargo-contract's purposes.

cc @cmichi @mmagician

[...]

PR #766, opened 2022-10-05
Use integrated wasm-opt

brson:
Compile `wasm-opt` and call it as a library instead of as an external binary.

This PR has three patches that result in completely removing support for the `wasm-opt` CLI in favor of the crate. If desired, the third commit can be excluded, leaving support for both methods.

Documentation for the `wasm-opt` bindings:

- https://github.com/brson/wasm-opt-rs
- https://docs.rs/wasm-opt/latest/wasm_opt/

Things to be aware of about the `wasm-opt` crate:

- It depends on Rust 1.48+
- It depends on a c++17 compiler, and should emit a readable error if one is not available.
- The `wasm-opt-sys` crate takes a non-negligible amount of time to build (minutes on my underpowered laptop). It also does not do any incremental recompilation, so if the build is invalidated it will rebuild the C++ code from scratch. The lack of incremental recompilation is a limitation self-imposed by not using cmake or other external build system.
- `wasm-opt` on windows does not support extended unicode paths (probably anything non-ascii, maybe latin-1, it's unclear). This is a limitation of binaryen and not a regression of the bindings. It may or may not be fixed in the future. The API will return an error if this occurs.
- `cargo tarpaulin` (code coverage) segfaults running any `wasm-opt` crates (https://github.com/brson/wasm-opt-rs/issues/59), reason unknown. This behavior could infect other crates that link to `wasm-opt`. If you use tarpaulin, you might check it.

Fixes https://github.com/paritytech/cargo-contract/issues/733

### DIFF
--- a/crates/cargo-contract/src/main.rs
+++ b/crates/cargo-contract/src/main.rs
@@ -50,6 +50,7 @@ use std::{
     str::FromStr,
 };
 
+use ::wasm_opt::OptimizationOptions;
 use anyhow::{
     anyhow,
     Error,
@@ -70,6 +71,12 @@ use assert_cmd as _;
 
 #[cfg(test)]
 use predicates as _;
+
+#[cfg(test)]
+use regex as _;
+
+// Only used on windows.
+use which as _;
 
 #[derive(Debug, Parser)]
 #[clap(bin_name = "cargo")]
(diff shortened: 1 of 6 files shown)


athei (approved):

ascjones (approved):
This is absolutely fantastic thank you @brson, it removes some major friction from our installation process!

> It depends on a c++17 compiler, and should emit a readable error if one is not available.

I wonder whether we should include that as a prerequisite. Which reminds me, we should remove `Step 2: install binaryen` from https://github.com/paritytech/cargo-contract#installation

athei:
We should update the README in here, too? Need to remove `wasm-opt` from the list of required dependencies.

brson:
I have updated the readme to remove the binaryen installation step, and added an entry to the changelog.

I am not sure where in the readme to mention the c++17 prerequisite.

brson:
I've also added a mention of c++17 to the installation section.

crates/cargo-contract/src/main.rs
```diff
@@ -71,6 +72,12 @@ use assert_cmd as _;
 #[cfg(test)]
 use predicates as _;
 
+#[cfg(test)]
+use regex as _;
+
+// Only used on windows.
+use which as _;
```
cmichi, line 79: Is it not possible to prefix with `#[cfg(target_os = "windows")]` here?

cmichi:
This is a great achievement, that hopefully will also benefit the broader Rust+Wasm ecosystem. 

Thanks a lot, @brson!

crates/cargo-contract/src/main.rs
brson: I don't think so. This is intentionally doing the opposite - making it look like `which` is used on not-windows. `cargo-contract` has a clippy lint turned on to check that linked crates are actually used, and `which` is only used on windows, so clippy doesn't see it on unix. What could be done is to change the `Cargo.toml` to only link to `which` on windows.

PR #769, opened 2022-10-06
Log failure instead of failing when decoding an event

### LINKED ISSUE #738: cargo-contract exits on event decoding failure
# Problem statement 

As for now, after submitting a transaction to a chain (either call, upload or instantiate) `cargo contract` listens for events and decodes them. In particular, every event that comes from _any_ contract is decoded using the metadata file corresponding to the 'entry' contract. In case of any failure, the error is propagated up to the user, which cause the whole program to exit and fail.

Such a situation is easily reproducible with any pair of contracts `A` and `B`, such that:
 - `A` calls `B`
 - `B` emits an event `E`
 - `E` is not compatible with `A`'s metadata (it is sufficient that `A` doesn't declare any events)

This happens _often_ in any non-trivial dApp and thus makes using `cargo-contract` impossible to deploy and interact with more interesting examples.

# Relevant code

https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/extrinsics/events.rs#L110

# Suggested solution

In case of decoding failure, we should just log warning / info that the event couldn't have been properly decoded with provided metadata and just continue.

pmikolajczyk41:
Fixes #738

### DIFF
--- a/crates/cargo-contract/src/cmd/extrinsics/events.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/events.rs
@@ -107,9 +107,14 @@ impl DisplayEvents {
                 ) && field.name() == Some("data")
                 {
                     tracing::debug!("event data: {:?}", hex::encode(&event_data));
-                    let contract_event = transcoder.decode_contract_event(event_data)?;
-                    let field = Field::new(String::from("data"), contract_event);
-                    event_entry.fields.push(field);
+                    if let Result::Ok(contract_event) =
+                        transcoder.decode_contract_event(event_data)
+                    {
+                        let field = Field::new(String::from("data"), contract_event);
+                        event_entry.fields.push(field);
+                    } else {
+                        tracing::warn!("Decoding contract event failed. It might have come from another contract.");
+                    }
                 } else {
                     let field_name =
                         field.name().map(ToOwned::to_owned).unwrap_or_else(|| {
--- a/crates/cargo-contract/src/cmd/extrinsics/events.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/events.rs
@@ -107,13 +107,17 @@ impl DisplayEvents {
                 ) && field.name() == Some("data")
                 {
                     tracing::debug!("event data: {:?}", hex::encode(&event_data));
-                    if let Result::Ok(contract_event) =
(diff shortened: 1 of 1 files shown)


crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -107,9 +107,14 @@ impl DisplayEvents {
...
-                    let contract_event = transcoder.decode_contract_event(event_data)?;
-                    let field = Field::new(String::from("data"), contract_event);
-                    event_entry.fields.push(field);
+                    if let Result::Ok(contract_event) =
+                        transcoder.decode_contract_event(event_data)
+                    {
+                        let field = Field::new(String::from("data"), contract_event);
+                        event_entry.fields.push(field);
+                    } else {
+                        tracing::warn!("Decoding contract event failed. It might have come from another contract.");
```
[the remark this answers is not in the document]
pmikolajczyk41: sure, done

crates/cargo-contract/src/cmd/extrinsics/events.rs
```diff
@@ -107,9 +107,18 @@ impl DisplayEvents {
                 ) && field.name() == Some("data")
                 {
                     tracing::debug!("event data: {:?}", hex::encode(&event_data));
-                    let contract_event = transcoder.decode_contract_event(event_data)?;
-                    let field = Field::new(String::from("data"), contract_event);
-                    event_entry.fields.push(field);
+                    match transcoder.decode_contract_event(event_data) {
+                        Result::Ok(contract_event) => {
```
ascjones, line 111:
```suggestion
                        Ok(contract_event) => {
```

ascjones (approved):
LGTM, thanks!

crates/cargo-contract/src/cmd/extrinsics/events.rs
pmikolajczyk41: unfortunately, we already import method `Ok()` from `anyhow` and there would be a collision

ascjones: That looks like a mistake, please remove that import.

pmikolajczyk41: Ok :)

PR #786, opened 2022-10-16
Update wasm-opt to 0.110.1

brson:
This minor release reduces the amount of binaryen source code packaged into the wasm-opt crates from 72 MB to 10 MB.


athei (approved):

cmichi (approved):
Thank you!

PR #787, opened 2022-10-17
Extract lib for invoking contract build

ascjones:
Introduces a new crate `contract-build` which allows building of `ink!` projects via a library interface.

3rd party Rust tools can use this crate instead of invoking the `cargo-contract` binary. This has the advantage of being more robust since:

1. It does not require the user to install `cargo-contract` on their machine
2. It allows the library user to control the version of `contract-build` functionality, avoiding compatibility issues.

### Note to reviewers
image
Almost all changes here are simply moving code around, so I won't change any bad code that I have copied from one location to another. I have noted a few things that can be improved/refactored (as ever), but I wanted to do the minimal changes required to extract the library for this PR.

### Future steps
Extract extrinsics functionality into a library


crates/build/README.md
```diff
@@ -0,0 +1,23 @@
+# contract-build
+
+A crate for building [`ink!`](https://github.com/paritytech/ink) smart contracts. Used by 
+[`cargo-contract`](https://github.com/paritytech/cargo-contract).
+
+## Usage
```
HCastano, line 6: We should try make this a [proper README compile test](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#include-items-only-when-collecting-doctests). 

Not sure if we're able to pull files in for the `manifest_path` though

crates/cargo-contract/Cargo.toml
```diff
@@ -14,34 +14,25 @@ description = "Setup and deployment tool for developing Wasm based smart contrac
...
 categories = ["command-line-utilities", "development-tools::build-utils", "development-tools::cargo-plugins"]
 include = [
-    "Cargo.toml", "src/**/*.rs", "README.md", "LICENSE", "build.rs", "templates", "src/**/*.scale",
+    "Cargo.toml", "src/**/*.rs", "README.md", "LICENSE", "build.rs", "src/**/*.scale",
 ]
 
 [dependencies]
+contract-build = { version = "2.0.0-beta", path = "../build" }
+contract-metadata = { version = "2.0.0-beta", path = "../metadata" }
+transcode = { package = "contract-transcode", version = "2.0.0-beta", path = "../transcode" }
```
HCastano, line 23: For consistency

```suggestion
contract-transcode = { version = "2.0.0-beta", path = "../transcode" }
```

HCastano (approved):
Good stuff!

crates/build/README.md
ascjones: Great idea. Done!

I guess it doesn't matter that the manifest path does not exist, `execute` will just return an `Err` so the test will pass.

I would add `no_run` but not sure how to add multiple code block attributes, and `rust` is already there.

ascjones:
Released https://crates.io/crates/contract-build/2.0.0-beta

PR #799, opened 2022-10-26
Make linting opt in with `--lint`

ascjones:
Since the release of `ink 4.0`, lints are optional and currently there are no lints which are run from the `ink!` repo.

This PR replaces `--skip-linting` with the inverse `--lint` to make linting opt-in by default instead of opt out. That way if/when we do add lints to `ink` there will be no changes required here.

### DIFF
--- a/crates/cargo-contract/src/cmd/build/mod.rs
+++ b/crates/cargo-contract/src/cmd/build/mod.rs
@@ -423,7 +423,7 @@ fn check_dylint_requirements(_working_dir: Option<&Path>) -> Result<()> {
     }
 
     // On windows we cannot just run the linker with --version as there is no command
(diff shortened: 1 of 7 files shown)


crates/cargo-contract/src/cmd/build/mod.rs
```diff
@@ -250,7 +252,7 @@ impl CheckCommand {
             unstable_flags,
             optimization_passes: OptimizationPasses::Zero,
             keep_debug_symbols: false,
-            skip_linting: false,
+            lint: false,
```
ascjones, line 255: Check command now defaults to **not** running lints.

cmichi (approved):
:+1:

PR #802, opened 2022-10-27
Upgrade wasm-opt to 0.110.2

brson:
The only change here is that it contains a [backport of a Binaryen patch](https://github.com/brson/wasm-opt-rs/issues/105) that slightly reduces output size for modules that do not contain a 'memories' section.


ascjones (approved):
Thanks!

PR #804, opened 2022-11-01
Log code hash if contract is already uploaded.

deuszx:
This PR adds printing of contract's code hash if `cargo contract upload` is called with a contract that has already been uploaded.

### DIFF
--- a/crates/cargo-contract/src/cmd/extrinsics/upload.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/upload.rs
@@ -73,7 +73,7 @@ impl UploadCommand {
         )?;
         let contract_metadata =
             contract_metadata::ContractMetadata::load(&crate_metadata.metadata_path())?;
-        let transcoder = ContractMessageTranscoder::from_metadata(contract_metadata)
+        let transcoder = ContractMessageTranscoder::try_from(contract_metadata)
             .context(format!(
                 "Failed to deserialize ink project metadata from contract metadata {}",
                 crate_metadata.metadata_path().display()
--- a/crates/cargo-contract/src/cmd/extrinsics/upload.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/upload.rs
@@ -73,6 +73,7 @@ impl UploadCommand {
         )?;
         let contract_metadata =
             contract_metadata::ContractMetadata::load(&crate_metadata.metadata_path())?;
+        let code_hash = contract_metadata.source.hash;
         let transcoder = ContractMessageTranscoder::try_from(contract_metadata)
             .context(format!(
                 "Failed to deserialize ink project metadata from contract metadata {}",
@@ -82,7 +83,7 @@ impl UploadCommand {
 
         let wasm_path = match &self.wasm_path {
             Some(wasm_path) => wasm_path.clone(),
-            None => crate_metadata.dest_wasm.clone(),
+            None => crate_metadata.dest_wasm,
         };
 
         tracing::debug!("Contract code path: {}", wasm_path.display());
@@ -132,9 +133,6 @@ impl UploadCommand {
                 }
                 Ok(())
             } else {
-                let code_hash = contract_metadata::ContractMetadata::load(
-                    &crate_metadata.metadata_path(),
-                )?.source.hash;
                 Err(format!(
                     "This contract has already been uploaded with code hash: {:?}",
                     code_hash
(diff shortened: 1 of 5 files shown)


crates/transcode/src/lib.rs
```diff
@@ -184,6 +178,13 @@ impl ContractMessageTranscoder {
         Ok(Self::new(ink_metadata))
     }
 
+    /// Constructs an instance of `Self` out of a contract metadata.
+    ///
+    /// Returns error if can't deserialize metadata's ABI.
+    pub fn from_metadata(metadata: contract_metadata::ContractMetadata) -> Result<Self> {
```
ascjones, line 184: This could be `impl TryFrom<ContractMetadata>`

crates/cargo-contract/src/cmd/extrinsics/upload.rs
```diff
@@ -126,7 +132,13 @@ impl UploadCommand {
                 }
                 Ok(())
             } else {
-                Err("This contract has already been uploaded".into())
+                let code_hash = contract_metadata::ContractMetadata::load(
+                    &crate_metadata.metadata_path(),
+                )?.source.hash;
```
ascjones, line 137: An simpler alternative to loading the metadata just to get the source hash would be to just hash the contract code inline here like https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/metadata.rs#L233.

Then would just be `blake2_hash(&code)`.

goral09: Thanks for the suggestion. `code` is already moved though in line 122. I can of course clone it and do as you suggest, if that's what you prefer. I'm not a fan of the approach since it would duplicate the logic that takes the contract and produces the code hash and *IF* it ever changes then this would have to be updated as well. I understand that reading from the file system is probably slower but the slowness here (in the one-shot command) seems justifiable if in return we get cleaner code and more future-proof.

deuszx: Thanks for the suggestion. `code` is already moved though in line 122. I can of course clone it and do as you suggest, if that's what you prefer. I'm not a fan of the approach since it would duplicate the logic that takes the contract and produces the code hash and *IF* it ever changes then this would have to be updated as well. I understand that reading from the file system is probably slower but the slowness here (in the one-shot command) seems justifiable if in return we get cleaner code and more future-proof.

I could just do a copy of `code_hash`, right after loading `contract_metadata` in line 76, and use it here instead.

crates/transcode/src/lib.rs
deuszx: Done in [`e0d3e79` (#804)](https://github.com/paritytech/cargo-contract/pull/804/commits/e0d3e795be6f2b6fedc6ee173b263db2f35b8127)

crates/cargo-contract/src/cmd/extrinsics/upload.rs
deuszx: I went with just making a copy of `code_hash` and reusing that instead. Done in [`5731420` (#804)](https://github.com/paritytech/cargo-contract/pull/804/commits/5731420b651153f1b074c63b35edf4c9feb038b5).

crates/cargo-contract/src/cmd/extrinsics/upload.rs
```diff
@@ -126,7 +133,10 @@ impl UploadCommand {
                 }
                 Ok(())
             } else {
-                Err("This contract has already been uploaded".into())
+                Err(format!(
+                    "This contract has already been uploaded with code hash: {:?}",
+                    code_hash
+                ).as_str().into())
```
ascjones, line 139:
```suggestion
                Err(anyhow::anyhow!("This contract has already been uploaded with code hash: {:?}", code_hash))
```

ascjones:
> User @deuszx, please sign the CLA [here](https://cla.parity.io/sign_cla/95355183/3).

Did you also sign the CLA?

deuszx:
> > User @deuszx, please sign the CLA [here](https://cla.parity.io/sign_cla/95355183/3).
> 
> Did you also sign the CLA?

I already did! Now, when I click on "Sign with GitHub" I'm getting:
> **ERROR: Unexpected status code 404 ()**

ascjones (approved):
LGTM but needs a `cargo +nightly fmt --all`

deuszx:
I will try closing and reopening the PR. Maybe the CLA bot will get fixed.

deuszx:
reopened in https://github.com/paritytech/cargo-contract/pull/805

PR #805, opened 2022-11-02
Log code hash if contract is already uploaded.

deuszx:
This PR adds printing of contract's code hash if cargo contract upload is called with a contract that has already been uploaded.


deuszx:
OK, @ascjones , it seems like now it picked it up me signing the CLA previously and is giving me a ✅ .

ascjones (approved):

PR #816, opened 2022-11-10
Update validation for renamed FFI functions

ascjones:
`seal` prefixes have been removed, need to check for the module name now.

See https://github.com/paritytech/ink/pull/1482#issuecomment-1309424077


athei (approved):

athei:
Will the CI use it automatically if we do a release?

ascjones:
It will use `master` https://github.com/paritytech/scripts/blob/8d3902c5356f6efed2fc116b234e83b58a69f656/dockerfiles/contracts-ci-linux/Dockerfile#L70

PR #826, opened 2022-11-17
Draft to manage in a better way nightly error 

lean-apple:


lean-apple:
At the moment, the addition of Channel::Nightly helps to manage the nightly case, yet I should probably add a better error handling depending of the `util::invoke_cargo(…)` output also.

ascjones:
Fixed in https://github.com/paritytech/cargo-contract/pull/848

Issue #671, opened 2022-08-04
Add `remove_code` extrinsic command


ascjones:
Extracted from #432.

Add a `remove-code` command, e.g. `cargo contract remove-code 0x12345...`

See the corresponding pallet-contracts method here: https://github.com/paritytech/substrate/blob/720093428b90363cb909162ef7c05e5bd51ae626/frame/contracts/src/lib.rs#L597.

Use e.g. `upload` as a starting point.

ascjones:
Closed by #837

PR #837, opened 2022-11-22
Add  `remove` command 

lean-apple:
In the wake of this issue, https://github.com/paritytech/cargo-contract/issues/671

## ToDo : 
 - [x] Improve how to get code_hash
 - [x] Add Tests with arg and with the manifest path only
 - [x] Update documentation


crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,242 @@
...
+        client: &Client,
+        signer: &PairSigner,
+    ) -> Result<CodeUploadResult<CodeHash, Balance>> {
+        let url = self.extrinsic_opts.url_to_string();
+        let token_metadata = TokenMetadata::query(client).await?;
+        let call_request = CodeRemoveRequest {
+            origin: signer.account_id().clone(),
+            code,
+        };
+        state_call(&url, "ContractsApi_remove_code", call_request).await
```
ascjones, line 164: I guess in order for this dry run to work we'd have to add `remove_code` to the runtime api: https://github.com/paritytech/substrate/blob/f465fee723c87b7348839b5140135b17d0a48743/frame/contracts/src/lib.rs#L1215

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,242 @@
...
+        let signer = super::pair_signer(self.extrinsic_opts.signer()?);
+
+        fn load_code(wasm_path: &Path) -> Result<Code> {
+            tracing::debug!("Contract code path: {}", wasm_path.display());
+            let code = fs::read(wasm_path)
+                .context(format!("Failed to read from {}", wasm_path.display()))?;
+            Ok(Code::Upload(code))
+        }
+
+        let code = match (self.wasm_path.as_ref(), self.code_hash.as_ref()) {
```
ascjones, line 87: I think we get rid of `wasm-path` and use (in order or precedence): `self.code_hash`, otherwise `contract_metadata.source.hash`

ascjones (commented):
Can see it's still a WIP but looks reasonable so far. The main thing to check is about that dry-run endpoint on `pallet-contracts`. We may want to consider that it is not necessary in this case to do the dry-run since we don't need the gas limit estimate like we do with other commands. The only thing to think about is how to elegantly handle the case where the code_has to be removed does not exist on chain.

ascjones:
Once https://github.com/paritytech/cargo-contract/pull/893 is merged we could use that too for determining the code hash

crates/cargo-contract/src/cmd/extrinsics/remove.rs
ascjones: See #893

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        async_std::task::block_on(async {
+            let url = self.extrinsic_opts.url_to_string();
+            let client = OnlineClient::from_url(url.clone()).await?;
+
+            if let Some(code_removed) = self
+                .remove_code(&client, sp_core::H256(code_hash), &signer, &transcoder)
+                .await?
+            {
+                let remove_result = RemoveResult {
```
ascjones, line 75: The `RemoveResult` type seems unnecessary here, could just operator on the `code_hash` inline.

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,133 @@
...
+        let signer = super::pair_signer(self.extrinsic_opts.signer()?);
+
+        let artifacts_path = artifacts.artifact_path().to_path_buf();
+        let code = artifacts.code.ok_or_else(|| {
+            anyhow::anyhow!(
+                "Contract code not found from artifact file {}",
+                artifacts_path.display()
+            )
+        })?;
+        let code_hash = code.code_hash();
```
ascjones, line 65: Thinking about this, I would consider adding back the `--code-hash` argument for this command. It's because this is the only extrinsic command which does not require any local contract artifacts at all. It can do it's job with the only the code hash. 

Of course it should still be able to use the local artifacts to infer the code hash, but this would be overridden if `--code-hash` is specified.

docs/extrinsics.md
```diff
@@ -97,6 +97,16 @@ cargo contract call \
...
 
+### `remove`
+
+Remove the Wasm code of the contract to the target chain. Invokes the [`remove_code`](https://github.com/paritytech/substrate/blob/master/frame/contracts/src/lib.rs#L581)
+dispatchable.
+
+e.g. `cargo contract remove --suri //Alice`
+
+Assumes that `cargo contract build` and `cargo contract upload` have already been run to produce the contract artifacts.
+Yet be sure your contract has not been instantiated ie it is not used by any contract.
```
ascjones, line 108: Needs more detail e.g.
```suggestion
This command will only succeed if there are no contract instances of this code. Contracts which have already been instantiated from this code must either `terminate` themselves or have their code changed via a `set_code` call to `pallet_contracts`.
```

ascjones (commented):
Looks good, I think we need the optional `--code-hash` arg back though.

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,138 @@
...
+        let signer = super::pair_signer(self.extrinsic_opts.signer()?);
+
+        let artifacts_path = artifacts.artifact_path().to_path_buf();
+
+        let artifacts_code = artifacts.code.ok_or_else(|| {
+            anyhow::anyhow!(
+                "Contract code not found from artifact file {}",
+                artifacts_path.display()
+            )
+        })?;
```
ascjones, line 71: Looks like it will fail here with no artifacts. But it shouldn't if the code hash is supplied.

One way to get around this would be to only attempt to load the artifacts in case `self.code_hash` is `None`

crates/transcode/src/decode.rs
```diff
@@ -180,8 +180,7 @@ impl<'a> Decoder<'a> {
         let discriminant = input.read_byte()?;
         let variant = variant_type
             .variants()
-            .iter()
-            .find(|v| v.index == discriminant)
+            .get(discriminant as usize)
```
ascjones, line 183: Why have you reverted this change?

lean-apple: Probably not done on purpose

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,131 @@
...
+                .await?
+            {
+                let remove_result = code_removed.code_hash;
+
+                if self.output_json {
+                    println!("{}", remove_result);
+                } else {
+                    name_value_println!("Code hash", format!("{:?}", remove_result));
+                }
+                Ok(())
```
ascjones, line 96:
```suggestion
                Result::<(), ErrorVariant>::Ok(())
```
This will fix one of your errors. It's not particularly nice, but it gives you a clue that you need to give a type hint for it to infer the error type of the `Result` here. So see if you can find a more elegant way to do this.

lean-apple: Thanks

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,147 @@
...
+        })
+    }
+
+    async fn remove_code(
+        &self,
+        client: &Client,
+        code_hash: CodeHash,
+        signer: &PairSigner,
+        transcoder: &ContractMessageTranscoder,
+    ) -> Result<Option<api::contracts::events::CodeRemoved>, anyhow::Error> {
```
SkymanOne, line 127: `api::contracts` can be imported globally in a file since it's used more than once

README.md
```diff
@@ -128,6 +128,10 @@ This can be either an event, an invocation of a contract message, or an invocati
 
 The argument has to be given as hex-encoding, starting with `0x`.
 
+##### `cargo contract remove`
+
+Remove a contract to a `pallet-contracts` enabled chain. See [extrinsics](docs/extrinsics.md).
```
SkymanOne, line 133:
```suggestion
Remove a contract from a `pallet-contracts` enabled chain. See [extrinsics](docs/extrinsics.md).
```

SkymanOne (approved):
LGTM! Just a few nitpicks :)
Also, should we have a confirmation just in case?

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,147 @@
...
+                Ok(interm_artifacts_code)
+            }
+            (None, None) => {
+                Err(
+                    anyhow::anyhow!(
+                    "No code_hash was provided or contract code was not found from artifact file {}. Please provide a code hash with --code-hash argument or specify the path for artifacts files with --manifest-path", 
+                    artifacts_path.display()
+                ))
+            }
+        };
```
ascjones, line 89:
```suggestion
        let final_code_hash = match (self.code_hash.as_ref(), artifacts.code.as_ref()) {
            (Some(code_h), _) => {
                Ok(code_h.0)
            }
            (None, Some(_)) => {
                artifacts.code_hash()
            }
            (None, None) => {
                Err(
                    anyhow::anyhow!(
                    "No code_hash was provided or contract code was not found from artifact file {}. Please provide a code hash with --code-hash argument or specify the path for artifacts files with --manifest-path", 
                    artifacts_path.display()
                ))
            }
        }?;
```
Then remove the `? from `final_code_hash` below

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        async_std::task::block_on(async {
+            let url = self.extrinsic_opts.url_to_string();
+            let client = OnlineClient::from_url(url.clone()).await?;
+            if let Some(code_removed) = self
+                .remove_code(
+                    &client,
+                    sp_core::H256(final_code_hash?),
```
ascjones, line 97:
```suggestion
                    sp_core::H256(final_code_hash),
```
here

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,147 @@
...
+
+                if self.output_json {
+                    println!("{}", &remove_result);
+                } else {
+                    name_value_println!("Code hash", format!("{remove_result:?}"));
+                }
+                Result::<(), ErrorVariant>::Ok(())
+            } else {
+                let art_code_hash = artifacts.code_hash();
+                let error_code_hash = hex::encode(art_code_hash?);
```
ascjones, line 113:
```suggestion
                let error_code_hash = hex::encode(final_code_hash);
```

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,147 @@
...
+                Err(anyhow::anyhow!(
+                    "This contract could not have been removed for the supplied code hash: {}", error_code_hash)
```
ascjones, line 115:
```suggestion
                    "Error removing the code for the supplied code hash: {}", error_code_hash)
```

docs/extrinsics.md
```diff
@@ -97,6 +97,23 @@ cargo contract call \
...
 - `--args` accepts a space separated list of values, encoded in order as the arguments of the message to invoke. 
 
+### `remove`
+
+Remove the Wasm code of the contract to the target chain. Invokes the [`remove_code`](https://github.com/paritytech/substrate/blob/master/frame/contracts/src/lib.rs#L581)
+dispatchable.
+
+e.g. `cargo contract remove --suri //Alice`
+
+Assumes that `cargo contract build` and `cargo contract upload` have already been run to produce the contract artifacts.
```
ascjones, line 107:
```suggestion
Assumes that `cargo contract build` and `cargo contract upload` have already been run to produce the contract artifacts.
```
`cargo contract build` is not a requirement locally now, since you can just supply the code hash. Also it is possible to `upload` code using `instantiate`.

docs/extrinsics.md
```diff
@@ -97,6 +97,23 @@ cargo contract call \
...
+This command will only succeed if there are no contract instances of this code. Contracts which have already been instantiated from this code must either `terminate` themselves or have their code changed via a `set_code` call to `pallet_contracts`.
+
+cargo contract instantiate \
```
ascjones, line 110:
```suggestion
cargo contract remove \
```

ascjones (commented):
Just a few suggestions but we are nearly there!

crates/cargo-contract/src/cmd/extrinsics/remove.rs
lean-apple: Thanks it seems far better

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,145 @@
...
+                artifacts.code_hash()
+            }
+            (None, None) => {
+                Err(
+                    anyhow::anyhow!(
+                    "No code_hash was provided or contract code was not found from artifact file {}. Please provide a code hash with --code-hash argument or specify the path for artifacts files with --manifest-path", 
+                    artifacts_path.display()
+                ))
+            }
+        };
```
ascjones, line 85:
```suggestion
        }?;
```

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,144 @@
...
+        events::DisplayEvents,
+        parse_code_hash,
+        ErrorVariant,
+    },
+    name_value_println,
+};
+use anyhow::{
+    Ok,
+    Result,
+};
```
ascjones, line 39:
```suggestion
use anyhow::Result;
```
Importing this `Ok` is why we are forced to use fully qualified `Result::<(), ErrorVariant>::Ok(())` below. Remove and then you can remove that,

crates/cargo-contract/src/cmd/extrinsics/remove.rs
```diff
@@ -0,0 +1,144 @@
...
+        })
+    }
+
+    async fn remove_code(
+        &self,
+        client: &Client,
+        code_hash: CodeHash,
+        signer: &PairSigner,
+        transcoder: &ContractMessageTranscoder,
+    ) -> Result<Option<api::contracts::events::CodeRemoved>, anyhow::Error> {
```
ascjones, line 124:
```suggestion
    ) -> Result<Option<api::contracts::events::CodeRemoved>, ErrorVariant> {
```
This will cause a much nicer error message to be displayed, similar to the other commands. 

Before: 

image

After: 

image

ascjones (commented):
Please address the final couple of comments then we can merge :+1:

crates/cargo-contract/src/cmd/extrinsics/remove.rs
lean-apple: Thanks !

crates/cargo-contract/src/cmd/extrinsics/remove.rs
ascjones: I have addressed this one: https://github.com/paritytech/cargo-contract/pull/837/commits/a857219036f9cc8aa933bdb9e49f43600787f659

ascjones (approved):
:tada:

lean-apple:
Thanks for the final touches

Issue #832, opened 2022-11-22
CodeRejected - Can't decode wasm code


wottpal:
Hey there, compiling ink! contracts as well as tests are running through smoothly without any issues. 

But then when I then try uploading the contract via contracts-ui ([weightv1 version](https://weightv1--contracts-ui.netlify.app/)) it rejects with the following error:

<img width="1304" alt="Screenshot 2022-11-22 at 12 33 09" src="https://user-images.githubusercontent.com/3930150/203303943-0aca362a-4f10-4f09-a54e-f3169e81c9f0.png">

I'm using:
* cargo-contract@1.5.0 with ink! v3
* rustc 1.67.0-nightly (b7bc90fea 2022-11-21)

And also my local substrate-contracts-node outputs:

```
2022-11-22 12:15:35.116 DEBUG tokio-runtime-worker runtime::contracts: CodeRejected: Can't decode wasm code    
```

But the same error happens when trying to deploy on a live testnet, e.g. Aleph Zero Testnet.

Maybe related to #773 and the newest rustc nightly isn't compatible anymore?

For reference, the full non-working `greeter.contract` file:

```
[...]

PR #838, opened 2022-11-22
add `-C target-cpu=mvp` rust flag to `build` command

agryaznov:
`nightly` rust [adds](https://github.com/paritytech/cargo-contract/issues/832) some new WASM features which are out of WASM MVP and hence not supported by `pallet_contracts` . 

This change should catch\avoid such cases on the `cargo contract build` step.

### DIFF
--- a/crates/cargo-contract/src/cmd/build/mod.rs
+++ b/crates/cargo-contract/src/cmd/build/mod.rs
@@ -308,7 +308,7 @@ fn exec_cargo_for_wasm_target(
         }
         let env = vec![(
             "RUSTFLAGS",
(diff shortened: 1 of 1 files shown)


agryaznov:
I think we should backport this to `1.5.x` as well

ascjones (approved):
LGTM. @athei?

athei:
> LGTM. @athei?

jup. I am curious if this actually resolves the bug. It is just a hunch of mine.

athei (approved):

wottpal:
gm. when will there be v1 & v2 tags/releases with this inside?

athei:
Should happen pretty soon. Probably today there should be a release. No promises, though. @agryaznov already started the branch. And I think we only backport this single chance for now.

PR #846, opened 2022-11-29
Remove `unstable` module check, add `--skip-wasm-validation`

ascjones:
Companion to https://github.com/paritytech/ink/pull/1522.

Imports from `__unstable__` modules now invalid. However I have added an escape hatch `--skip-wasm-validation` which skips this check in case the user wants to build an incompatible version of `ink!`.


agryaznov (approved):
LGTM

Issue #746, opened 2022-09-16
Wrong error output ("mismatching dependencies") when nightly is used


cmichi:
```bash
$ rustup show active-toolchain nightly-aarch64-apple-darwin
$ cargo contract build                              
ERROR: Mismatching versions of `parity-scale-codec` were found!
Please ensure that your contract and your ink! dependencies use a compatible version of this package.
```

This error output is incorrect, the dependencies are all compatible.

The error happens because the function `assert_compatible_ink_dependencies()` executes `util::invoke_cargo(…)`, which in turn executes `assert_channel()?;`. The `assert_channel()` function fails because the toolchain is `nightly`. The error is bubbled up and the function `assert_compatible_ink_dependencies` erroneously assumes it was because of incompatible dependencies.

The error handling has to be improved here, so that the correct output is printed ("Contracts should be built using stable").

boyswan:
I'm coming across this same error messages when _not_ using nightly:

```
❯ rustup default
stable-x86_64-apple-darwin (default)

❯ cargo contract -V
cargo-contract 2.0.0-alpha.3-unknown-x86_64-apple-darwin
```
```
[dependencies]
ink = "4.0.0-alpha.3"
```

If I downgrade cargo-contract to 2.0.0-alpha.2, then I get an `ERROR: No 'ink_lang' dependency found`.  I've also tried using `master` from the ink repo, but keep running into one of these two errors.

HCastano:
Closed via #848

PR #848, opened 2022-12-02
Remove rust toolchain channel check

ascjones:
@HCastano reported the [following](https://github.com/paritytech/ink/pull/1523#pullrequestreview-1201375457) error when using `cargo +nighlty test` for E2E tests using the new [`contract-build`](https://github.com/paritytech/ink/pull/1523) library. First of all the error is misleading and is an occurrence of https://github.com/paritytech/cargo-contract/issues/746.

The issue is that the `+nightly` gets inherited and the `assert_channel` check fails. This PR removes this check entirely and allows the underlying `cargo build` command to inherit whatever toolchain the user needs. It still does add `RUSTC_BOOTSTRAP` in case we are building with `stable`.

This should now be more in line with `cargo` philosophy and give more flexibility to contract authors with their toolchain management.

See also https://github.com/paritytech/cargo-contract/pull/698 where stable toolchain was first introduced.

### DIFF
--- a/crates/build/src/util/mod.rs
+++ b/crates/build/src/util/mod.rs
@@ -22,36 +22,11 @@ use anyhow::{
     Context,
     Result,
 };
-use rustc_version::Channel;
-use semver::Version;
 use std::{
     ffi::OsStr,
     path::Path,
     process::Command,
 };
-
-/// This makes sure we are building with a minimum `stable` toolchain version.
-fn assert_channel() -> Result<()> {
-    let meta = rustc_version::version_meta()?;
-    let min_version = Version::new(1, 63, 0);
-    match meta.channel {
-        Channel::Stable if meta.semver >= min_version => Ok(()),
-        Channel::Stable => {
-            anyhow::bail!(
-                "The minimum Rust version is {}. You are using {}.",
-                min_version,
-                meta.semver
-            )
-        }
-        _ => {
-            anyhow::bail!(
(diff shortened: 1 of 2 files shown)


crates/build/src/util/mod.rs
```diff
@@ -101,9 +75,6 @@ where
         cmd.current_dir(path);
     }
 
-    // Allow nightly features on a stable toolchain
-    cmd.env("RUSTC_BOOTSTRAP", "1");
```
ascjones, line 105: This is only required when doing `cargo build` so have moved it to `exec_cargo_for_wasm_target`.

crates/build/src/util/mod.rs
```diff
@@ -22,37 +22,12 @@ use anyhow::{
...
     path::Path,
     process::Command,
 };
 
-/// This makes sure we are building with a minimum `stable` toolchain version.
-fn assert_channel() -> Result<()> {
-    let meta = rustc_version::version_meta()?;
-    let min_version = Version::new(1, 63, 0);
-    match meta.channel {
-        Channel::Stable if meta.semver >= min_version => Ok(()),
```
ascjones, line 38: This min version check can be replaced by adding [`rust_version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) to the `ink` crate, might be a more suitable location for it?

HCastano (approved):
Cool 👍

Issue #527, opened 2022-04-24
Add `--feature` argument


athei:
Currently, there is no way to use custom features in a contract with cargo contract. We should add this CLI argument and forward it to cargo.

PR #853, opened 2022-12-06
Pass `--features` through to `cargo` 

ascjones:
Closes https://github.com/paritytech/cargo-contract/issues/527.

Note that we do not also add the `--all-features` and `--no-default-features` flags, since they would interfere with the Wasm build where we already do `--no-default-features` to prevent the `std` feature being built.


HCastano (approved):
I didn't actually expect you to add this, lol

PR #878, opened 2022-12-21
Fixed installation procedure

shunsukew:
`cargo install --force --locked cargo-contract` requires `dylint-link` being installed. Otherwise,  compile fails.

To avoid developers' confusion when starting with cargo-contract for the first time, The installation order should be reversed.

Screen Shot 2022-12-21 at 15 43 38

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,13 +39,13 @@ Modern releases of gcc and clang, as well as Visual Studio 2019+ should work.
 
 * Step 1: `rustup component add rust-src`.
 
-* Step 2: `cargo install --force --locked cargo-contract`.
-
-* Step 3: (**Optional**) Install `dylint` for linting.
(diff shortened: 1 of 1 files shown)


HCastano:
Hey @shunsukew. The installation instructions on `master` are for `cargo-contract@2.0.0-beta` releases, which make linting optional. 

When you install from `cargo` without specifying a version you'll get the latest stable release, which is `cargo-contract@1.5.1`. 

So maybe what we can do here instead is mention in the `README` that people should install the beta version by adding  `--version 2.0.0-beta.1`.

ascjones:
@HCastano is correct, these README instructions reflect the code on `master`. Note the README for the latest `1.5.1` non pre-release shows the instructions in the order that you specify: https://crates.io/crates/cargo-contract

PR #879, opened 2022-12-21
Require env_type transcoders to be Send + Sync

obrok:
They are already required to be `'static`, so this seems like a reasonable requirement. In practice, it seems only `AccountId` and `Hash` transcoders are inserted into this collection. Adding these two traits allows a `ContractMetadataTranscoder` to be shared between threads via an `Arc`.  This is very useful, especially given a `ContractMetadataTranscoder` is not even `Clone`, which makes it unwieldy to load the metadata once and use it around the program.

### DIFF
--- a/crates/transcode/src/env_types.rs
+++ b/crates/transcode/src/env_types.rs
@@ -134,12 +134,12 @@ impl From<&Path<PortableForm>> for PathKey {
 pub type TypesByPath = HashMap<PathKey, u32>;
 
 /// Implement this trait to define custom encoding for a type in a `scale-info` type registry.
-pub trait CustomTypeEncoder {
+pub trait CustomTypeEncoder: Send + Sync {
     fn encode_value(&self, value: &Value) -> Result<Vec<u8>>;
 }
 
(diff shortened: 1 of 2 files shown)


crates/transcode/src/env_types.rs
```diff
@@ -45,15 +45,15 @@ use std::{
 /// Provides custom encoding and decoding for predefined environment types.
 #[derive(Default)]
 pub struct EnvTypesTranscoder {
-    encoders: HashMap<u32, Box<dyn CustomTypeEncoder>>,
-    decoders: HashMap<u32, Box<dyn CustomTypeDecoder>>,
+    encoders: HashMap<u32, Box<dyn CustomTypeEncoder + Send + Sync>>,
```
ascjones, line 48: Would it also work if we added those constraints to the trait itself e.g. `CustomTyoeEncoder: Send + Sync`?

obrok: Yeah, that works as well, updated the PR.

ascjones (approved):

PR #888, opened 2023-01-01
Upgrade wasm-opt to 0.111.0

brson:
This upgrades binaryen to version 111.

Binaryen (and wasm-opt) now enables the `SignExt` and `MutableGlobals` features by default, which are also enabled in the LLVM backend. In the future Binaryen will align its default feature selection with the LLVM backend.

This is _probably_ is ok for cargo-contract - I assume, but don't know, that the substrate wasm vm allows these two common features.

This also points to potential future optimizations by explicitly turning on more features that the substrate vm supports, but aren't being enabled in wasm-opt.

The [changelog](https://github.com/brson/wasm-opt-rs/blob/master/CHANGELOG.md) for 0.111.0.

I see that the cargo-contract current version is a beta. If a major release is impending, it might be prudent to delay merging this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4405,9 +4405,9 @@ checksum = "1c38c045535d93ec4f0b4defec448e4291638ee608530863b1e2ba115d4fff7f"
 
 [[package]]
 name = "wasm-opt"
-version = "0.110.2"
+version = "0.111.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b68e8037b4daf711393f4be2056246d12d975651b14d581520ad5d1f19219cec"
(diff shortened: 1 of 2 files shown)


ascjones (approved):

Robbepop:
`SignExt` and `MutableGlobals` Wasm proposals are supported by the newer `wasmi` versions.

ascjones:
Apologies @brson, I was a bit trigger happy merging this and failed to read your description properly. However I did look at the release notes which mention `mvp_features_only`. So rather than just revert to `110` I have added that #891. What do you think?

brson:
@ascjones I asked the binaryen developers about this and they said that `mvp_features_only` should behave the same way wrt features as binaryen 110, and that looks to be true to me. I think it should accomplish what you want.

PR #891, opened 2023-01-05
Enable `wasm-opt` MVP features only

ascjones:
As @brson pointed out in https://github.com/paritytech/cargo-contract/pull/888, this upgrade brings new Wasm features enabled by default. @Robbepop says these are supported by `wasmi` but still need to be enabled in `pallet-contracts` (/cc @athei). 

Originally I intended just to revert to `wasm-opt` `0.110`, but noticed in the [`CHANGELOG`](https://github.com/brson/wasm-opt-rs/blob/master/CHANGELOG.md) that there is the option of `.mvp_features_only()`, which I have added as an alternative in this PR.

### DIFF
--- a/crates/build/src/wasm_opt.rs
+++ b/crates/build/src/wasm_opt.rs
@@ -66,6 +66,10 @@ impl WasmOptHandler {
         );
 
         OptimizationOptions::from(self.optimization_level)
+            // Binaryen (and wasm-opt) now enables the `SignExt` and `MutableGlobals`
+            // features by default, so we want to disable those for now since
+            // `pallet-contracts` still needs to enable these.
+            .mvp_features_only()
             // the memory in our module is imported, `wasm-opt` needs to be told that
             // the memory is initialized to zeroes, otherwise it won't run the
             // memory-packing pre-pass.


Robbepop (approved):

Robbepop:
Since LLVM enables those Wasm proposals by default, too I wonder where we deny those proposals for Rust/LLVM? 🤔

ascjones:
> Since LLVM enables those Wasm proposals by default, too I wonder where we deny those proposals for Rust/LLVM? thinking

@brson do you have any insight into this?

brson:
@ascjones @Robbepop I looked a little deeper, and here is the LLVM change referenced by the [binaryen release notes](https://github.com/WebAssembly/binaryen/blob/main/CHANGELOG.md#v111) on this subject:

https://reviews.llvm.org/D125728

It appears to actually be a change to _clang's_ default wasm configuration, not LLVM's. It is also scheduled for the LLVM 16 release, while rustc nightly is on 15. So as far as I can tell, rustc should not enable this features unless it does so explicitly.

athei:
We will only enable any proposals if we can proof that they benefit us. Because all they do is make life more complicated. Look at the multi value proposal which would make returning from functions a non constant instruction. If you overlook this fact you created yourself a security vulnerability. So there needs to be a very strong case to allow any proposals on pallet-contracts.

PR #893, opened 2023-01-06
Extrinsics: allow specifying contract artifact directly 

### LINKED ISSUE #892: Unnecessary requirements when uploading a contract
Currently, `cargo contract upload` requires `Cargo.toml` and, by extensions, all of the contract's dependencies. This means that in practice I can't simply give someone my `*.wasm` or `*.contract` files for upload.

If I understand the code correctly, it uses that information to construct a [`transcoder`](https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/extrinsics/upload.rs#L80) which is then [passed into](https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/extrinsics/upload.rs#L126) [`upload_code`](https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/extrinsics/upload.rs#L176) for [decoding events](https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/extrinsics/upload.rs#L189) upon uploading. 

Is it really required? Are there any contract-specific events that can be returned during contract upload? Or is the current state an unfortunate result of reusing code (`DisplayEvents`) that seemed similar enough?

For example, both the official wallet and ContractsUI work fine with just the `*.wasm` or `*.contract` files.

ascjones:
Extrinsic commands need to locate contract artifacts: for `upload` we require the contract Wasm and for `instantiate` and `call` the contract metadata. Currently this is designed to be similar to other `cargo` commands: providing a `--manifest-path` argument to point to the `Cargo.toml` of the contract project, if not present assuming the working directory is the project root. From there it can locate the artifacts in the `./target/ink/` directory. This is good for a development workflow from inside a project dir like:

`cargo contract build --release` 
`cargo contract upload --suri //Alice`
`cargo contract instantiate ...` 

Or from a different directory relative to the contract project directory:

`cargo contract build --manifest-path ../my-contract/Cargo.toml --release` 
`cargo contract upload --manifest-path ../my-contract/Cargo.toml --suri //Alice`
`cargo contract instantiate --manifest-path ../my-contract/Cargo.toml ...` 

However, this workflow is designed for a developer who is working with the contract source locally. 

This PR adds the ability to provide the path to a contract artifact file directly, for use cases where a user does not have the contract source project locally.

For all extrinsics, there is now an optional positional argument for the path to an artifact file. This can be one of `.wasm` (code only), `.json` (metadata only) or `.contract` (metadata + wasm). e.g.

`cargo upload ../path/to/mycontract.wasm`
`cargo instantiate ../path/to/mycontract.contract`
`cargo call ..path/to/metadata.json`

When specified, it will use that artifact file. When not specified it will fall back to the existing `--manifest-path` method.

**Note:** The existing `instantiate` command has a `--wasm-path` argument. This has been replaced by this new positional argument for consistency across the extrinsic commands.

Closes #892

### DIFF
--- a/crates/build/src/crate_metadata.rs
+++ b/crates/build/src/crate_metadata.rs
@@ -137,6 +137,13 @@ impl CrateMetadata {
     pub fn metadata_path(&self) -> PathBuf {
         self.target_directory.join(METADATA_FILE)
     }
+
+    /// Get the path of the contract bundle, containing metadata + code
+    pub fn contract_bundle_path(&self) -> PathBuf {
+        let target_directory = self.target_directory.clone();
+        let fname_bundle = format!("{}.contract", self.contract_artifact_name);
+        target_directory.join(fname_bundle)
+    }
 }
 
 /// Get the result of `cargo metadata`, together with the root package id.
--- a/crates/build/src/crate_metadata.rs
+++ b/crates/build/src/crate_metadata.rs
@@ -138,7 +138,7 @@ impl CrateMetadata {
         self.target_directory.join(METADATA_FILE)
     }
 
-    /// Get the path of the contract bundle, containing metadata + code
+    /// Get the path of the contract bundle, containing metadata + code.
     pub fn contract_bundle_path(&self) -> PathBuf {
         let target_directory = self.target_directory.clone();
         let fname_bundle = format!("{}.contract", self.contract_artifact_name);
--- a/crates/cargo-contract/src/cmd/extrinsics/mod.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/mod.rs
@@ -192,29 +192,33 @@ impl ExtrinsicOpts {
 /// Contract artifacts for use with extrinsic commands.
 #[derive(Debug)]
 pub struct ContractArtifacts {
-    pub metadata: Option<ContractMetadata>,
+    /// The expected path of the file containing the contract metadata.
+    metadata_path: PathBuf,
+    /// The deserialized contract metadata if the expected metadata file exists.
+    metadata: Option<ContractMetadata>,
+    /// The Wasm code of the contract if available.
     pub code: Option<Vec<u8>>,
 }
 
 impl ContractArtifacts {
     /// Given a contract artifact path, load the contract code and metadata where possible.
     pub fn from_artifact_path(path: &Path) -> Result<Self> {
-        let (metadata, code) =
+        let (metadata_path, metadata, code) =
             match path.extension().and_then(|ext| ext.to_str()) {
                 Some("contract") | Some("json") => {
                     let metadata = ContractMetadata::load(path)?;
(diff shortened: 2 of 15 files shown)


docs/extrinsics.md
```diff
@@ -97,6 +97,17 @@ cargo contract call \
 - `--message` the name of the contract message to invoke.
 - `--args` accepts a space separated list of values, encoded in order as the arguments of the message to invoke. 
 
+## Specifying the contract artifact
+
+The above examples assume the working directory is the contract source code where the `Cargo.toml` file is located.
+This is used to determine the location of the contract artifacts. Alternatively, there is an optional positional 
+argument to each of the extrinsic commands which allows specifying the contract artifact file directly. e.g.
```
cmichi, line 104:
```suggestion
argument to each of the extrinsic commands which allows specifying the contract artifact file directly. E.g.
```

crates/cargo-contract/src/cmd/extrinsics/upload.rs
```diff
@@ -71,27 +62,13 @@ impl UploadCommand {
...
-            Some(wasm_path) => wasm_path.clone(),
-            None => crate_metadata.dest_wasm,
-        };
-
-        tracing::debug!("Contract code path: {}", wasm_path.display());
-        let code = std::fs::read(&wasm_path)
-            .context(format!("Failed to read from {}", wasm_path.display()))?;
+        let code = artifacts
+            .code
+            .ok_or_else(|| anyhow::anyhow!("Contract code not found"))?; // todo: add more detail
```
cmichi, line 70: Forgotten ToDo comment?

crates/cargo-contract/src/cmd/extrinsics/mod.rs
```diff
@@ -163,6 +198,93 @@ impl ExtrinsicOpts {
...
+            .context("Failed to deserialize ink project metadata from contract metadata")
+    }
+}
+
+/// The Wasm code of a contract.
+#[derive(Debug)]
+pub struct WasmCode(Vec<u8>);
+
+impl WasmCode {
+    pub fn code_hash(&self) -> [u8; 32] {
```
cmichi, line 283: Comment for public function would be nice.

crates/build/src/crate_metadata.rs
```diff
@@ -137,6 +137,13 @@ impl CrateMetadata {
     pub fn metadata_path(&self) -> PathBuf {
         self.target_directory.join(METADATA_FILE)
     }
+
+    /// Get the path of the contract bundle, containing metadata + code
```
cmichi, line 141:
```suggestion
    /// Get the path of the contract bundle, containing metadata + code.
```

Just for consistency with the others.

cmichi (approved):

PR #904, opened 2023-01-16
wasm-opt with --mvp-features

kvinwang:
cargo-contract-1.5.1 still produces sign-ext instructions which aren't allowed by the contract runtime.
Let's fix it.

### DIFF
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -627,7 +627,8 @@ fn do_optimization(
         // the memory in our module is imported, `wasm-opt` needs to be told that
         // the memory is initialized to zeroes, otherwise it won't run the
         // memory-packing pre-pass.
-        .arg("--zero-filled-memory");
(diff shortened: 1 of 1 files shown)


src/cmd/build.rs
```diff
@@ -627,7 +627,8 @@ fn do_optimization(
         // the memory in our module is imported, `wasm-opt` needs to be told that
         // the memory is initialized to zeroes, otherwise it won't run the
         // memory-packing pre-pass.
-        .arg("--zero-filled-memory");
+        .arg("--zero-filled-memory")
+        .arg("--disable-sign-ext");
```
ascjones, line 631: Would `--mvp-features` have the same effect, but also disabling mutable globals? See https://github.com/paritytech/cargo-contract/pull/891

kvinwang: Good, done.

ascjones (approved):

PR #928, opened 2023-01-25
Add substrate-contracts-node run step in CI for integration tests

lean-apple:


.github/workflows/ci.yml
```diff
@@ -112,6 +112,20 @@ jobs:
           crate: dylint-link
           version: 2
 
+      - name: Download and run latest `substrate-contracts-node` binary for macOS
+        if: runner.os == 'macOS' 
+        run: |
+            curl -L -o substrate-contracts-node.zip 'https://gitlab.parity.io/parity/mirrors/substrate-contracts-node/-/jobs/artifacts/main/download?job=build-mac' && \
+            unzip substrate-contracts-node.zip && \
+            chmod +x artifacts/substrate-contracts-node-mac/substrate-contracts-node &&
+            ./artifacts/substrate-contracts-node-mac/substrate-contracts-node -linfo,runtime::contracts=debug  2>&1 | tee /tmp/contracts-node.log &
```
cmichi, line 121: For the ink! CI we use GitHub Actions (for Mac OS and Windows) and GitLab (for Linux). 

For the `cargo-contract` CI we have just GHA (for Mac OS, Windows and Linux). Hence you need to account for Linux here as well. Probably this code from the Mac step works on Linux too, if the Docker image has `unzip` installed.

lean-apple: Yes I have noticed but it was a first test to check if at least it was working for Mac and WIndows but it doesnt

cmichi: Here's the Linux command: https://github.com/paritytech/ink-examples/blob/main/.github/workflows/ci.yml#L74-L80.

.github/workflows/ci.yml
```diff
@@ -84,19 +84,37 @@ jobs:
...
+          chmod +x artifacts/substrate-contracts-node-mac/substrate-contracts-node &&
+          ./artifacts/substrate-contracts-node-mac/substrate-contracts-node -linfo,runtime::contracts=debug  2>&1 | tee /tmp/contracts-node.log &&
+          cargo test --profile debug-ci --all-features -- --test-threads=1 &    
+      
+      - name: Download - run latest `substrate-contracts-node` binary and run tests - windows
+        if: runner.os == 'Windows'
+        run: |
+          cargo install contracts-node --git https://github.com/paritytech/substrate-contracts-node.git --force --locked && \
+          substrate-contracts-node -lruntime::contracts=debug  2>&1 | tee /tmp/contracts-node.log &&
+          cargo test --profile debug-ci --all-features -- --test-threads=1 &
```
cmichi, line 110: Can you remove this line and run the CI again with `windows-2022`?

lean-apple:
@ascjones have found there is currently an incompatibility betweem `wasm-runtime`  and the `windows-sys` version - `0.45`.  
The fix to set up the `windows-sys` version will be pushed in another PR and the `windows`os part has been removed atm.

.github/workflows/ci.yml
```diff
@@ -84,12 +84,24 @@ jobs:
       - name: Cache
         uses: Swatinem/rust-cache@v2
 
+      - name: Install, run latest `substrate-contracts-node` binary and run tests
+        if: runner.os == 'Linux'
+        run: |
+          curl -L -o substrate-contracts-node.zip 'https://gitlab.parity.io/parity/mirrors/substrate-contracts-node/-/jobs/artifacts/main/download?job=build-linux' && \
+          unzip substrate-contracts-node.zip && \
+          chmod +x artifacts/substrate-contracts-node-linux/substrate-contracts-node &&
+          ./artifacts/substrate-contracts-node-linux/substrate-contracts-node -lruntime::contracts=debug  2>&1 | tee /tmp/contracts-node.log &&
```
ascjones, line 93: Note that the integration tests themselves spawn (or attempt to) `substrate-contracts-node` on the path. So they should still work if you remove this line?

lean-apple: Yep it seems to work

ascjones (approved):

PR #929, opened 2023-01-26
Use package name instead of lib name, default to "rlib"

### LINKED ISSUE #927: `cargo-contract new` should not emit `name` for the `[lib]` section
The `[lib]` section in the template:

1. Should not contain a `name` key: This should just left out and default to the package name. Keeping it sync is annoying and a major footgun when renaming the contract.

2. Crate type should contain `rlib` by default: Otherwise you can't depend on the contract

Both of those together kept me busy for a while with cryptic error messages when trying to depend on a fresh contract.

ascjones:
Closes #927.

I have a vague memory of some reasoning that the package name and lib name were separate. However I cannot remember or imagine what it might be, so LFG.


athei:
I think `rlib` is the default for libs anyways. Do we even need `crate-type` in the manifest then?

HCastano:
> I think `rlib` is the default for libs anyways. Do we even need `crate-type` in the manifest then?

From [here](https://doc.rust-lang.org/reference/linkage.html) it doesn't necessarily say that it's the default. It just says that the compiler will choose the best option

crates/build/templates/new/_Cargo.toml
```diff
@@ -11,11 +11,12 @@ scale = { package = "parity-scale-codec", version = "3", default-features = fals
 scale-info = { version = "2.3", default-features = false, features = ["derive"], optional = true }
 
 [lib]
-name = "{{name}}"
```
HCastano, line 14: Yeah this goes back to at least 2019 🤷

ascjones:
> I think `rlib` is the default for libs anyways. Do we even need `crate-type` in the manifest then?

I've tested this with the `delegator` `ink!` example, removing the `crate-type` section from the `adder` manifest, and the it compiles fine with the same resulting contract size.

athei:
> > I think `rlib` is the default for libs anyways. Do we even need `crate-type` in the manifest then?
> 
> From [here](https://doc.rust-lang.org/reference/linkage.html) it doesn't necessarily say that it's the default. It just says that the compiler will choose the best option

Yes. But the default should always allow depending on it from another crate. We actively prevented this by setting it to only `cdylib` before this PR. I just tested without any crate type: I can depend on a contract.

crates/build/src/crate_metadata.rs
```diff
@@ -66,34 +66,46 @@ impl CrateMetadata {
...
 
-        // Normalize the package and lib name.
+        // Normalize the package name.
         let package_name = root_package.name.replace('-', "_");
-        let lib_name = &root_package
+
+        if let Some(lib_name) = &root_package
             .targets
             .iter()
             .find(|target| target.kind.iter().any(|t| t == "cdylib"))
```
athei, line 75: This happens after we added it, right?

crates/build/src/lib.rs
```diff
@@ -288,6 +288,7 @@ fn exec_cargo_for_wasm_target(
         Workspace::new(&crate_metadata.cargo_meta, &crate_metadata.root_package.id)?
             .with_root_package_manifest(|manifest| {
                 manifest
+                    .with_added_crate_type("cdylib")?
                     .with_removed_crate_type("rlib")?
```
athei, line 292: I think what we want to express is that we want to set it to only `cdylib` and remove everything else.

crates/build/src/crate_metadata.rs
ascjones: Please elaborate.

We are using the package name now, the `lib_name` is just replaced by a check to show a warning in case anyone is depending on the name of the artifact being the same as the `lib` name.

athei: We search for a lib using `cdylib`.  But it is not necessarily in the source manifest. I am asking if it happens after we modified it with:
```
.with_added_crate_type("cdylib")?
```

ascjones: Ah gotcha. No this all happens before any modifications are made. It is about reading the metadata for the contract project.

The heuristic previously was assuming a `cdylib` target. Now it only keeps that in order to warn users if they are still relying on it.

athei: But they could just add a name entry there without adding `cdylib`. Then this error message would not be emitted. We basically just need to match any `[lib]` table. There can only be one per package anyways.

ascjones: True. If they did that previously though the name is ignored because of the `cdylib` predicate, and the package name will be used. The thinking was to only produce the warning in the case when the lib name would have been used for the name previously. Hence the additional check for whether `package_name != lib_name`.

athei (approved):

HCastano:
@ascjones can you add a `CHANGELOG` entry since this changes the default behaviour of contract dependencies?

PR #931, opened 2023-01-27
Display the causes of an error

ascjones:
Error information was missing the error chain, and only displaying the `context` message:

image

This PR use the `Debug` impl to display the error chain with the cause:

image

### DIFF
--- a/crates/build/src/crate_metadata.rs
+++ b/crates/build/src/crate_metadata.rs
@@ -155,8 +155,8 @@ fn get_cargo_metadata(manifest_path: &ManifestPath) -> Result<(CargoMetadata, Pa
     let mut cmd = MetadataCommand::new();
     let metadata = cmd
         .manifest_path(manifest_path.as_ref())
-        .exec()
-        .context("Error invoking `cargo metadata`")?;
+        .exec()?;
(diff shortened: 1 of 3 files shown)


athei:
Okay I don't want to shave the yak here but can you do the same for RPC error? I tried to submit an extrinsic which was to heavy for the block. The only thing I get back is:
```
ERROR: Rpc error: RPC error
```

Probably something directly coming back from subxt. There should be more information. It is the same error whether your extrinsic is invalid or you URL is wrong or missing. This is very confusing.

SkymanOne:
Any chance we can nicely format the error put and avoid?
```
... exited with error: error: ...
```

ascjones:
> Okay I don't want to shave the yak here but can you do the same for RPC error? I tried to submit an extrinsic which was tot heavy for the block. The only thing I get back is:
> 
> ```
> ERROR: Rpc error: RPC error
> ```
> 
> Probably something directly coming back from subxt. There should be more information. It is the same error whether your extrinsic is invalid or you URL is wrong or missing. This is very confusing.

Indeed this something with the `subxt::Error`. Certainly it used to display at least some unstructured information for an invalid transaction. Needs more investigation so won't do it as part of this PR. See #934

athei (approved):

PR #944, opened 2023-02-02
Remove `+nightly` from docs and READMEs

SkymanOne:
New release of `cargo-contract` and `ink!` no longer require nightly toolchain for compiling smart contracts. Therefore, we remove it from docs.


cmichi (approved):
Thx!

PR #952, opened 2023-02-04
Rename `metadata.json` to `{contract_name}.json`

### LINKED ISSUE #950: Rename `metadata.json` to `{contract_name}.json`
I don't think it makes sense to name the `contract` and `wasm` file after the contract but not the metadata file. It is just annoying to not be able to see from the file name which contract it belongs to.

SkymanOne:
Closes #950

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Rename `metadata.json` to `{contract_name}.json` - [#952](https://github.com/paritytech/cargo-contract/pull/952)
 
 ### Fixed
 - Respect the lockfile [#948](https://github.com/paritytech/cargo-contract/pull/948)
--- a/crates/build/src/crate_metadata.rs
+++ b/crates/build/src/crate_metadata.rs
@@ -36,8 +36,6 @@ use std::{
 use toml::value;
 use url::Url;
 
-const METADATA_FILE: &str = "metadata.json";
-
 /// Relevant metadata obtained from Cargo.toml.
 #[derive(Debug)]
 pub struct CrateMetadata {
@@ -147,7 +145,8 @@ impl CrateMetadata {
 
     /// Get the path of the contract metadata file
     pub fn metadata_path(&self) -> PathBuf {
-        self.target_directory.join(METADATA_FILE)
+        let metadata_file = format!("{}.json", self.contract_artifact_name);
+        self.target_directory.join(metadata_file)
     }
 
     /// Get the path of the contract bundle, containing metadata + code.
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -827,7 +827,7 @@ mod unit_tests {
         let raw_result = r#"{
   "dest_wasm": "/path/to/contract.wasm",
   "metadata_result": {
-    "dest_metadata": "/path/to/metadata.json",
+    "dest_metadata": "/path/to/contract.json",
     "dest_bundle": "/path/to/contract.contract"
   },
   "target_directory": "/path/to/target",
@@ -844,7 +844,7 @@ mod unit_tests {
         let build_result = BuildResult {
             dest_wasm: Some(PathBuf::from("/path/to/contract.wasm")),
             metadata_result: Some(MetadataResult {
-                dest_metadata: PathBuf::from("/path/to/metadata.json"),
+                dest_metadata: PathBuf::from("/path/to/contract.json"),
                 dest_bundle: PathBuf::from("/path/to/contract.contract"),
             }),
             target_directory: PathBuf::from("/path/to/target"),
--- a/crates/build/src/metadata.rs
+++ b/crates/build/src/metadata.rs
@@ -54,8 +54,6 @@ use std::{
 };
 use url::Url;
 
-pub const METADATA_FILE: &str = "metadata.json";
-
 /// Metadata generation result.
 #[derive(serde::Serialize)]
 pub struct MetadataResult {
--- a/crates/cargo-contract/src/cmd/extrinsics/mod.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/mod.rs
@@ -82,7 +82,6 @@ pub use balance::{
     TokenMetadata,
 };
 pub use call::CallCommand;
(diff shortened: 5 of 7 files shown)


cmichi (commented):
Changelog entry plz.

kvinwang:
Oh, this is a breaking change. We have tools reading the contract name from the `metadata.json`.

ascjones:
@kvinwang makes a good point. We should think carefully before introducing breaking changes.

https://github.com/paritytech/cargo-contract/issues/950#issuecomment-1418951627

ascjones:
On balance, I think this is reasonable to introduce as a breaking change into the `2.0.0` release. https://github.com/paritytech/cargo-contract/issues/950#issuecomment-1420657886

The main thing is for consistency with other artefact file names.

Downstream tools which depend on the static name will instead have to handle any `*.json` file as a contract metadata file. A breaking change, yes, but relatively easy to fix.

ascjones (approved):

athei (approved):

ascjones:
> here should be no guarantee that there is only one *.json file in the target/ink directory, and it is a metadata file.

The guarantee now is that the `ink/mycontract.wasm` artifact always exists, and unless it was built with `--build-artifacts code-only`, there will exist `mycontract.json` (metadata only) and `mycontract.contract` (metadata + code).

ascjones:
The way to determine the contract name is to take the first part of the `.wasm` code artefact. So `erc20.wasm` -> `erc20.json`. No matter if there is a `foo.json` alongside or not.

The edge case with this is that if you rename your contract, then there will be multiple `.wasm` files in the `ink` directory. In this case you could just choose the most recent one or look at `Cargo.toml` to determine the package name.

> Cargo.toml in which the contract name isn't well-defined either and it isn't script friendly and isn't always available

We have recently changed it so the contract name is always the package name https://github.com/paritytech/cargo-contract/pull/929. But anyway it should not be required in this case because of the above heuristic (all contract artifacts have the same root name, different extensions)

kvinwang:
> The way to determine the contract name is to take the first part of the `.wasm` code artefact. So `erc20.wasm` -> `erc20.json`. No matter if there is a `foo.json` alongside or not.

I don't think this is a good design. This is like asking the tools to guess `what's my name`. A `metadata.json` definitely solves the problem in the simplest and most robust way.

kvinwang:
> The tool doesn't need to guess what the name is, the name is now the name of the `json` file itself. The `target/ink` directory generated by `cargo-contract` only contains `json` files which are contract metadata, so that is a good enough guarantee.

Given there is `ink/` dir with guaranteed a `mycontract.contract`, a `mycontract.wasm` and a `mycontract.json` inside.
But without guarantee, there is not any `foo.json`, `bar.wasm` inside.
How can I know which file is the contract metadata file or contains the contract name? 
The first `.json` file found?

ascjones:
> But without guarantee, there is not any foo.json, bar.wasm inside.

I'm not exactly sure what you mean here, but if you are asking how to identify the file when it is not inside `target/ink`, then there are no guarantees anymore because `cargo-contract` is only in control of artifacts in that directory. That was also true when it was called `metadata.json`, because as well as being copied it could also be renamed. So you are in full control of the name, so you can also `cp target/ink/mycontract.json metadata.json` if you wish when copying it to your assets directory.

SkymanOne:
@kvinwang I reckon the obvious benefit of matching the metadata file with the contract's name is that sometimes you only have metadata file and meaningful name helps to understand which contract it belongs to.

We already encountered an instance of such problem: we have a workshop https://github.com/paritytech/ink-workshop where game.contract is deployed by someone else and the player had to upload metadata in polkadot.js ui to be able to read the game's contracts.

kvinwang:
> I'm not exactly sure what you mean here, but if you are asking how to identify the file when it is not inside `target/ink`, then there are no guarantees anymore because `cargo-contract` is only in control of artifacts in that directory.

I mean there is no guarantee that `cargo-contract 2.1` or any future version never generates a `foo.json` in the `ink/` dir there. If cargo-contract generates an additional `foo.json`, it would not be considered a breaking change. But it would totally break downstream build tools.

Let me show a piece of pseudocode code here.

Suppose we have a downstream build script like this:
```bash
cargo contract build --release
CONTRACT_NAME=`jq -r .contract.name target/ink/metadata.json`
```

Now, we might need to be changed to:
```bash
cargo contract build --release
CONTRACT_NAME=`cat target/ink/*.json | jq -r .contract.name`
```

Do you think the command `cat target/ink/*.json` would be never broken by any future `cargo-contract` updates? No, obviously, `cargo-contract` can not give such a weird guarantee.

I think a better solution might be adding an option to assign the metadata file name by the caller. So the script would be like:
```bash
cargo contract build --release --metadata target/ink/metadata.json
CONTRACT_NAME=`jq -r .contract.name target/ink/metadata.json`
```

kvinwang:
> @kvinwang I reckon the obvious benefit of matching the metadata file with the contract's name is that sometimes you only have metadata file and meaningful name helps to understand which contract it belongs to.
> 
> We already encountered an instance of such problem: we have a workshop https://github.com/paritytech/ink-workshop where game.contract is deployed by someone else and the player had to upload metadata in polkadot.js ui to be able to read the game's contracts.

Downstream tools can easily rename the `metadata.json` to `mycontract.json`, but not vice versa.

SkymanOne:
> > @kvinwang I reckon the obvious benefit of matching the metadata file with the contract's name is that sometimes you only have metadata file and meaningful name helps to understand which contract it belongs to.
> > We already encountered an instance of such problem: we have a workshop https://github.com/paritytech/ink-workshop where game.contract is deployed by someone else and the player had to upload metadata in polkadot.js ui to be able to read the game's contracts.
> 
> Downstream tools can easily rename the `metadata.json` to `mycontract.json`, but not vice versa.
> 
> I think a better solution might be adding an option to assign the metadata file name by the caller. Like: `cargo contract build --release --metadata target/ink/metadata.json`

Why? :)

kvinwang:
> Why? :)

I explained it in [this comment](https://github.com/paritytech/cargo-contract/pull/952#issuecomment-1421708090) above.

ascjones:
```
cargo contract build --release
CONTRACT_NAME=`jq -r .contract.name target/ink/metadata.json`
```
To make such a script more robust and agnostic of the file path of the metadata, you can read the path of the metadata from the build result e.g.:

```
cargo contract build --output-json | jq -r .metadata_result.dest_metadata
```
image

kvinwang:
> cargo contract build --output-json | jq -r .metadata_result.dest_metadata

Cool! Didn't know there is so much useful info in the JSON output.

PR #958, opened 2023-02-08
Remove `test` command

ascjones:
Users can use `cargo test` directly for a better experience.


sergejparity (approved):

agryaznov (approved):

SkymanOne (approved):

HCastano:
@ascjones 

While I generally agree that using `cargo` directly is nicer, I don't think we should've remove this so hastily. 

This leaves a gap in the user experience since we already have `check` and `build` which mirror `cargo`. It would be natural for a user to expect that `cargo contract test` also exists.

As a compromise we can have `cargo contract test` basically just shell to `cargo test` under the hood.

HCastano:
> Anyway for whatever reason it was not always working correctly, I wouldn't be averse to having it return if it did work consistently.

That's why I assumed it was doing other stuff, because there were issues with it, haha

> and we don't mirror every cargo subcommand. Although these are the most common.

Yep, but the lack of it definitely feels awkward to me, and I'm sure users will feel the same way

PR #964, opened 2023-02-09
Do not postprocess or generate metadata if contract unchanged

### LINKED ISSUE #949: cargo contract does too much work on unchanged contracts
Follow up of #926.

The following behaviour still slows down builds of unchanged contracts. This is important because of the e2e-tests which re-compile dependency contracts which are usually unchanged ( https://github.com/paritytech/ink/issues/1622 ):

- postprocessing of wasm file
- recompiling of metadata-gen v0.1.0

ascjones:
Closes #949.

Checks if the original compiled `.wasm` file in `target/ink/wasm32-unknown-unknown/release/` has changed. If it is unchanged after building, then skips the postprocessing and metadata gen iff all of the resulting contract artefacts are already there. 

Note: because it must perform the build step first before determining whether the contract has changed, the first step now shows **[1/*]**

When there are changed to the contract:

image

When there have been no changes to the contract:

image


cmichi (approved):
Could you add a line to the user output, explaining that metadata was not updated because contract is unchanged?

A `log::debug!` when deciding to not build further would also be helpful.

ascjones:
> Could you add a line to the user output, explaining that metadata was not updated because contract is unchanged?
> 
> A `log::debug!` when deciding to not build further would also be helpful.

There is already a `tracing::info!` there for this case (we no longer use `log`).

Here is the message saying artifacts are unchanged:

image

PR #966, opened 2023-02-10
Error if mismatching # of args for instantiate/call

ascjones:
Currently if you forget to supply `--args` to an instantiate/call, it will fail with a trap because of invalid input. e.g calling a constructor which has args but forgetting to add `--args`:

```
contract contract instantiate --suri //Alice
```

This PR adds a simple check which validates the number of args supplied is equal to the expected number of arguments to the ink! constructor or message.


cmichi (approved):
Nice!

PR #967, opened 2023-02-10
Pretty-print call dry-run return data

ascjones:
Regression introduced in #722 where it uses the `Debug` output for the returned dry-run value, instead of the `Display`. See before and after:

image


cmichi (approved):
Someone at the workshop actually ran into this! Thanks for fixing!

PR #970, opened 2023-02-14
Release `v2.0.0`

ascjones:
## [2.0.0]

Major release compatible with `ink! 4.0.0`. All the changes in aggregate since `1.5`:

### Added
- Add support for ink!'s `version` metadata field - [#641](https://github.com/paritytech/cargo-contract/pull/641)
- Add Rust specific build info to metadata - [#680](https://github.com/paritytech/cargo-contract/pull/680)
- Log code hash if contract is already uploaded - [#805](https://github.com/paritytech/cargo-contract/pull/805)
- Add remove command - [#837](https://github.com/paritytech/cargo-contract/pull/837)

### Changed
- Build contracts and dylint driver with stable - [#698](https://github.com/paritytech/cargo-contract/pull/698)
- Compile dylints when compiling the contract - [#703](https://github.com/paritytech/cargo-contract/pull/703)
- Move transcode example to doc test, add helper method - [#705](https://github.com/paritytech/cargo-contract/pull/705)
  - Note that alongside this PR we released [`contract-transcode@0.2.0`](https://crates.io/crates/contract-transcode/0.2.0)
- Replace custom RPCs by `state_call` - [#701](https://github.com/paritytech/cargo-contract/pull/701)
- Removed requirement to install binaryen. The `wasm-opt` tool is now compiled into `cargo-contract` - [#766](https://github.com/paritytech/cargo-contract/pull/766).
- Make linting opt in with `--lint` - [#799](https://github.com/paritytech/cargo-contract/pull/799)
- Update to weights v2 - [#809](https://github.com/paritytech/cargo-contract/pull/809)
- Update validation for renamed FFI functions - [#816](https://github.com/paritytech/cargo-contract/pull/816)
- Denominated units for balances in events - [#750](https://github.com/paritytech/cargo-contract/pull/750)
- Upgrade wasm-opt to 0.110.2 - [#802](https://github.com/paritytech/cargo-contract/pull/802)
- Pass `--features` through to `cargo` - [#853](https://github.com/paritytech/cargo-contract/pull/853/files)
- Remove rust toolchain channel check - [#848](https://github.com/paritytech/cargo-contract/pull/848/files)
- Bump minimum requirement of `scale-info` in template to `2.3` - [#847](https://github.com/paritytech/cargo-contract/pull/847/files)
- Remove `unstable` module check, add `--skip-wasm-validation` - [#846](https://github.com/paritytech/cargo-contract/pull/846/files)
[...]

PR #971, opened 2023-02-16
Bump template to ink! 4.0

HCastano:
Looks like we forgot to do this in #970.

Not a huge deal since `cargo` will still pull in ink! 4.0, but we'll
probably want to issue a patch release soon in order to not confuse
users.


cmichi (approved):
We should also add an E2E test and the `ink_e2e` dev-dep to the template. Can you do it?

PR #982, opened 2023-02-24
Limit input length for `decode` command

### LINKED ISSUE #980: Lack of input validation in the `decode` command
# Summary
The Cargo contract has a decode command to parse the encoded input or output data and extract the underlying values. The feature has two flags. One to indicate the type of data to decode and the other for data itself that has to be a hex value. However, the current implementation of the function is accepting more bytes than the target type expects, which could cause misinterpretation of the data.

# Tasks

- [x] Reproducer
- [x] PR Fix

SkymanOne:
Closes #980 

The hex bytes are consumed for decoding values of arguments on demand. If there is not enough data to fill the buffer, the error is already thrown. Previously, spare bytes were just ignored. Since this PR, an error will be thrown if input buffer contains spare bytes.

Here is the small demo demonstrating that extra bytes in buffer do not affect decoding of value.
```rust
//boolean is 1 byte long
let value = true;
let encoded: Vec<u8> = vec![1, 12, 3];
let mut encoded: &[u8] = encoded.as_ref();
let decoded: bool = bool::decode::<&[u8]>(&mut encoded).unwrap();
// is true
assert_eq!(decoded, value);
//prints [12, 3]
println!("{encoded:?}")
```

### DIFF
--- a/crates/transcode/src/lib.rs
+++ b/crates/transcode/src/lib.rs
@@ -361,7 +361,7 @@ impl ContractMessageTranscoder {
                 args.iter().fold(format!("`{label}`"), |init, arg| {
                     format!("{}, `{}`", init, arg.0)
                 });
-            let encoded_bytes = util::encode_bytes(data);
+            let encoded_bytes = hex::encode_upper(data);
             return Err(anyhow::anyhow!(
                 "input length was longer than expected by {} byte(s).\nManaged to decode {} but `{}` bytes were left unread",
                 data.len(),
--- a/crates/transcode/src/lib.rs
+++ b/crates/transcode/src/lib.rs
@@ -786,7 +786,9 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "input length was longer than expected by 1 byte(s)")]
+    #[should_panic(
+        expected = "input length was longer than expected by 1 byte(s).\nManaged to decode `flip` but `00` bytes were left unread"
+    )]
     fn fail_decode_input_with_extra_bytes() {
         let metadata = generate_metadata();
         let transcoder = ContractMessageTranscoder::new(metadata);
@@ -794,6 +796,23 @@ mod tests {
         let encoded_bytes = hex::decode("633aa55100").unwrap();
         let _ = transcoder
             .decode_contract_message(&mut &encoded_bytes[..])
+            .unwrap();
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "input length was longer than expected by 2 byte(s).\nManaged to decode `Event1`, `name`, `from` but `0C10` bytes were left unread"
+    )]
+    fn fail_decode_contract_event_with_extra_bytes() {
+        let metadata = generate_metadata();
+        let transcoder = ContractMessageTranscoder::new(metadata);
+
+        // raw encoded event with event index prefix
+        let encoded = (0u8, [0u32; 8], [1u32; 8], [12u8, 16u8]).encode();
+        // encode again as a Vec<u8> which has a len prefix.
+        let encoded_bytes = encoded.encode();
+        let _ = transcoder
+            .decode_contract_event(&mut &encoded_bytes[..])
             .unwrap();
     }
 }
--- a/crates/transcode/src/lib.rs
+++ b/crates/transcode/src/lib.rs
@@ -278,9 +278,17 @@ impl ContractMessageTranscoder {
         }
 
         if !data.is_empty() {
+            let arg_list_string: String = args
+                .iter()
+                .fold(format!("`{}`", event_spec.label()), |init, arg| {
+                    format!("{}, `{}`", init, arg.0)
+                });
+            let encoded_bytes = util::encode_bytes(data);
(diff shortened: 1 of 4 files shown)


crates/transcode/src/lib.rs
```diff
@@ -277,6 +277,13 @@ impl ContractMessageTranscoder {
             args.push((Value::String(name), value));
         }
 
+        if !data.is_empty() {
```
ascjones, line 280: There is one other usage of this`decode_contract_event` method, for decoding emitted contract events: https://github.com/paritytech/cargo-contract/blob/0de5b430c22361b33abb5873fec8823b95e69108/crates/cargo-contract/src/cmd/extrinsics/events.rs#L215

I believe it should continue to work, but it is worth double checking

ascjones (commented):
The alternative approach would have been to do the bytes length check just once directly in `decode.rs` after the decoding is complete. This way is also okay, with adding it to each of the called methods, although as mentioned `decode_contract_event` is used in one other place so we need to check whether that still works.

crates/transcode/src/lib.rs
SkymanOne: Checked. It used when parsing events from the result of `call` or `instantiate` commands. Everything works.

crates/transcode/src/lib.rs
```diff
@@ -739,4 +760,16 @@ mod tests {
...
+
+    #[test]
+    #[should_panic(expected = "input length was longer than expected by 1 byte(s)")]
+    fn fail_decode_input_with_extra_bytes() {
+        let metadata = generate_metadata();
+        let transcoder = ContractMessageTranscoder::new(metadata);
+
+        let encoded_bytes = hex::decode("633aa55100").unwrap();
+        let _ = transcoder
+            .decode_contract_message(&mut &encoded_bytes[..])
```
ascjones, line 772: *nit* this does only test one of the code paths which has the length check.
So could add unit tests for those other 2 `decode_` functions.

Or what would be really nice would be to update the existing integration tests to test these cases: https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/tests/decode.rs#L27

crates/transcode/src/lib.rs
```diff
@@ -277,6 +277,13 @@ impl ContractMessageTranscoder {
             args.push((Value::String(name), value));
         }
 
+        if !data.is_empty() {
+            return Err(anyhow::anyhow!(
```
HCastano, line 281: So I don't think we need to bail out entirely here. 

Let's say someone accidentally puts extra bytes into the `decode` command, it could still be useful to show them which bytes were successfully decoded as well as the extra ones.

E.g

```
Managed to decode "get", but had "BE EF" leftover.
```

crates/transcode/src/lib.rs
```diff
@@ -331,6 +361,21 @@ impl ContractMessageTranscoder {
...
+                    format!("{}, `{}`", init, arg.0)
+                });
+            let encoded_bytes = util::encode_bytes(data);
+            return Err(anyhow::anyhow!(
+                "input length was longer than expected by {} byte(s).\nManaged to decode {} but `{}` bytes were left unread",
+                data.len(),
+                arg_list_string,
+                encoded_bytes
+            ))
+        }
```
ascjones, line 377: The code to produce this error message is duplicated 3 times so would be good to either extract to a function, or perform one time at a higher level (e.g. in `decode.rs`).

SkymanOne: It would require me to push the error to the higher level. I'll see what works the best

crates/transcode/src/lib.rs
```diff
@@ -347,6 +353,24 @@ impl ContractMessageTranscoder {
             Ok(Value::Unit)
         }
     }
+
+    /// Checks if buffer empty, otherwise returns am error
+    fn validate_length(data: &[u8], label: &str, args: &[(Value, Value)]) -> Result<()> {
```
SkymanOne, line 358: Extracted check to a separate function instead of moving it to a higher layer because each decode function returns a generic `anyhow` error. We would need to return a custom error and map all `anyhow` errors inside each `decode` function to a specific error. We will also need to handle the error on the higher `decode` function. Additionally, events and messages/constructors have different structs that are used to extract `label`. That will also unnecessary overcomplicate error handling.

Therefore, it is better that all validation happens on the lower level at each `decode` function independently.

crates/transcode/src/util.rs
```diff
@@ -18,3 +18,7 @@
 pub fn decode_hex(input: &str) -> Result<Vec<u8>, hex::FromHexError> {
     hex::decode(input.trim_start_matches("0x"))
 }
+
+pub fn encode_bytes(buffer: &[u8]) -> String {
```
ascjones, line 22: This seems redundant since it is only used once and you could just do the `hex::encode_upper(buffer)` inline.

ascjones (approved):
LGTM

ascjones:
Since @HCastano had also reviewed this, would also have been nice to ask for his approval too.

PR #992, opened 2023-02-28
Return artifact paths when contracts unchanged

ascjones:
*WIP*

Since #964, if the contract is unchanged the contract optimization and metadata generation steps are skipped. Currently this means that the `optimization_result` and `metadata_result` fields on `BuildResult` are both `None`. Tools and library users depend on those results to get the paths to the contract Wasm and metadata artifacts. So on a fresh contract build those would have values, and on a subsequent build with no changes to the contract those results will both be `None`

This PR ensures that users can still determine those artifact paths even if the optimization and metadata steps are skipped.

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -613,7 +613,7 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
         }
     };
 
-    let build = || -> Result<(Option<(OptimizationResult, BuildInfo)>, BuildSteps)> {
+    let build = || -> Result<(Option<OptimizationResult>, BuildInfo, PathBuf, BuildSteps)> {
         let mut build_steps = BuildSteps::new();
         let pre_fingerprint = Fingerprint::try_from_path(&crate_metadata.original_wasm)?;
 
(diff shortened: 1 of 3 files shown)


crates/build/src/lib.rs
```diff
@@ -30,6 +30,9 @@ mod validate_wasm;
 mod wasm_opt;
 mod workspace;
 
+#[deprecated(since = "2.0.2", note = "Use MetadataArtifacts instead")]
+pub use self::metadata::MetadataArtifacts as MetadataResult;
```
ascjones, line 34: Renamed this type, so have added this alias to be backwards compatible.

cmichi:
image

Missing line break here.

crates/build/src/tests.rs
```diff
@@ -507,31 +510,97 @@ fn unchanged_contract_skips_optimization_and_metadata_steps(
...
+        res2.metadata_result.as_ref().unwrap().dest_bundle.exists(),
+        "Contract bundle should have been generated"
     );
 
     Ok(())
 }
 
+/// Get the last modified date of the given file.
+fn file_last_modified(path: &Path) -> SystemTime {
+    fs::metadata(path).unwrap().modified().unwrap()
```
SkymanOne, line 601: Should the `Result` be handled here? Are we sure that the file exists under given path?

ascjones:
> image
> 
> Missing line break here.

I believe I fixed that in https://github.com/paritytech/cargo-contract/pull/992/commits/a153e2a2fbba5abee18daf8793aa0bef389664dc (see image in PR description)

crates/build/src/tests.rs
ascjones: At the moment we are assuming the file exists, yes. In the cases it is used in the tests that should be true. That said I have replaced the `unwrap`s with `ok_or_else` to properly display the error message in case of an error to make it easier to debug these errors if they arise. https://github.com/paritytech/cargo-contract/pull/992/commits/dfc92152e9816aae151e5ae981411c293d5b8667

PR #998, opened 2023-03-02
Add `encode` command

### LINKED ISSUE #986: Add sub-command `cargo contract encode`
```
cargo contract encode [options] --message <MESSAGE> --args <ARGS>
```

There is currently no encoding command in `cargo contract`. An encode function would
allow developers to encode the arguments of a function call, making it easier to test different
scenarios and use cases and debug the code.

SkymanOne:
Closes #986

Some changes:
- extracted `contract_artifacts()` in extrinsics into a separate public function `find_contract_artifacts()` because we need it in order to retrieve the metadata, but `ExtrinsicOpts` also require `--suri` to be specified that we don't need for encoding.
- Added additional E2E test to test encoding


crates/cargo-contract/tests/decode.rs
```diff
@@ -170,3 +170,88 @@ fn decode_works() {
         .failure()
         .stderr(predicates::str::contains(error_msg));
 }
+
+#[test]
+fn encode_works() {
```
ascjones, line 175: Nitpick: this file is called `decode.rs`

crates/cargo-contract/src/cmd/encode.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+impl EncodeCommand {
+    pub fn run(&self) -> Result<()> {
+        let artifacts =
+            find_contract_artifacts(self.manifest_path.as_ref(), self.file.as_ref())?;
+        let transcoder = artifacts.contract_transcoder()?;
+
+        let call_data = transcoder.encode(&self.message, &self.args)?;
+        let call_data_encoded = hex::encode_upper(call_data);
```
ascjones, line 54: Why have you chosen to use upper case hex? I don't mind it, but we should be mindful of consistency where we have used `hex::encode` in other places. If uppercase hex is preferable we should have it everywhere, I have never considered the pros and cons before though.

ascjones (commented):
Looks good, just a couple of nitpicks and questions

crates/cargo-contract/src/cmd/encode.rs
SkymanOne: Because that's how we specify custom selectors in tests in examples

ascjones: True, but this is showing the whole encoded message not just the selector. 

It seems to be a matter of taste whether hex values are preferred to be expressed as upper or lowercase. I don't really have an opinion, I just think it should be consistent within this tool. So if we prefer upper case hex representation of bytes then we should do that everywhere (in `cargo-contract`)

SkymanOne: I did it in other places too

ascjones: I can still find a few other places with `hex::encode` though (okay we don't want to change it for the `LowerHex` impls. It doesn't matter that much in the end though. Just nice to be consistent.

ascjones (approved):
LGTM

PR #999, opened 2023-03-02
Dry-run `instantiate`, `call` and `upload` commands by default

### LINKED ISSUE #979: Default to dry-run's for sub-commands + Add `--exec`
The usage of the call command together with the `--skip-dry-run` and `--dry-run` flags has posed to be not ideal for users. 

We had discussed this and came up with the following ideas to improve the usability.

### Default to dry-run's by default

This should be done in all commands: `instantiate`, `call`, etc.
Explicitly write to stdout that "no state changes were performed".

### Add a new flag `--exec` (or similar) to `call`, `instantiate`, etc.

Using this flag will issue the transaction on-chain.

### Add a new sub-command `cargo contract exec`

It will be a shortcut for `cargo contract call --exec`.

SkymanOne:
Closes #979 

calling command like
```
cargo contract call --suri //Alice --message flip --contract 5DbdchiXhFDLPjfu7xmaCofpFbDXoPeXSmwXep3PKv9TGSqK
```
will dry run it without actually submitting the extrinsic.

Adding flag `-x/--execute` will actually execute it.

### DIFF
--- a/crates/cargo-contract/src/cmd/extrinsics/mod.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/mod.rs
@@ -124,8 +124,8 @@ pub struct ExtrinsicOpts {
     #[clap(flatten)]
     verbosity: VerbosityFlags,
     /// Dry-run the extrinsic via rpc, instead of as an extrinsic. Chain state will not be mutated.
-    #[clap(long)]
-    dry_run: bool,
+    #[clap(short('x'), long)]
+    execute: bool,
     /// The maximum amount of balance that can be charged from the caller to pay for the storage
     /// consumed.
     #[clap(long)]
@@ -291,7 +291,7 @@ impl ContractArtifacts {
 }
 
 /// The Wasm code of a contract.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct WasmCode(Vec<u8>);
 
 impl WasmCode {
--- a/crates/cargo-contract/src/cmd/extrinsics/mod.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/mod.rs
@@ -361,8 +361,8 @@ pub fn display_contract_exec_result_debug<R, const WIDTH: usize>(
     Ok(())
 }
 
-pub fn display_dry_run_result_warning() {
-    println!("You call {} been executed.", "has not".bold());
+pub fn display_dry_run_result_warning(command: &str) {
+    println!("Your {} call {} been executed.", command, "has not".bold());
     println!(
             "To submit the transaction and execute the call on chain, add {} flag to the command.",
             "-x/--execute".bold()
--- a/crates/cargo-contract/src/cmd/extrinsics/mod.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/mod.rs
@@ -141,27 +141,10 @@ pub struct ExtrinsicOpts {
 impl ExtrinsicOpts {
     /// Load contract artifacts.
     pub fn contract_artifacts(&self) -> Result<ContractArtifacts> {
-        let artifact_path = match (self.manifest_path.as_ref(), self.file.as_ref()) {
-            (manifest_path, None) => {
-                let crate_metadata = CrateMetadata::from_manifest_path(manifest_path)?;
-
-                if crate_metadata.contract_bundle_path().exists() {
-                    crate_metadata.contract_bundle_path()
-                } else if crate_metadata.metadata_path().exists() {
-                    crate_metadata.metadata_path()
-                } else {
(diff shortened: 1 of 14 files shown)


HCastano:
How come you went with this approach instead of renaming commands, e.g to `query` for example?

SkymanOne:
I felt like "query" is not present anywhere in the docs + it is usually associated with retrieval of data from DB. So having `call` and `call -x` is more intuitive since the same logic is in `cargo-release`.

I actually forgot to add `cargo contract exec`. I'll add it on Monday if you think we need it.

crates/cargo-contract/src/main.rs
```diff
@@ -121,6 +121,9 @@ enum Command {
     /// Call a contract
     #[clap(name = "call")]
     Call(CallCommand),
+    /// Executes a call
+    #[clap(name = "exec")]
+    Exec(CallCommand),
```
ascjones, line 126: I think this would be confusing. Either we use the `--exec` option or have an `exec` command, but not both.

ascjones: I realise this is what is specified in https://github.com/paritytech/cargo-contract/issues/979

SkymanOne: So what's the consensus here? Do we want to have both or keep just one?

SkymanOne:
Added warning message for dry-run that `-x/--execute` flag needs to be added to submit the transaction and execute the call on chain

crates/cargo-contract/src/main.rs
ascjones: At the moment each command maps directly to its corresponding dispatchable in `pallet-contracts`: `upload`, `remove` `instantiate`, `call`. And each of those (except `remove`) can be run via RPC with no effect ("dry run"), or run as a transaction ("executed"). So having a flag makes sense for running each command in a particular mode for this mental model.

So having a new `exec` command as a shortcut doesn't fit with this model. It may well be nice to have, but we have to ask is having to type `call --exec` instead of just `exec` that much of an improvement, to justify the introduction of a new command and new terminology?

I am open to having `exec` and/or `query` commands, with good justification. We could do these as simple follow up PRs

SkymanOne: Agreed. It's not a big deal. I'll remove `exec` command then

crates/cargo-contract/src/cmd/extrinsics/mod.rs
```diff
@@ -361,6 +361,14 @@ pub fn display_contract_exec_result_debug<R, const WIDTH: usize>(
     Ok(())
 }
 
+pub fn display_dry_run_result_warning() {
+    println!("You call {} been executed.", "has not".bold());
```
ascjones, line 365:
```suggestion
    println!("Your call {} been executed.", "has not".bold());
```
Also I would not use the word `call` here, since it has an overloaded and specific meaning in this context. i.e. `instantiate`, `call`...

So you could pass the name of the command to this function or maybe use transaction instead.

crates/cargo-contract/src/cmd/extrinsics/upload.rs
```diff
@@ -78,50 +79,51 @@ impl UploadCommand {
...
-                            result: String::from("Success!"),
-                            code_hash: format!("{:?}", result.code_hash),
-                            deposit: result.deposit,
-                        };
-                        if self.output_json {
-                            println!("{}", upload_result.to_json()?);
-                        } else {
-                            upload_result.print();
-                        }
+            match self.upload_code_rpc(code.clone(), &client, &signer).await? {
```
ascjones, line 82: Same here only do dry run if not `--execute`

ascjones (changes requested):

crates/cargo-contract/src/cmd/extrinsics/upload.rs
```diff
@@ -114,14 +115,14 @@ impl UploadCommand {
                 } else {
                     upload_result.print();
                 }
-                Ok(())
             } else {
                 let code_hash = hex::encode(code_hash);
-                Err(anyhow::anyhow!(
-                    "This contract has already been uploaded with code hash: 0x{code_hash}"
-                )
+                return Err(anyhow::anyhow!(
```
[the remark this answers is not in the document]
SkymanOne: You cannot convert `bail!` into desired error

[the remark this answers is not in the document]
ascjones: Cool why the refactor to to return errs and put the `Ok()` at the end anyway. Totally fine of course just wondering the reasoning for doing that as part of this.

Also please fix indentation here if possible.

PR #1001, opened 2023-03-03
Change default `opt-level` to `s`

agryaznov:
It [turned out](https://github.com/paritytech/ink/pull/1702) that being optimized with this option, the most of our example contracts get smaller wasm blob.


ascjones (commented):
So we just need to wait for the waterfall to pass on https://github.com/paritytech/ink/pull/1702, and then go ahead and merge if the numbers look good.

ascjones:
https://github.com/paritytech/ink/pull/1702 closed

Issue #796, opened 2022-10-24
Add flag `--output-json` for `cargo contract info`


cmichi:
Follow-up to https://github.com/paritytech/cargo-contract/issues/783.

The output of `cargo contract info` should be written out as JSON when `cargo contract info --output-json` is invoked.

We already have this flag for `cargo contract build`, so the code for this functionality already exists in `cargo-contract`:

```
$ cargo contract build --help
…
      --output-json
          Export the build output in JSON format
```

PR #993, opened 2023-03-01
Add `info` command

lean-apple:
Closes https://github.com/paritytech/cargo-contract/issues/783

###  Done : 

- [x] Decode Result from API Storage Contract Info of 
- [x] Handing if no contract/AccountId is provided - Add a check before making the request ? 
- [x] Format the result 
- [x] Add Integration Test 
- [x] Add documentation for the new command

PR #1007, opened 2023-03-08
Add flag `--output-json` for  `info` command

lean-apple:
In the wake of https://github.com/paritytech/cargo-contract/pull/993

Closes https://github.com/paritytech/cargo-contract/issues/796


lean-apple:
Current state :
Screenshot 2023-03-10 at 10 52 51

lean-apple:
I still need to add a test for the flag and to prepare the global tests for the upcoming flags but open to comments.

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -94,15 +112,34 @@ impl InfoCommand {
...
+    storage_item_deposit: u128,
+}
 
-    fn print_and_format_contract_info(info: ContractInfo) {
-        let convert_trie_id = hex::encode(info.trie_id.0);
-        name_value_println!("TrieId:", format!("{}", convert_trie_id));
-        name_value_println!("Code hash:", format!("{:?}", info.code_hash));
-        name_value_println!("Storage items:", format!("{:?}", info.storage_items));
+impl InfoToJson {
+    pub fn to_json(&self) -> Result<String> {
```
cmichi, line 126: Why `pub`?

cmichi, line 126: For `pub` functions there should be comments.

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -67,7 +70,22 @@ impl InfoCommand {
 
             match info_result {
                 Some(info_result) => {
-                    InfoCommand::print_and_format_contract_info(info_result);
+                    let output_type = match self.output_json {
+                        true => OutputType::Json,
+                        false => OutputType::HumanReadable,
+                    };
```
ascjones, line 76: It is redundant to define and construct this enum. can just do `if self.output_json`.

I can see that it is copied from `build`, but in that case it is passed down to other commands via `ExecuteArgs` so makes more sense. Whereas in this case the conditional is evaluated inline.

crates/cargo-contract/src/cmd/info.rs
ascjones: IMO this method does not need to exist, can just do `serde_json::to_string_pretty(info_to_json)?` inline.

crates/cargo-contract/src/cmd/info.rs
lean-apple: Yep I hesitate to do this in this simple way

crates/cargo-contract/src/cmd/info.rs
lean-apple: Comments Added - as discussed with @ascjones after it is ok like this atm

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -50,6 +50,9 @@ pub struct InfoCommand {
         default_value = "ws://localhost:9944"
     )]
     url: url::Url,
+    /// Export the call output as JSON.
```
cmichi, line 53:
```suggestion
    /// Export the instantiate output in JSON format.
```

This is a user-facing text, we should use the same description as other sub-commands do for`--output-json`.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Add `info` command - [#993](https://github.com/paritytech/cargo-contract/pull/993)
+- Add `--output-json` flag for `info` command - [#1007](https://github.com/paritytech/cargo-contract/pull/1007)
```
cmichi, line 11:
```suggestion
```

cmichi (approved):
Thanks, Lea!

PR #1010, opened 2023-03-09
Custom `AccountId32` impl, remove substrate deps

ascjones:
Removes the substrate dependencies from `contract-transcode`.

We copy the `AccountId32` type from `subxt`: https://github.com/paritytech/subxt/blob/ce0a82e3227efb0eae131f025da5f839d9623e15/subxt/src/utils/account_id.rs, with a few modifications so that it can work in place of `sp_core::crypto::AccountId32`

### DIFF
--- a/crates/transcode/Cargo.toml
+++ b/crates/transcode/Cargo.toml
@@ -18,6 +18,8 @@ path = "src/lib.rs"
 
 [dependencies]
 anyhow = "1.0.69"
+base58 = { version = "0.2.0" }
+blake2 = { version = "0.10.4", default-features = false }
 contract-metadata = { version = "2.0.2", path = "../metadata" }
 escape8259 = "0.5.2"
 hex = "0.4.3"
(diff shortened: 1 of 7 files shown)


ascjones:
Clippy CI failure fixed in https://github.com/paritytech/cargo-contract/pull/1011

SkymanOne (approved):
I assume the maths for account derivation is correct :)

ascjones:
> I assume the maths for account derivation is correct :)

Indeed, all copied from `substrate`. Though would be nice to have some tests for that...

ascjones:
> > I assume the maths for account derivation is correct :)
> 
> Indeed, all copied from `substrate`. Though would be nice to have some tests for that...

Actually there is already a test at the bottom of the file that checks those using the account keyring.

SkymanOne:
`contract-transcode` build was failing due to missing features in `primitive-types` dependency. See fix here - https://github.com/paritytech/cargo-contract/pull/1016/commits/d89ca994b84f4b519bca1a11b6c3b5fec220b603

crates/transcode/Cargo.toml
```diff
@@ -28,16 +30,18 @@ itertools = "0.10.5"
...
 serde = { version = "1.0.154", default-features = false, features = ["derive"] }
 serde_json = "1.0.94"
-sp-core = "16.0.0"
-sp-runtime = "18.0.0"
+thiserror = "1.0.39"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
 ink = "4.0.1"
+sp-core = { version = "18.0.0", default-features = false }
```
HCastano, line 43: Is `default-features = false` needed here?

ascjones: No, probably not.

ascjones: Tacked it on to https://github.com/paritytech/cargo-contract/pull/1026

Issue #800, opened 2022-10-27
Output pristine Wasm code for `cargo contract info`


cmichi:
Follow-up to https://github.com/paritytech/cargo-contract/issues/783.

## Goal

`cargo contract info --contract …` should output the contract's pristine Wasm blob. 

By default the output should be hex-encoded. A flag `--binary` should be added, once supplied the binary Wasm blob should be output.

## Background

The `pallet-contracts` stores two versions of the Wasm contract blob on-chain:

* The "pristine" Wasm blob is exactly what was uploaded to the chain by the deployer.
* The "instumented" Wasm blob was enriched with instructions for gas metering.

## How to get this info

The `pallet-contracts` exposes a method for getting this info. `subxt` should already generate code for it as well.

<img width="1593" alt="Screenshot 2022-10-26 at 01 07 32" src="https://user-images.githubusercontent.com/241530/198212514-afbb8c14-8fc6-4395-9e60-f085d161032b.png">

Something along those lines should be possible for `subxt`:

```rust
let info_addr = node_runtime::storage()
    .contracts()
    .pristine_code(&contract);
```

PR #1013, opened 2023-03-09
Add flag `--binary` for  `info`command

lean-apple:
In the wake of https://github.com/paritytech/cargo-contract/pull/993

Closes https://github.com/paritytech/cargo-contract/issues/800


lean-apple:
The output for the pritstine_code seems to correspond but it needs to be formatted now. 
Screenshot 2023-03-09 at 20 47 39 (2)

cmichi:
@lean-apple Please read the original issue https://github.com/paritytech/cargo-contract/issues/800 again and adapt this PR to those specifications (namely binary output if `--binary` is supplied without `--output-json`).

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -85,29 +88,34 @@ impl InfoCommand {
...
                                 if self.binary {
-                                    println!("{}", pristine_code.to_json()?);
-                                } else {
                                     println!("{}", info_to_json.to_json()?);
+                                } else {
+                                    println!("{}", basic_info.to_json()?);
                                 }
                             } else if self.binary {
-                                println!("{:?}", pris_w.0)
+                                std::io::stdout().write_all(&pris_w.0).expect("");
```
cmichi, line 115:
```suggestion
                                std::io::stdout().write_all(&pris_w.0).expect("writing to stdout failed");
```

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -110,30 +153,61 @@ impl InfoCommand {
...
+        name_value_println!(
+            "Storage deposit:",
+            format!("{:?}", info.storage_item_deposit)
+        );
+    }
 }
 
 #[derive(serde::Serialize)]
-struct InfoToJson {
+struct GlobalInfoToJson {
```
cmichi, line 185: Could you add a Doc comment for this struct?

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -55,6 +61,9 @@ pub struct InfoCommand {
     /// Export the instantiate output in JSON format.
     #[clap(name = "output-json", long)]
     output_json: bool,
+    /// Display the pristine Wasm code.
```
cmichi, line 64:
```suggestion
    /// Display the contract's Wasm bytecode.
```

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -110,30 +153,61 @@ impl InfoCommand {
...
+        let pristine_code_call = api::storage().contracts().pristine_code(hash);
+
+        let prinstine_bytes = client
+            .storage()
+            .at(None)
+            .await?
+            .fetch(&pristine_code_call)
+            .await?;
+
+        Ok(prinstine_bytes)
```
cmichi, line 170:
```suggestion
        client
            .storage()
            .at(None)
            .await?
            .fetch(&pristine_code_call)
            .await
```

cmichi:
I've given it some more thought how people will use it and I think we should adapt the behavior a bit.

Could you change the behavior to this, please?

* `cargo contract info --contract …`: Output everything line by line, including the Wasm bytecode in hex.
* `cargo contract info --contract … --binary`: Output just the binary Wasm bytecode, which is what this PR does now, so no change needed.
* `cargo contract info --contract  … --output-json`: Output the JSON, include the Wasm bytecode in hex.

crates/cargo-contract/src/cmd/info.rs
cmichi: The context for this change is that "pristine" is a term that is used only in `pallet-contracts` internally. So for user-facing output of `cargo contract` we should just use `wasm` instead of `pristine_wasm`.

PR #1016, opened 2023-03-10
`2.1.0` Release

SkymanOne:
### Changed
- Dry-run `instantiate`, `call` and `upload` commands by default - [#999](https://github.com/paritytech/cargo-contract/pull/999)

### Added
- Add `cargo contract encode` command - [#998](https://github.com/paritytech/cargo-contract/pull/998)

### Fixed
- Limit input length for `decode` command - [#982](https://github.com/paritytech/cargo-contract/pull/982)
- Pass contract features to metadata gen package - [#1005](https://github.com/paritytech/cargo-contract/pull/1005)
- Custom AccountId32 impl, remove substrate deps - [#1010](https://github.com/paritytech/cargo-contract/pull/1010)
  - Fixes issue with with incompatible `wasmtime` versions when dependant project has old substrate dependencies.

PR #1017, opened 2023-03-13
Add CI check for all individual crates

ascjones:
Currently the CI only runs workspace level `build` and `test` commands. However it is possible that these will succeed but when building an individual crate it will fail. e.g the CI for https://github.com/paritytech/cargo-contract/pull/1010 succeeded, but when it came to releasing it, it didn't work because of [missing features](https://github.com/paritytech/cargo-contract/pull/1016/commits/d89ca994b84f4b519bca1a11b6c3b5fec220b603).

My guess is that there is another crate which depends on `primitive-types` at the workspace level, which pulls in the required features via feature unification.

This PR iterates over the crates in the `crates` dir and runs `check` on each one. When performing a release on an individual crate, `cargo publish` will run `cargo check` before publishing, so we want to catch any errors which would arise there in the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,28 @@ jobs:
           command: fmt
           args: --check
 
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+
+      - name: Cache
+        uses: Swatinem/rust-cache@v2
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Check each crate
+        run: |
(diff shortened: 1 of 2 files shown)


SkymanOne (approved):
Thanks!

.github/workflows/ci.yml
```diff
@@ -30,6 +30,28 @@ jobs:
...
+          profile: minimal
+          toolchain: stable
+
+      - name: Cache
+        uses: Swatinem/rust-cache@v2
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Check each crate
```
HCastano, line 48: @ascjones do you mind making a small follow-up explaining why we need this check?

For someone just reading the CI file they wouldn't have enough context to know why we manually check all crates

PR #1032, opened 2023-03-16
Speed up CI :t-rex: 

ascjones:
Since https://github.com/paritytech/cargo-contract/pull/1028, the `test` jobs are taking considerably longer, from 10-20 mins up to 1 hour.

For now this PR is just running the tests without the integration tests to see how long this will take. I can think of a few solutions:

1) Ensure sharing of build artifacts between integration tests (similar to `build` tests)
2) Only run the integration tests before merging (use merge groups?)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,8 @@ jobs:
         shell: bash
 
       - name: Run tests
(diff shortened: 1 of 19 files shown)


ascjones:
Closing, might open a fresh PR when I have a solution

PR #1057, opened 2023-03-30
Upgrade `scale-info` to `2.5`

ascjones:
Accessors are deprecated and fields are made public, see https://github.com/paritytech/scale-info/pull/176


SkymanOne (approved):
😗

PR #1076, opened 2023-04-15
Initial RISC-V support

athei:
This PR adds initial support for RISC-V. It allows the user to supply `--target riscv` which then will build a contract into `riscv32i`. The feature is more or less complete in that it produces a contract bundle containing RISC-V bytecode. However, there is no executor where to run those contracts on, yet. Additionally, there is no post processing done on the resulting file. This will come later in order to bring the file size down.

## Changes

- Add `--target` command line flag. Defaults to `wasm`.
- Build contracts as `bin` target rather than `cdylib`. This is required for RISC-V but changes nothing for wasm.
- Change the `wasm` file extension for the contract code to `code` so that it is agnostic to the bytecode. We added a symlink from the `.wasm` file when building for wasm to have backwards compat.
- Add a preliminary linker script for RISC-V that links the code according to our current VM implementation. This is of course subject to change,
- Unify optimization and post processing steps. The differentiation seems arbitrary and makes no sense for RISC-V. Post processing encapsulates both now.
- 
## Caveats

- Some tests are failing because they use the latest released version of ink! which doesn't include https://github.com/paritytech/ink/pull/1718
- When building without `--release` the linking fails. This is an old bug we know from wasm which appears with `-Zbuild-std + lto + overflow-checks`. The idea is to get rid of overflow checks and ban unchecked arithmetic from contracts. I will present a solution for this soon.

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -41,6 +41,7 @@ pub use self::{
         Features,
         Network,
         OutputType,
+        Target,
         UnstableFlags,
         UnstableOptions,
         Verbosity,
@@ -83,6 +84,7 @@ use parity_wasm::elements::{
 use semver::Version;
 use std::{
     fs,
+    io::Write,
     path::{
         Path,
         PathBuf,
@@ -113,6 +115,7 @@ pub struct ExecuteArgs {
     pub lint: bool,
     pub output_type: OutputType,
     pub skip_wasm_validation: bool,
+    pub target: Target,
 }
 
 /// Result of the build process.
@@ -228,7 +231,7 @@ impl BuildResult {
 /// The `[workspace]` will be added if it is missing to ignore `workspace` from parent `Cargo.toml`.
 ///
 /// To disable this and use the original `Cargo.toml` as is then pass the `-Z original_manifest` flag.
-fn exec_cargo_for_wasm_target(
+fn exec_cargo_for_onchain_target(
     crate_metadata: &CrateMetadata,
     command: &str,
     features: &Features,
@@ -236,14 +239,15 @@ fn exec_cargo_for_wasm_target(
     network: Network,
     verbosity: Verbosity,
     unstable_flags: &UnstableFlags,
+    target: Target,
 ) -> Result<()> {
     let cargo_build = |manifest_path: &ManifestPath| {
         let target_dir = &crate_metadata.target_directory;
         let target_dir = format!("--target-dir={}", target_dir.to_string_lossy());
 
         let mut args = vec![
-            "--target=wasm32-unknown-unknown".to_owned(),
-            "-Zbuild-std".to_owned(),
+            format!("--target={}", target.llvm_target()),
+            "-Zbuild-std=core,alloc".to_owned(),
             "--no-default-features".to_owned(),
             "--release".to_owned(),
             target_dir,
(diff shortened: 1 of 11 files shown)


athei:
Ready for review now. I will make a follow up PR that takes care of this overflow checks mess.

crates/build/src/lib.rs
```diff
@@ -257,14 +263,31 @@ fn exec_cargo_for_wasm_target(
...
+            fs::create_dir_all(&crate_metadata.target_directory)?;
+            let path = crate_metadata
+                .target_directory
+                .join(".riscv_memory_layout.ld");
+            fs::write(&path, include_bytes!("../riscv_memory_layout.ld"))?;
+            env.push((
+                "RUSTFLAGS",
+                Some(format!(
+                    "{} -Clink-arg=-T{}",
+                    target.rustflags(),
```
ascjones, line 282: I'd prefer to see the `rustflags()` inline here and below for readability, we are already inside a match on the target.

crates/build/src/lib.rs
```diff
@@ -621,16 +644,29 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
...
+
+            // we changed the .wasm file ending to the target agnostic .code
+            // in order to not break existing code we create a symlink
+            let symlink = crate_metadata.dest_code.with_extension("wasm");
+            fs::remove_file(&symlink).ok();
+            #[cfg(unix)]
+            if matches!(target, Target::Wasm) {
+                std::os::unix::fs::symlink(&crate_metadata.dest_code, &symlink)?;
+            };
+
```
ascjones, line 669: Could we do this on windows too? https://doc.rust-lang.org/stable/std/os/windows/fs/fn.symlink_file.html

crates/build/src/lib.rs
```diff
@@ -611,8 +635,7 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
     let build =
         || -> Result<(Option<OptimizationResult>, BuildInfo, PathBuf, BuildSteps)> {
             let mut build_steps = BuildSteps::new();
-            let pre_fingerprint =
-                Fingerprint::try_from_path(&crate_metadata.original_wasm)?;
+            let (pre_fingerprint, target_file_path) = Fingerprint::new(&crate_metadata)?;
```
ascjones, line 638: Let's avoid this tuple return by constructing `target_file_path` here and passing it to `Fingerprint::new`

crates/build/src/lib.rs
```diff
@@ -650,34 +686,32 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
...
-                &crate_metadata.original_wasm,
-            )?
-            .ok_or_else(|| {
-                anyhow::anyhow!(
-                    "Expected '{}' to be generated by build",
-                    crate_metadata.original_wasm.display()
-                )
-            })?;
+            let post_fingerprint =
+                Fingerprint::new(&crate_metadata)?.0.ok_or_else(|| {
```
ascjones, line 690: By removing the tuple return we can avoid the anon field access here

crates/build/src/lib.rs
```diff
@@ -686,29 +720,40 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
...
+                        WasmOptHandler::new(optimization_passes, keep_debug_symbols)?;
+                    handler.optimize(
+                        &crate_metadata.dest_code,
+                        &crate_metadata.contract_artifact_name,
+                    )?;
+                }
+                Target::RiscV => {
+                    fs::copy(&crate_metadata.original_code, &crate_metadata.dest_code)?;
+                }
+            }
```
ascjones, line 743: A downside of having both targets produce the final `contract.code` file is that you can't distinguish from the file extension the format of the contents i.e. whether it contains `wasm` or `risc-v` code.

athei: I did this on purpose because I don't want stale build artifacts for other architectures lingering around in the directory. Then you still don't really now what is in your bundle without looking at timestamps. It looked to me as the more confusing choice.

crates/build/src/lib.rs
athei: This documentation says you need admin privileges to create a symlink (lol windows). This is why I didn't do it.

crates/build/src/lib.rs
athei: I think the best solution is to keep the arch specific extension and just clean up old files on rebuild.

ascjones: The alternative is to have the artifacts in target specific dirs, there already exists e.g. `target/ink/wasm32-unknown-unknown`

image

In any case we should add a field to the metadata which specifies the target code architecture, maybe somewhere in here
image

crates/build/src/lib.rs
athei: No longer relevant. We use arch specific file extensions now which means we keep the wasm file as-is.

crates/build/src/lib.rs
athei: Yes that might work to. But I went with keeping them in the top level dir and just clean up stale files. Metadata changes will be done in a follow up PR.

crates/build/src/lib.rs
athei: I moved the path up into `CrateMetadata`.

PR #1084, opened 2023-04-20
Update `ink!` requirements to `4.2.0`

ascjones:


athei (approved):
I wonder if we provide much value by bumping this on every minor. Since new projects will go to the newest minor automatically anyways. So we could just set it to `"4`" and save us some chores.

ascjones:
> I wonder if we provide much value by bumping this on every minor. Since new projects will go to the newest minor automatically anyways. So we could just set it to `"4`" and save us some chores.

It is not always necessary, but in this case I believe it might be since it includes some metadata additions e.g. https://github.com/paritytech/ink/pull/1741. So in this case e.g. we want to guarantee that the minimum version is `4.2` so that it includes those changes. It might be possible if another dependency has `"=4.0` for ink then it could pull in an older version still.

PR #1113, opened 2023-05-15
Suggest adding `no_main` for `error[E0601]`

ascjones:
Since https://github.com/paritytech/cargo-contract/pull/1076, `no_main` is now required. This PR adds a suggestion in case of the error `error[E0601]: `main` function not found in crate...

This is the result:
image

Note that the suggestion must appear *before* the compiler error because we can't know for sure how many lines the error message will encompass, so we just print it first.

Introduces [duct](https://docs.rs/duct/latest/duct/) which allows for easily building command expressions and working with capturing the `stdout` and `stderr`.

## Maintaing the progress bar

An earlier iteration was using `BufReader::lines` to read the captured `stderr` and check for the error code for the missing main function. However this came with the downside of the familiar progress bar (see below) not being displayed.

image

I have been able to get that to work again with some hacking. The result involves reading the reader one byte at a time and redirecting that straight to `stderr`, while maintaining a small buffer to check for the known error. I am aware that this is not the most efficient way, but it was the only way to get the progress bar displaying as it does when not capturing `stderr`.

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -326,6 +326,13 @@ fn exec_cargo_for_onchain_target(
 /// Executes the supplied cargo command, reading the output and scanning for known errors.
 /// Writes the captured stderr back to stderr and maintains the cargo tty progress bar.
 fn invoke_cargo_and_scan_for_error(cargo: duct::Expression) -> Result<()> {
+    macro_rules! eprintln_red {
+        ($value:expr) => {{
+            use colored::Colorize as _;
+            ::std::eprintln!("{}", $value.bright_red().bold());
+        }};
+    }
+
     let cargo = util::cargo_tty_output(cargo);
 
     let missing_main_err = "error[E0601]".as_bytes();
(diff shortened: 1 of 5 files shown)


crates/build/src/lib.rs
```diff
@@ -320,6 +323,46 @@ fn exec_cargo_for_onchain_target(
...
+        for byte in buffer[0..bytes_read].iter() {
+            err_buf.push_back(*byte);
+            if err_buf.len() > missing_main_err.len() {
+                err_buf.pop_front();
+            }
+        }
+        if missing_main_err == err_buf.make_contiguous() {
+            eprintln!(
+                "\n\n{}",
+                "Ensure the contract is annotated with `no_main` e.g.:"
```
cmichi, line 349:
```suggestion
                "Ensure the contract is annotated with `no_main`, e.g.:"
```

cmichi:
As discussed via dm, let's omit the tail end of the error output and exit like this:

```
$ cargo contract build
 [1/*] Building cargo project
   Compiling flipper v4.2.0 (/private/var/folders/s5/5gcp8ck95k39z006fj059_0c0000gn/T/cargo-contract_zJEp2z)

Exited with error: [E0601]
Your contract must be annotated with the `no_main` attribute.

Examples how to do this:
  - #![cfg_attr(not(feature = "std"), no_std, no_main)]
  - #[no_main]
```

(with the last part in red)

ascjones:
image

cmichi (approved):

Issue #1112, opened 2023-05-15
Make output format of `cargo contract info` consistent with other subcommands


cmichi:
## The Problem

This is the format we typically use:

```
$ cargo contract call --contract 5CVpqJFfaYGQv4TdsQ8Syx5wts2YT5ddD1ahDzSqGAFyjcRc  --suri //Alice --message was_it_ten
          Result ModuleError: Contracts::ContractTrapped: ["Contract trapped during execution."]
    Gas Consumed Weight { ref_time: 96029499, proof_size: 1961 }
    Gas Required Weight { ref_time: 9798418432, proof_size: 262144 }
 Storage Deposit StorageDeposit::Charge(0)
```

This is the format `cargo contract info` currently has:

```
$ cargo contract info --contract 5CVpqJFfaYGQv4TdsQ8Syx5wts2YT5ddD1ahDzSqGAFyjcRc                                        
     TrieId: 60e2e799fc7964173da23d9d7ed1d6fbbc19e0cb1008b32fed6b68bbc10f13fa
  Code hash: 0x7b6bb32c4cd224adcfe3e2fcc04d2ebf3f0f5b85303638b9a4cb08a283a4ef2b
Storage items: 1
Storage deposit: 100000000000
```

`cargo contract info` should use the same output format here. This is where the output happens: https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/info.rs#L129-L138.

And this is where the correct output format happens: https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/extrinsics/mod.rs#L328-L352.

## How to reproduce

(1) Start [`substrate-contracts-node`](https://github.com/paritytech/substrate-contracts-node) in one terminal.

(2) 
```
cd flipper/
cargo contract build
cargo contract instantiate --suri //Alice --constructor new_default -x
cargo contract info --contract 5GtGKWAPR41tJPjbutx8tLccKYLtrgv27Kg35uTN7K9DfH3r
```

ascjones:
Closed by #1120

PR #1120, opened 2023-05-19
Make output format of cargo contract info consistent with other subcommands

AlexD10S:
Change in the function basic_display_format_contract_info() to make the command `cargo contract info` use the same output format with other subcommands.
Fix of the Issue: https://github.com/paritytech/cargo-contract/issues/1112

Output:
Screenshot 2023-05-19 at 11 26 27


ascjones (commented):
Good stuff. Can you upload a screenshot of how the result looks?

ascjones (approved):

Issue #1021, opened 2023-03-14
Show Gas consumption by default for dry-runs


cmichi:
The console output should contain the Gas estimate for dry-runs.

Currently:

```bash
➜  flipper git:(master) ✗ cargo contract instantiate --suri //Alice --args true                             
      Result Success!
    Contract 5Hgzda4MwEL99i4mz2Szz43txpKafaEsBZu9N9me5LHkfArW
    Reverted false
        Data Bytes([0, 0])
Your instantiate call has not been executed.
To submit the transaction and execute the call on chain, add -x/--execute flag to the command.
```

Should include:

```bash
   Gas limit Weight(ref_time: 323301935, proof_size: 0)
```

PR #1121, opened 2023-05-19
Show Gas consumption by default for dry-runs

AlexD10S:
Modify the output in `cargo contract instantiate ` to show the gas consumed estimated when dry-run.
Fix the issue https://github.com/paritytech/cargo-contract/issues/1021

Output:
Screenshot 2023-05-19 at 15 57 17


crates/cargo-contract/src/cmd/extrinsics/instantiate.rs
```diff
@@ -467,6 +467,11 @@ impl InstantiateDryRunResult {
             DEFAULT_KEY_COL_WIDTH
         );
         name_value_println!("Data", format!("{:?}", self.data), DEFAULT_KEY_COL_WIDTH);
+        name_value_println!(
+            "Gas limit",
```
ascjones, line 471:
```suggestion
            "Gas consumed",
```
It is true that this value is used for the gas limit for transactions. But it is still the `Gas consumed` by the dry-run. Does it fit with the max col width?

PR #1123, opened 2023-05-19
Dry-run result output improvements

ascjones:
## Instantiate

- Replace useless "`Success!`" field with actual decoded data (it will be the decoded `Result<Result<(), _>>`).

Before:
image

After (The `Bytes` has now been decoded into `Ok()`)
image

## Call

Similarly, replaced the useless `"Success!"` with the decoded data.

Before:
image

After:
image


cmichi (approved):
Coolio! Possibly `CHANGELOG.md`.

agryaznov (approved):

Issue #1062, opened 2023-04-04
Add flag `-y` that says `Yes` to every prompt


cmichi:
As a shortcut for e.g. `--skip-confirm --skip-dry-run`.

PR #1127, opened 2023-05-24
Add flag -y as a shortcut for --skip-confirm

AlexD10S:
PR for this issue: Add flag -y that says Yes to every prompt (https://github.com/paritytech/cargo-contract/issues/1062)

It adds `-y` as a shortcut for `--skip-confirm`. 
Examplo of use:
`cargo contract instantiate --suri //Alice --constructor new  --args true -x -y` instead of `cargo contract instantiate --suri //Alice --constructor new  --args true -x --skip-confirm`

### DIFF
--- a/crates/cargo-contract/src/cmd/extrinsics/integration_tests.rs
+++ b/crates/cargo-contract/src/cmd/extrinsics/integration_tests.rs
@@ -201,8 +201,6 @@ async fn build_upload_instantiate_call() {
         .expect("failed to execute process");
     let stdout = str::from_utf8(&output.stdout).unwrap();
     let stderr = str::from_utf8(&output.stderr).unwrap();
-    println!("{:?}", stdout);
-    println!("{:?}", stderr);
     assert!(output.status.success(), "instantiate failed: {stderr}");
 
     let contract_account = extract_contract_address(stdout);


ascjones:
We should clarify then. @cmichi ?

cmichi:
Yup, I meant what Andrew said.

cmichi:
@AlexD10S Don't forget updating the `CHANGELOG.md`.

AlexD10S:
Does make sense to have at the same time  `y` with one of the flags `skip_confirm` or `skip_dry_run`? I don't see a problem, but in case is better to restrict the use will  be done adding `conflicts_with`.
```
    /// Before submitting a transaction, do not dry-run it via RPC first.
    #[clap(long, conflicts_with = "y")]
    skip_dry_run: bool,
    /// Before submitting a transaction, do not ask the user for confirmation.
    #[clap(long, conflicts_with = "y")]
    skip_confirm: bool,
```

ascjones:
> Does make sense to have at the same time `y` with one of the flags `skip_confirm` or `skip_dry_run`? I don't see a problem, but in case is better to restrict the use will be done adding `conflicts_with`.
> 
> ```
>     /// Before submitting a transaction, do not dry-run it via RPC first.
>     #[clap(long, conflicts_with = "y")]
>     skip_dry_run: bool,
>     /// Before submitting a transaction, do not ask the user for confirmation.
>     #[clap(long, conflicts_with = "y")]
>     skip_confirm: bool,
> ```

Yes I think it should have `conflicts_with`

crates/cargo-contract/src/cmd/extrinsics/mod.rs
```diff
@@ -133,11 +133,15 @@ pub struct ExtrinsicOpts {
     #[clap(long)]
     storage_deposit_limit: Option<BalanceVariant>,
     /// Before submitting a transaction, do not dry-run it via RPC first.
-    #[clap(long)]
+    #[clap(long, conflicts_with = "y")]
     skip_dry_run: bool,
     /// Before submitting a transaction, do not ask the user for confirmation.
-    #[clap(long)]
+    #[clap(long, conflicts_with = "y")]
```
ascjones, line 139: Possibly we could change this whole PR just by adding `y` as the short for `skip-confirm`

ascjones (changes requested):
Sorry for the delay, revisiting this. 

After reading the issue again, I believe it incorrectly states that `--skip-dry-run` is saying yes to a prompt. In fact, what it does is prevent the pre tx submission dry-run which is used for gas estimation. This can be used if the user wants to submit a transaction with their own custom limits, or if the dry-run fails and they want to submit the tx anyway.

Therefore I believe it doesn't make sense to group it with `--skip-confirm`. It may be true that when manually specifying gas limits the user then wants to skip the confirm step, but I think that is okay to have as a separate option still.

crates/cargo-contract/src/cmd/extrinsics/mod.rs
AlexD10S: No worries, is great to have your feedback.
This was my initial though too. Do you suggest to come back to the [initial commit](https://github.com/paritytech/cargo-contract/pull/1127/commits/01fc69a7ccba7925e52f9c13212ac0247201cc85#diff-0930d98a0a16379942c5f6a0dc0daabf6ed8c9d1bd0f56749fadaaeb8fc03e46L139).

ascjones: Yes looks good :+1:

ascjones (approved):

Issue #68, opened 2020-08-05
Solang support


ascjones:
Might be nice to be able to build contracts written in solidity, by invoking https://github.com/hyperledger-labs/solang. 

e.g. `cargo contract build flipper.sol` would detect the `sol` extension and attempt to build using `solang`. If the `solang` binary is not installed the user could be prompted to install it via https://solang.readthedocs.io/en/latest/installing.html.

atenjin:
Interesting. But I think we may do not need cargo contract to call Solang to compile. cargo-contract just  needs to take charge of ink self maybe better.

xermicus:
@ascjones I self-assigned this to me, it came up in our solang daily stand-up. I can look at potential solutions for this (after we fix the metadata issue on solang). This would be nice to have! TBH. this could also work the other way round (e.g. if we could somehow make use of cargo-contract in solang itself).

CC @extraymond

extraymond:
Thanks! I think it'll be mostly useful for use cases including extrinsics and rpcs parts where pallet-contracts are assume, which solang is.

ltfschoen:
> xermicus

@xermicus i was just wondering if you've had time to start on this? i'm eager to try to do it all or help you finish it.
whilst i'm at it i may as well add [Vyper](https://docs.vyperlang.org/) support too

xermicus:
No I currently don't work on this

ltfschoen:
i was just thinking about this, maybe in the CLI there could be options like `--solidity-wasm` and `--solidity-evm` to allow the user to also choose what compiler to use if it detects any .sol files so they may generate either WASM or EVM bytecode, and in the future eWASM like i mentioned here https://substrate.stackexchange.com/questions/8721/how-to-interact-between-a-smart-contract-written-in-solidity-from-a-smart-contra

ltfschoen:
> @ltfschoen I answered regarding cross contracts call directly in the SE question.
> 
> > choose what compiler to use if it detects any .sol files so they may generate either WASM or EVM bytecode
> 
> Right now I don't think this is useful.
> 
> * Cross calls between different runtimes is complex and wouldn't work on most chains
> * If the Solidity contract compiles fine with Solang, I see no reason to use an EVM version instead
> * Many real world contracts require small changes to compile with Solang (which may break compatibility with solc)
> * eWasm is dead in the water and most likely not happening. And even if it would, the contracts would not work just like that on the contracts pallet.
> 
> That said, do you have a specific use case in mind?

[...]

PR #1137, opened 2023-06-01
Relates to #68. Solang Solidity support compiles to WASM using Solang

ltfschoen:
**Proposed changes**
This PR relates to issue https://github.com/paritytech/cargo-contract/issues/68.

User can:
* Run `cargo run -p cargo-contract contract build --solidity-filename flipper.sol --compiler-target="substrate"` to compile the Solidity smart contract to Substrate WASM using Solang

Note: use of solang with "solana" target is not supported, only "substrate". compiling to evm is not supported

Tried to add support for Windows and other OS, but only tested on macOS

**Usage**
* Install Solang https://github.com/hyperledger/solang
  * Note: On macOS it may be necessary to set permissions with 
      ```bash
      sudo chown -R $(whoami) /usr/local/Caskroom/solang && brew install --cask hyperledger/solang/solang
      ```
* Create a flipper.sol file in the project root directory with its contents the same as https://solang.readthedocs.io/en/latest/examples.html#flipper
* Run the following to build a Flipper contract after running `https://github.com/paritytech/cargo-contract && cd cargo-contract`:
```bash
cargo build
```
* Run `cargo run -p cargo-contract contract build --solidity-filename flipper.sol` OR `cargo run -p cargo-contract contract build --solidity-filename flipper.sol` to compile the Solidity smart contract to Substrate WASM using Solang
* View flipper.wasm and flipper.contract outputs
 
* Note: The logs have numerous unnecessary newline characters and i'm returning a dummy `BuildResult` upon success.


xermicus (changes requested):
I'm not very keen about the idea having cargo contract to invoke `solc`. I think the original idea of integrating Solang was that in a further step, interfaces for cross contract calls between the contracts can be automatically generated (you need the metadata for this). Additionally, the Solidity contracts compiled with Solang can then be deployed and called using cargo contract as well.

However, compilation of Solidity contracts for EVM chains seems to be broadly out of scope for cargo contract (a plugin for the _rust package manager_). This feature would be useful regarding only very few parachains anyways (Astar with their XVM comes to mind) and I don't think we should solve their specific problems. Also I have no idea how generating interfaces for those cross runtime calls even would look like. Node interactions with Ethereum-like nodes are not possible neither. 

In order to compile and deploy many real world Solidity contracts, it's not just a matter of invoking `solc`. They might need dependencies, a specific version of the compiler etc. In reality you'd need tooling like hardhat or foundry to manage Solidity contracts, making this a feature close to no use. The same could be said about `solang`, but we are about to implement a feature allowing to read the compiler configuration out of a `Solang.toml`, which should defuse to problem to some extend.

Regarding the compilation with Solang, there's a number of issues with your implementation:
- Not a fan of the bash script since we invoke the solang binary anyways
- `solang compile` has CLI flags too, we need a way to specify them or this won't work
- If cargo contract is invoked with the `--release` flag, this must be passed to Solang
- Making the target configurable makes no sense, it'll always be `substrate`
- No tests

CC @ascjones

ltfschoen:
@xermicus thanks for your feedback that provides clarification. yes i did notice in the CLI target option [here](https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/build.rs#L120) that it used the `enum Target` [here] (https://github.com/paritytech/cargo-contract/blob/master/crates/build/src/args.rs#L174) which is either `Wasm` or `RiscV`, so it was wrong to add an extra `compile_to` option, and thinking about it i'd imagine the team would have already had an `EVM` option if that was necessary. It would have required more [CLI options](https://docs.soliditylang.org/en/v0.8.20/using-the-compiler.html) than i included anyway that would require unnecessary maintenance for a feature that was out of scope. i've also recently dabbled in foundry and mud and understand where you're coming from.

Status:
* [ ] Not a fan of the bash script since we invoke the solang binary anyways
  * 3 Jun '23 - agreed, it doesn't seem to propagate errors. as i understand it i should actually be importing solang into cargo-contract as a Rust dependency and trying to call the `solang` CLI commands from within the cargo-contract` CLI instead.
* [ ] solang compile has CLI flags too, we need a way to specify them or this won't work
  * 3 Jun '23 - agreed, i'll have to add them once i've integrated solang as a dependency into cargo-contract 
* [ ] If cargo contract is invoked with the --release flag, this must be passed to Solang
  * 3 Jun '23 - agreed, i've passed it to solang via bash now, but will need to import solang as a dependency and pass it to solang that way instead
* [ ] The target should not be configurable as it'll always be substrate
  * 3 Jun '23 - agreed, the `solang` target is no longer configurable so it'll always be the default of substrate. i'll have to pass a target value of "substrate" to solang's CLI when it's imported as a dependency
* [ ] No tests
  * 3 Jun '23 - agreed, must be added, maybe the first test once i've integrated solang as a dependency into cargo-contract would be generating debug and release builds and checking that the file sizes are different

ltfschoen:
regarding 1) and 2)
  * 4 Jun '23 - 
    * i've tried importing solang into crates/cargo-contract/Cargo.toml as a rust dependency with `solang = { version = "^0.3.0" }` but that generated this error https://github.com/hyperledger/solang/issues/1351
    * so i forked the solang repo and made this commit to try to fix that dependency conflict and to expose the `shell_complete` function as public https://github.com/hyperledger/solang/compare/main...ltfschoen:solang:cargo-contract?expand=1, and then use that fork branch as a dependency in cargo-contract here https://github.com/paritytech/cargo-contract/commit/a026ee30ed01d80729622a1f713068fb2b3895b5, but unfortunately it gave me the following error on my old macOS machine, but unfortunately it's an old macOS 10.15 and they won't let me update it anymore so i can't run `brew install llvm@15` because my Xcode 11.3.1 is outdated and it needs Xcode 12.4, but the App Store requires macOS 13 or later to update Xcode. but i tried it in a Docker container running Debian on that same macOS machine where i had clang 14 installed already so there were conflicts, and it gave the same error:

```bash
error: No suitable version of LLVM was found system-wide or pointed
              to by LLVM_SYS_150_PREFIX.
       
              Consider using `llvmenv` to compile an appropriate copy of LLVM, and
              refer to the llvm-sys documentation for more information.
       
              llvm-sys: https://crates.io/crates/llvm-sys
              llvmenv: https://crates.io/crates/llvmenv
   --> /Users/me/.cargo/registry/src/github.com-1ecc6299db9ec823/llvm-sys-150.1.0/src/lib.rs:489:1
    |
489 | / std::compile_error!(concat!(
490 | |     "No suitable version of LLVM was found system-wide or pointed
491 | |        to by LLVM_SYS_",
492 | |     env!("CARGO_PKG_VERSION_MAJOR"),
...   |
499 | |        llvmenv: https://crates.io/crates/llvmenv"
500 | | ));
    | |__^

error: could not compile llvm-sys due to previous error
warning: build failed, waiting for other jobs to finish..
```

* so i followed the installation instructions here https://crates.io/crates/llvm-sys and the instructions here https://apt.llvm.org/
```bash
apt install -y lsb-release wget software-properties-common gnupg
bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
apt-get install -y clang-15 lldb-15 lld-15
```
* but that still caused the same error even if i did a full installation, so i made further changes:
```bash
apt-get remove clang-14
apt-get remove -y clang-14 lldb-14 lld-14 libclang-14-dev
whereis ld
echo "deb http://apt.llvm.org/bullseye/ llvm-toolchain-bullseye-15 main" >> /etc/apt/sources.list.d/llvm-toochain-bullseye-15.list;
echo "deb-src http://apt.llvm.org/bullseye/ llvm-toolchain-bullseye-15 main" >> /etc/apt/sources.list.d/llvm-toochain-bullseye-15.list;
cd /usr/bin/ld && rm ld && sudo ln -s /usr/bin/ld.lld-15 ld
apt-get update
apt-get install libstdc++5
echo 'export CMAKE_C_COMPILER="/usr/bin/clang-15"' >> "${HOME}/.bashrc" && \
echo 'export CMAKE_CXX_COMPILER="/usr/bin/clang-15"' >> "${HOME}/.bashrc" && \
echo 'export CMAKE_ASM_COMPILER="/usr/bin/clang-15"' >> "${HOME}/.bashrc" && \
echo 'export CC="/usr/bin/clang-15"' >> "${HOME}/.bashrc" && \
echo 'export CXX="/usr/bin/clang-15"' >> "${HOME}/.bashrc" && \
echo 'export ASM="/usr/bin/clang-15"' >> "${HOME}/.bashrc" && \
echo 'export CMAKE_CXX_STANDARD=15' >> "${HOME}/.bashrc" && \
echo 'export LLVM="/usr/bin/clang-15"' >> "${HOME}/.bashrc" && \
echo 'export LLVM_VERSION_MAJOR=15' >> "${HOME}/.bashrc" && \
echo 'export LLVM_VERSION_MINOR=0' >> "${HOME}/.bashrc" && \
echo 'export LLVM_VERSION_PATCH=0' >> "${HOME}/.bashrc" && \
echo 'export PATH="$PATH:/usr/bin/ld"' >> "${HOME}/.bashrc" && \
\. "${HOME}/.bashrc"
apt-get install -y --no-install-recommends \
        clang-15 lldb-15 lld-15 libclang-15-dev binutils
cd /opt
wget https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.7/llvm-15.0.7.src.tar.xz
tar xJf llvm-15.0.7.src.tar.xz
mkdir -p llvm-15.0.7.src/build
find / -name config-ix.cmake
cmake -D CMAKE_INSTALL_PREFIX=/opt/llvm-15.0.7 -D CMAKE_BUILD_TYPE=Debug -D CMAKE_C_COMPILER=/usr/bin/clang-15 -D CMAKE_CXX_COMPILER=/usr/bin/clang-15 -D CMAKE_ASM_COMPILER=/usr/bin/clang-15 ./llvm-15.0.7.src
```

* i tried manually changing the following with `vim /opt/llvm-15.0.7.src/CMakeLists.txt` to `set(CMAKE_CXX_STANDARD 15...` instead of `set(CMAKE_CXX_STANDARD 14...`
* i'm not sure if `ASM=/usr/bin/clang-15` is correct, possibly needs to be `ASM=clang-15` 
* note: i had clang 14 installed previously
* i used downlad from https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.7
* i changed installation to /opt/llvm-15.0.7 instead of $HOME/llvm-15.0.7
* but still got error below:
```bash
CMake Error at CMakeLists.txt:327 (include):
  include could not find load file:

    GNUInstallPackageDir

CMake Error at cmake/modules/CheckCompilerVersion.cmake:98 (message):
  libstdc++ version must be at least 5.1.
Call Stack (most recent call first):
  cmake/config-ix.cmake:15 (include)
  CMakeLists.txt:776 (include)
```

the next steps were going to be something like
```bash
cmake --build . --target install
cd your/crate/path
LLVM_SYS_150_PREFIX=/opt/llvm-15.0.7 cargo build
clang --version
llvm --version
```
so i'll need to re-attempt this when i get a faster machine as it takes too long on this old dual cpu macOS

xermicus:
Solang requires [a specific LLVM fork](https://solang.readthedocs.io/en/latest/installing.html#step-1-install-the-llvm-libraries) fork, so it can't be used as a dependency for cargo contract. What I meant was that, instead of invoking a script that then invokes solang, rather just directly invoke solang. Sorry if this wasn't very clear.

 Currently, solang as the binary crate is not explicitly intended to be used as a library. [We are aware](https://github.com/hyperledger/solang/milestone/9) but right now we have different priorities.

ascjones:
Thanks for picking this up Luke, but we no longer think having `cargo-contract` invoke `solang` makes a lot of sense. 

The philosophy is now to keep the core of the tool as lean as possible, which at the moment means support for building and interacting with `ink` contracts. 

So we are assuming that `solang` devs will manage and invoke the compiler directly, as is customary with different languages.

ltfschoen:
@xermicus FYI, before @ascjones closed the issue and PR i had an attempt at integrating the `solang` CLI into cargo-contract in the last few commits of this PR [here](https://github.com/ltfschoen/cargo-contract/pull/1/files). i totally agree with @ascjones, but it was a great learning exercise and i learnt a lot!

xermicus:
@ltfschoen Nice, thanks for sharing this and reporting all the issues you found on the way to us! Probably a good place for such compiler integrations are tooling like for example [swanky](https://github.com/AstarNetwork/swanky-cli).

Issue #1065, opened 2023-04-06
Add `cargo contract build --verifiable`


cmichi:
This is a cross-cutting issue. We need to get a grip on verifiable builds once and forever. 

The core issue is that the Rust compiler doesn't give us a deterministic build of a contract. We looked into this in the past and it's just not possible currently. The only way to have proper verifiable contracts on-chain is:

* The deployed contract has been built in a specified environment.
* The verifier rebuilds the contract in this specified environment.

We have the unfortunate situation now that some block explorers have defined their own specified environment. Meaning they require contracts to be build in this specified environment to be verifiable. This is not ideal. Users have to know about this before they deploy and it applies only to that Explorer then. We want to have one common denominator for everyone out there who wants to verify a contract.

Here's the outline of how we'll get there:

- [x] We add a flag `cargo contract build --verifiable`.
- [x] When invoked we spawn a pre-defined ink! Docker image, build the contract in there and write the Dockerhub id into the metadata. 
- [x] The `Dockerfile` should be part of the `cargo-contract` repository and published to Dockerhub.
- [x] The Docker image should contain ink!, Solang and ask! dependencies, so that in principle contracts from those languages can be build in verifiable mode.
- [x] This will create new optional dependencies for `cargo-contract`: Rust docker bindings and docker.
- [x] Create follow-up issue: Have `cargo-contract` warn if `cargo contract deploy` is invoked against a production chain. Likely by having a hard-coded list for the moment.
- [x] Create follow-up issue: Have Contracts UI warn if users want to deploy unverified contracts to a production chain.

SkymanOne:
For the second point, do you mean the image name? Docket Hub ID is basically a username, and I don't see the point of keeping that information in metadata.

SkymanOne:
Closing as completed

Issue #1058, opened 2023-03-31
Post processing of generated Wasm failed with rustc 1.70.0-nightly 


ashutoshvarma:
**Summary**
Building contracts with `rustc 1.70.0-nightly` failed due to error in post processing step due to error in `parity-wasm` while loading the wasm module.

https://substrate.stackexchange.com/questions/7881/error-loading-of-original-wasm-failed

```
 [2/5] Post processing wasm file
ERROR: Loading of original wasm failed

Caused by:
    0: Loading of wasm module at '/home/oreki/Projects/meta-tx/target/ink/registry/wasm32-unknown-unknown/release/registry.wasm' failed
    1: Unknown opcode 192
```

https://github.com/paritytech/cargo-contract/blob/1d28f4ef3863e1f6a438f59890545eb87e1e863a/crates/build/src/lib.rs#L468-L471

**Reproduce**
Update the nightly to `v1.70.0` and build flipper example with `cargo +nightly contract build`

peetzweg:
Just had the same issue. If you are blocked by this, switch to the `stable` toolchain instead of `nightly`.
Afaik `cargo contract` is not aiming to always work with `nightly` and `stable` should be used from now on.

```
rustup default stable
```

ashutoshvarma:
I'm aware of the workarounds,
1. Switch to `stable` rustc
2. If using `openbrush` (using nightly is required for openbrush still), use nightly v1.69.0 or older.

> Afaik cargo contract is not aiming to always work with nightly and stable should be used from now on.

I understand that with ink! v4 `nightly` toolchain is no longer required and `stable` is recommended but `openbrush` which majority of ink! ecosystem is using still depends on nightly, so this issue might help guys who are likely to struggle with newer nightly toolchain.
Also, I don't think it is a cargo-contract specific problem, it it more an issue in `parity-wasm` which although deprecated, is still used inside cargo-contract. That adds more reason to replace parity-wasm with something that is not deprecated (like wasm-opt/binaryen tools).

ganesh1233456:
I am still facing issue even though I change nightly v1.68 or v1.69 or v1.70 it doesn't work.

ashutoshvarma:
@ganesh1233456 try with nightly `rustc 1.67.1 (d5a82bbd2 2023-02-07)`, it worked for me.

blairmunroakusa:
[...]

PR #1148, opened 2023-06-06
Standardised verifiable builds

SkymanOne:
Refers to #1065

This PR aims to bring the workflow to produce verifiable and reproducible builds using `cargo contract --verifiable` and docker container with the pre-defined based on Debian distro.

In detail, the command spawns the container with the binding volume to the contact folder (See image documentation for more details), then executes `cargo contract build --release --output-json <some additional flags> > target/build_result.json` to build the contract in the release mode and save the serialized `BuildResult` output to a file. We need that in order to allow the host machine to access the results and print them in the STDOUT in the requested format.

After the execution, the container and `build_result.json` files are deleted. The host then accesses metadata files, deserializes them back into the rust object, add the `image` name, and writes them back.

It was noted that building an image on different architectures (AMD64 vs ARM64) results in the production of different WASM binaries. Therefore, it was decided to stick with the single AMD64 image for the builds. While it may have an impact on the build times on ARM machines, this is the only way we can guarantee consistency in the reproducible builds. 

The image has been published to: https://hub.docker.com/r/paritytech/contracts-verifiable

Since the docker image is supposed to track the `cargo-contract` use `paritytech/contracts-verifiable:test` tag which can be manually specified with the `--image` flag.

Once the PR is merged, `master` tag tracking master branch will be published.

The follow up PR will be focused on the CI to automate the image release process.

## Tasks

- [x] Successfully spawn and remove containers upon execution
- [x] Persist image digest id in the metadata
- [x] Save output to the host folder
- [x] Respect cargo flags passed to the host
- [x] Allowing to specify docker image to use

## Some extra goals I think worth considering

1. <del>Allowing to specify docker image to use</del>
2. Specify the docker command to be used for the build
3. Recognise path dependant multi-contract projects and mount directories automatically ( by @deuszx )

Points 2 and 3 can be considered as a follow-up issues


crates/build/src/lib.rs
```diff
@@ -927,10 +807,371 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
...
+        steps.increment_current();
+        exec_cargo_dylint(crate_metadata, *verbosity)?;
+        Ok(())
+    } else {
+        steps.set_total_steps(total_steps);
+        Ok(())
+    }
+}
+
+pub fn docker_build(args: ExecuteArgs) -> Result<BuildResult> {
```
ascjones, line 989: I reckon move this into its own `docker.rs` file, this file is getting way too big now.

SkymanOne:
We need to approve the `dockerfile` first to be published on registry before releasing the cargo-contract. Once the image is approved and published, `docker.rs` will just pull it from registry instead of trying to find the local copy.

build-image/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+    apt-get remove -y gnupg && \
+    apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# A contract workdir which should be bind mounted to the container. See README for detailed documentation.
+WORKDIR /contract
+
+# A default entry point for manual execution
+ENTRYPOINT [ "cargo", "contract", "build", "--release" ]
```
deuszx, line 78: I've tried building this Dockerfile on my machine - Mac Mini M2 Pro - and this are my findings:
1. The build took 7732 seconds to finish (that's over 2 hours 😬 ):
```
[+] Building 455.9s (5/7)
[+] Building 456.5s (5/7)
[+] Building 588.1s (5/7)
[+] Building 3495.5s (5/7)
[+] Building 3495.6s (5/7)
[+] Building 3495.8s (5/7)
[+] Building 3495.9s (5/7)
[+] Building 7732.4s (8/8) FINISHED
```
2. The resulting image is over 5GB big:
```
REPOSITORY                            TAG       IMAGE ID       CREATED             SIZE
<none>                                <none>    46ca20145887   21 minutes ago      5.44GB
```
3. They way it's built is not arch-independent (or OS) and when building on Mac I was getting lots of warnings about the allocator, one for every dependency built:
```
=> => #    Compiling is-terminal v0.4.7
 => => # <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
 => => # <jemalloc>: (This is the expected behaviour if you are running under QEMU)
```
4. At the peak, it used almost 23GB of RAM. At the beginning my OS was killing the process when it hit the initial RAM ceiling (16GB).

Consider taking a look at our [Dockerfile](https://github.com/Cardinal-Cryptography/docker-ink-dev/blob/main/Dockerfile) and the accompanying [Makefile](https://github.com/Cardinal-Cryptography/docker-ink-dev/blob/main/Makefile).

For comparison:
1. Build time was 797s:
```
➜  docker-ink-dev git:(main) ✗ make build-image
docker buildx build --pull --platform linux/arm64/v8  -t cardinal-cryptography/ink-dev-arm64:1.1.0 --load . \
	&& docker tag cardinal-cryptography/ink-dev-arm64:1.1.0 cardinal-cryptography/ink-dev-arm64:latest
[+] Building 534.8s (11/16)
[+] Building 545.7s (11/16)
[+] Building 797.5s (17/17) FINISHED
```
2. Image size is slightly over 1GB (this image contains (ink-wrapper)[ink-wrapper/aleph_client.rs at main · Cardinal-Cryptography/ink-wrapper], w/o it's 800MB):
```
gorskimateusz@colima:/Users/gorskimateusz$ docker images
REPOSITORY                            TAG       IMAGE ID       CREATED          SIZE
cardinal-cryptography/ink-dev-arm64   1.1.0     0f3f4874078e   19 seconds ago   1.13GB
cardinal-cryptography/ink-dev-arm64   latest    0f3f4874078e   19 seconds ago   1.13GB
```
3. Makefile instructions detect the underlying architecture and use correct flags when building so no warnings are logged and the process is much faster.
4. At peak, the process consumed ~1.3GB of RAM.

build-image/README.md
```diff
@@ -1,2 +1,58 @@
...
+## Usage
+
+The default entry point is `ENTRYPOINT [ "cargo", "contract", "build", "--release" ]`
+and work directory is set to `/contract`. You need to mount the folder with the contract to the container.
+
+```bash
+docker run -d \
+    --name ink-container \
+    --mount type=bind,source="$(pwd)",target="/contract" \
+    paritytech/contracts-verified
```
deuszx, line 57: Worth mentioning that this won't quite work if you have path-dependencies on other contracts in `Cargo.toml`, since the mounted volume will see only the _current_ contract (dir) and not things that are "next to" (or anywhere else). For example:
```sh
- /foo
- - lib.rs
- - Cargo.toml
- /bar
- - lib.rs
- - Cargo.toml
```
and in `foo/Cargo.toml` you have
```toml
[dependencies]
bar = { path = "../bar", features = ["ink-as-dependency"]
```

In this case, you need to run the verification in folder common to both and use `--manifest-path foo/Cargo.toml`.

SkymanOne: This will work if you point root dir of 2 contracts to `/contract` and specify `--manifest-path <relative/dir>` as you correctly said.

deuszx: Yes, it's just that's not what this README say - that's all I'm saying. The `ENTRYPOINT` for the docker is already defined so if you call it w/o overriding then it won't work. This won't be obvious to people that use the tool since the non-verified version will work correctly w/o any additional args.

build-image/Dockerfile
SkymanOne: Yes, this is because you are using ARM64 base when building an image on ARM host. To guarantee the _absolute_ determinism in builds. We need to use a completely identical architecture for the builds. That's why we are sticking with AMD64 base image. In fact, the user won't be required to build the image. We are gonna publish on docker hub.  Deploying and running the container from the pre-built image takes ~1-2m depending on the complexity of the contract.

Additionally, the image contains solang and ask! dependencies as we also need to provide a deterministic environment for these languages.

I will take a look at the dockerfile again and see how we can see reduce the size.

build-image/README.md
SkymanOne: Good point. I'll update the README to include instructions for multi-contract projects.

build-image/Dockerfile
deuszx: I can imagine that there will have to exist a docker image for rust toolchain + cargo contract combination as each can change the outcome. Will you be publishing docker images for every one?

SkymanOne: we aim to use latest rust toolchain and stable releases of ink! and cargo-contract. The contract's metadata file will contain the tag of the image used to compile it, so that block explorers can pull the image and verify the contract with it.

Whenever there are updates, we are just gonna publish a new tag.

deuszx: But contracts can choose to use their own toolchain and override yours with rust-toolchain file. Are you going to ignore that?

SkymanOne: We don't ignore that. We persist the toolchain and custom compilation flags used in the metadata of the contract.

SkymanOne: I'm confirming with ask! and solang devs, if they require deterministic builds. If not, then we are likely to adopt slim, platform-agnostic rust image that your dockerfile uses.

deuszx: > We don't ignore that. We persist the toolchain and custom compilation flags used in the metadata of the contract.

You wrote earlier that you aim to use latest toolchain, stable release of ink! and stable release of cargo-contract. That's a triple. If you say you will publish a new docker image every time one of the three release new stable version then that's OK - that's basically a _yes_ to my previous question:
> Will you be publishing docker images for every one?

crates/cargo-contract/src/cmd/extrinsics/call.rs
```diff
@@ -108,7 +109,7 @@ impl CallCommand {
 
         let signer = super::pair_signer(self.extrinsic_opts.signer()?);
 
-        async_std::task::block_on(async {
+        Runtime::new().unwrap().block_on(async {
```
ascjones, line 112: Please replace this and other `Runtime::new().unwrap()` occurrences with `?`, using `map_err` where necessary.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        let client = Docker::connect_with_socket_defaults()?;
+
+        let images = client
+            .list_images(Some(ListImagesOptions::<String> {
+                all: true,
+                ..Default::default()
+            }))
+            .await
+            .unwrap();
```
ascjones, line 78: Remove `unwrap`

ascjones (commented):
Some stray unwraps need removing, there might be some more.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,302 @@
...
+        };
+
+        let container_id = match client
+            .create_container(options.clone(), config.clone())
+            .await
+        {
+            Ok(response) => response.id,
+            Err(_) => {
+                // container might exist, so we delete the previous one
+                let _ = client.remove_container("ink-container", None).await;
```
deuszx, line 186: Is this going download the dependencies every time it's re-run? even for the same contract?

Since we're not mounting the local `~/.cargo` folder it will have to download and compile dependencies inside docker. If you kill the previously-run container it will have to do it again each time.

build-image/README.md
deuszx: Ideally, this tool would detect types of dependencies in the contact, recognize where there are path-dependent ones and then mount them automatically - without the need for manual intervention. That way it's completely opaque to the user and the experience is unified.

crates/build/src/lib.rs
```diff
@@ -159,8 +161,10 @@ pub struct BuildResult {
     pub build_artifact: BuildArtifacts,
     /// The verbosity flags.
     pub verbosity: Verbosity,
+    /// Image used for the verifiable build
+    pub image: Option<String>,
```
deuszx, line 165: Where can one specify it? i.e. how can a user override the image used? Imagine a new version of ink released, or rust, I should be able to use it without installing new cargo contract.

SkymanOne: I have not added the command yet because the image has not been published and there is no way to test the functionality until then. Once the image is published. I will add the argument to the command to specify the running image

crates/build/src/docker.rs
SkymanOne: If the container compiled the contract before, and there were no changes in the metadata or code, then it will not rebuild the contract. Otherwise yes, everytime the command is executed, a new container is built and all ink! dependencies are pulled from scratch. I was thinking of reusing the same container for the contract it was made, but then it will bloat the local storage. 

I might enable container reuse by default and then add the flag to remove the container after the build

build-image/README.md
SkymanOne: Can do as a follow-up issue - to detect path dependencies for other contracts and automatically mount the relevant dirs to the contract.

Let's roll out the core functionality for now.

crates/build/src/docker.rs
deuszx: Another option is to start the container in the background (`docker run`) and then send build commands to it (`docker exec`). If no container is running - start new one as daemon, if it's already running: reuse. You can set a TTL on it and the system would kill it after specified time.

SkymanOne: You still need to create a new container for each contract as you cannot change the mounting config. Even having the TTL, I am not sure what are the implications of having a running idle container on the performance of external system that's gonna run the command.

SkymanOne:
Updated image. I used @deuszx dockerfile for inspiration. I managed to install cargo-contract from crates.io and still have a size of ~890Mb. 

As I said earlier, we need to be consistent with the system architecture. I have compiled flipper contract using the ARM version of the image and the X64_86 one. The two produced binaries were different.

We need to approve the dockerfile of the image first, so I can contact CI team to publish it. Then, I will proceed with finishing off this PR.

@ascjones @deuszx

build-image/Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM docker.io/bitnami/minideb:bullseye-amd64 as slimmed-rust
```
xermicus, line 1: Given this is just a builder, is there a specific reason not to use just the official debian image? If yes, it'd be nice to drop a comment

build-image/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+FROM slimmed-rust as cc-builder
+
+# This is important, see https://github.com/rust-lang/docker-rust/issues/85
+ENV RUSTFLAGS="-C target-feature=-crt-static"
+
+RUN apt-get -y update && apt-get -y install gcc g++
+# Install cargo contract
+RUN cargo install cargo-contract
```
xermicus, line 44: My gut feeling is that we might want to parameterize this and use `--locked`

build-image/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+
+# Cleanup after `cargo install`
+RUN rm -rf ${CARGO_HOME}/"registry" ${CARGO_HOME}/"git" /root/.cache/sccache
+
+# apt clean up
+RUN apt-get remove -y gnupg && \
+    apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
```
xermicus, line 53: FYI: Every command that writes unneeded files must be chained into a single `RUN` together with the cleanup commands, otherwise this won't work. Even if this is just a builder image it should be done correctly (those layers have to live somewhere on local storage too).

crates/build/src/docker.rs
deuszx: exit code 137 means that your system killed the container b/c it tried to use more memory that it was allowed to.

build-image/Dockerfile
deuszx: if @SkymanOne took that from the Dockerfile I linked earlier then I should explain - I chose it b/c of active development of the relevant repository and reputation of the organization that's owning it https://github.com/bitnami/minideb

build-image/Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM docker.io/bitnami/minideb:bullseye-amd64 as slimmed-rust
+
+# metadata
+LABEL io.parity.image.vendor="Parity Technologies" \
+    io.parity.image.title="contracts-verifiable" \
+    io.parity.image.documentation="https://github.com/paritytech/scripts/blob/master/\
+    dockerfiles/ink-ci-linux/README.md" \
```
ascjones, line 7:
```suggestion
```
correct link specified below

build-image/Dockerfile
ascjones: Yes, we want to specify the `--version`.

deuszx:
Can you add a CI step somewhere [here](https://github.com/paritytech/cargo-contract/blob/master/.github/workflows/ci.yml#L174) where you build the same contract on multiple OSes and verify that their outcomes match?

build-image/Dockerfile
SkymanOne: Yes. I took your dockerfile and minideb has better security policy and smaller size even than slim version of Debian.

build-image/Dockerfile
xermicus: Sure, for example [here](https://docs.docker.com/engine/reference/builder/#run)

deuszx: I've reread that page and can't find the confirmation for your words -- i.e. it _does_ mention ofc that this creates another layer and _might_ not work if previous layers don't install anything into these dirs but if they did, then this should work.

@SkymanOne , if you want to check for yourself there's a handy tool called `dive` https://github.com/wagoodman/dive where you can check how each layer modifies the container. If what @xermicus wrote is true, then there should be some stuff in `/var/lib/apt/lists/*` dir.

build-image/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+If you have multi-contract project:
```
deuszx, line 32:
```suggestion
For multi-contract projects, like in the example below:
```

build-image/README.md
```diff
@@ -0,0 +1,44 @@
...
+my-app/
+├─ ink-project-a/
+│  ├─ Cargo.toml
+│  ├─ lib.rs
+├─ ink-project-b/
+│  ├─ Cargo.toml
+│  ├─ lib.rs
+├─ rust-toolchain
+```
+Make sure you mount `my-app` directory and then call
```
deuszx, line 43:
```suggestion
Make sure to run the tool in `my-app` directory
```
I don't think the readers of this guide need to know about _mounting_. After all, for the final consumers of `cargo-contract` we don't want them to need to know about docker inside.

SkymanOne: My belief is if people want to run the container manually then they know what they are doing, and, in that case, we should be precise in the terminology, because there are multiple ways to _run the tool in directory_

deuszx: But the command you give below is for `cargo contract` and not docker. Which means these people don't even know they're running container manually.

SkymanOne:
> Can you add a CI step somewhere [here](https://github.com/paritytech/cargo-contract/blob/master/.github/workflows/ci.yml#L174) where you build the same contract on multiple OSes and verify that their outcomes match?

Need to publish a docker image first.

build-image/Dockerfile
SkymanOne: Yes, indeed. `/var/lib/apt/lists` does not get removed. I will take @xermicus advice in 😄

SkymanOne: We have indeed reduced the intermediate layer size. I am going to publish the image now and we can push further releases if necessary

deuszx:
> > Can you add a CI step somewhere [here](https://github.com/paritytech/cargo-contract/blob/master/.github/workflows/ci.yml?rgh-link-date=2023-06-12T10%3A41%3A42Z#L174) where you build the same contract on multiple OSes and verify that their outcomes match?
> 
> Need to publish a docker image first.

Should we discuss the image first, publish it and then finish off this PR? There are some TODOs in the code that are blocked by that.

crates/metadata/src/lib.rs
```diff
@@ -753,6 +768,7 @@ mod tests {
                         "example_name": "increment"
                     }
                 },
+                "image": "parity/ver-build:latest",
```
deuszx, line 771: In `creates/build/README.md` you're using `paritytech/contracts-verified`.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,346 @@
...
+                all: true,
+                ..Default::default()
+            }))
+            .await?;
+        let build_image = images.iter().find(|i| {
+            i.labels.get("io.parity.image.title")
+                == Some(&"contracts-verifiable".to_string())
+        });
+        if build_image.is_none() {
+            return Err(anyhow::anyhow!("No image found"))
```
deuszx, line 102:
```suggestion
            return Err(anyhow::anyhow!("No image found. Expected `contracts-verifiable`"))
```

crates/build/src/docker.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+        let build_image = build_image.unwrap();
```
deuszx, line 104: Propagate error to the caller rather than unwraping.

crates/build/src/docker.rs
SkymanOne: This will be replaced with image pulling from the registry. This is a temporary solution for now

crates/build/src/docker.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        let crate_metadata = CrateMetadata::collect(&manifest_path, target)?;
+        let host_folder = crate_metadata
+            .manifest_path
+            .absolute_directory()?
+            .as_path()
+            .to_owned();
+
+        let file_path = host_folder.join(Path::new("target/build_result.json"));
```
deuszx, line 113: Did you make sure that user will be able to remove this file? ie that it's not created for user with incorrect privileges.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,346 @@
...
+                anyhow::bail!(e);
+            }
+        };
+
+        let mut build_result: BuildResult = serde_json::from_str(&result_contents)
+            .map_err(|_| {
+                anyhow::anyhow!(
+                    "Error parsing output from docker build. The build probably failed!"
+                )
+            })?;
```
deuszx, line 293: There's a shorthand for that [`serde_json::from_reader`](https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html).

crates/build/src/docker.rs
```diff
@@ -0,0 +1,346 @@
...
+                &verbosity,
+            )?;
+        }
+
+        Ok(BuildResult {
+            output_type,
+            verbosity,
+            ..build_result
+        })
+    })
```
deuszx, line 345: This method is pretty large, wdyt about extracting its steps into separate functions like: `construct_cmd`, `create_container`, etc? It would make it much easier to read through and review.

crates/build/src/docker.rs
SkymanOne: https://doc.rust-lang.org/std/fs/struct.Permissions.html all files are created with o+r+w permissions, for the user that executes the command, so I don't think we'll have a problem there

crates/build/src/docker.rs
SkymanOne: Generally speaking. I don't see where I can decompose the function since none of the blocks are resued anywhere in the codebase.

deuszx: In my mind you extract to a function not only for reuse but also to improve readability. Right now, this `docker::build` command is ~300 lines long, fully of error creation, path manipulation and mutable values - it's just difficult to follow and b/c of that review properly.

crates/build/src/docker.rs
deuszx: This is is still here, not sure why it got resolved.

crates/build/src/docker.rs
deuszx: Looks better, thanks.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,386 @@
...
+    let mut cmds = vec![format!(
+            "mkdir -p target && cargo contract build {} --output-json > target/build_result.json",
+            build_args
+        )];
+
+    entrypoint.append(&mut cmds);
+
+    // in order to optimise the container usage
+    // we are hashing the inputted command
+    // in order to reuse containers for different permutations of arguments
```
deuszx, line 235: I think this comment is still wrong.

crates/build/src/docker.rs
SkymanOne: Because this code will be removed once the image is published. We don’t need to search the local image, we just specify the image name and tags and it will be pulled automatically from the registry or local copy used

crates/build/src/docker.rs
```diff
@@ -103,244 +120,267 @@ pub fn docker_build(args: ExecuteArgs) -> Result<BuildResult> {
...
+    let container_id = if container.is_none() {
         client
-            .stop_container(&container_id, Some(StopContainerOptions { t: 20 }))
-            .await?;
+            .create_container(options.clone(), config.clone())
+            .await?
+            .id
+    } else {
+        container
+            .unwrap()
```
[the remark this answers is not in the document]
SkymanOne: Thanks, that's useful hint

crates/build/src/docker.rs
deuszx: There are more places in the code that will need to be changed once you publish. Why not publish first (in a separate PR, add a Dockerfile for reproducible builds) and then come back to this PR and fix all these places?

SkymanOne: The image has already been published. The code has been updated and proper error handling introduced

crates/build/src/metadata.rs
```diff
@@ -199,6 +185,32 @@ pub(crate) fn execute(
     Ok(())
 }
 
+pub fn write_metadata(
```
ascjones, line 188: I think this would be more intuitive as an inherent method on `ContractMetadata`, since we are dealing with an instance of that in both cases.

build-image/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+
+FROM slimmed-rust as ink-dev
+
+COPY --from=cc-builder /usr/local/cargo/bin/cargo-contract /usr/local/bin/cargo-contract
+
+WORKDIR /contract
+
+# default entry point
+ENTRYPOINT ["cargo", "contract", "build", "--release"]
```
ascjones, line 67: So does this mean `cargo contract build --verifiable` runs the release build by default? I understand that it usually only makes sense to do this for release builds.

However might be a bit unintuitive if the `build` command behaves slightly differently with the `--verifiable` command.

SkymanOne: The entrypoint is overwritten in rust code actually. However, the behavior is similar: the build will always be executed in release and I believe we should keep it that way since I don't really know the case when you want to produce the verifiable build in debug mode.

My understanding of `--verifiable` is to be an alternative flag to `--release`

crates/build/src/metadata.rs
SkymanOne: Yes and no. This function present a very specific functionality by **overwriting** the metadata.

crates/cargo-contract/src/cmd/build.rs
```diff
@@ -121,6 +122,13 @@ pub struct BuildCommand {
     /// The maximum number of pages available for a wasm contract to allocate.
     #[clap(long, default_value_t = contract_build::DEFAULT_MAX_MEMORY_PAGES)]
     max_memory_pages: u32,
+    /// Executes the build inside a docker container to produce a verifiable bundle.
+    /// Requires docker daemon running.
+    #[clap(long, short = 'V', default_value_t = false)]
```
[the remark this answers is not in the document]
SkymanOne: then we won't use the short notation

SkymanOne:
As the scope of this PR is already big, it has been decided to move the introduction of the GitHub workflow items to a separate PR.

deuszx:
@SkymanOne  - you mentioned somewhere in earlier comments that when you tried to build the image on Mac you got an error with **exit code 137** - most often it's caused by the system killing the container for using too much memory. See the screenshot attached from building the image on Mac.  


Screenshot 2023-06-19 at 10 32 02

SkymanOne:
> @SkymanOne  - you mentioned somewhere in earlier comments that when you tried to build the image on Mac you got an error with **exit code 137** - most often it's caused by the system killing the container for using too much memory. See the screenshot attached from building the image on Mac. There's a couple of minutes when it uses these 26GB for
> 
> ```
> 
>  => => #     Updating crates.io index
> 
> ```
> 
> Screenshot 2023-06-19 at 10 32 02
> 
> 
> 
> **UPDATE:** peaked at 26.9GB 
> 
> 

I have come across this already. I guess `docker exec` occupies even more because I could run the `cargo contract build` in exec

crates/build/src/docker.rs
ascjones, line 1: Please copy the standard file header into this file. Also some module level docs with at least a brief overview would be nice.

deuszx (commented):
There are places in this PR where I'd like the code to be improved but many of them are inherited from previous work on the tool, so I won't hold the PR b/c of that. 

The issues I've raised, are now resolved.

Last thing - personally I don't feel comfortable merging new feature to `main` branch w/o having proper tests covering it.

build-image/Dockerfile
ascjones: > I don't really know the case when you want to produce the verifiable build in debug mode

I agree, the reason that I brought it up is that my mental model was that the effect of this command was to execute `build` faithfully in the container with the exact same args.

> My understanding of --verifiable is to be an alternative flag to --release

Okay, then it should be a variant on `BuildMode` to avoid confusion. As it is you could still have `--release` and it would produce the same result with and without that. 

A small con to this would be it is mixing `debug` and `release` cargo profiles with a different concept.

crates/cargo-contract/src/cmd/build.rs
```diff
@@ -145,6 +153,15 @@ impl BuildCommand {
...
         };
 
+        let image = if self.verifiable {
+            match &self.image {
+                Some(i) => Some(ImageVariant::Custom(i.clone())),
+                None => Some(ImageVariant::Default),
+            }
+        } else {
+            None
+        };
```
ascjones, line 163: Yes here self.verifiable == false && self.image.is_some() is invalid. In any case there is some duplication of mapping from `Option<ImageVariant>` to `ImageVariant` which only should happen in one place.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,521 @@
...
+                &network,
+                &unstable_flags,
+                &build_artifact,
+                max_memory_pages,
+            )?;
+
+            let image_variant = match image {
+                Some(i) => i,
+                None => ImageVariant::Default,
+            };
```
ascjones, line 124: Duplication here

crates/build/src/metadata.rs
ascjones: Okay then call it `overwrite_metadata`. From an API perspective it is more idiomatic to have operations which operate on an instance to be methods instead of free functions e.g. instead of:

```rust
crate::metadata::write_metadata(
            metadata_artifacts,
            metadata,
            build_steps,
            verbosity,
        )?;
```

have

```rust
metadata.write(
            metadata_artifacts,
            build_steps,
            verbosity,
        )?;
```

SkymanOne: This method is simply an extract of a snippet from `crate::medata::execute` and contains some printing log operations unsuitable for `metadata.write` idiomatically IMHO.

ascjones: It is a matter of opinion, I think operating on an instance where appropriate for code sharing is more readable and discoverable. But up to you in the end.

crates/cargo-contract/src/cmd/build.rs
ascjones: The duplication is of the match statement, conversion could/should happen all in one place `Option<String>` -> `ImageVariant`. At the moment we match `None` -> `None`, then the `None` gets translated to a `ImageVariant::Default`.

ascjones (commented):
We may need to add some CI tests for this on different platforms, hope that the GH runners can handle it. I haven't got it to work yet on my Ubuntu 22.0 installation. Had to first do `sudo usermod -aG docker $USER` to avoid the `sudo`. But still getting: 

image

crates/build/src/docker.rs
```diff
@@ -0,0 +1,512 @@
...
+}
+
+/// Reads the `BuildResult` produced by the docker execution
+fn read_build_result(host_folder: &Path, file_path: &PathBuf) -> Result<BuildResult> {
+    let file = std::fs::File::open(file_path)?;
+    let mut build_result: BuildResult =
+        match serde_json::from_reader(BufReader::new(file)) {
+            Ok(result) => result,
+            Err(_) => {
+                std::fs::remove_file(file_path)?;
```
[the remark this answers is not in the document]
SkymanOne: We can't deserialise file only when it is empty and that happens only when build fails

build-image/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+ARG GCC_VERSION
+ARG G_VERSION
+
+# This is important, see https://github.com/rust-lang/docker-rust/issues/85
+ENV RUSTFLAGS="-C target-feature=-crt-static"
+
+    # Install required packages for `cargo-contract`
+RUN apt-get -y update && apt-get -y install gcc=${GCC_VERSION} g++=${G_VERSION} \
+    # Install cargo contract
+    && cargo install cargo-contract --locked --version $CARGO_CONTRACT_VERSION \
```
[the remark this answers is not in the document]
SkymanOne: Sure, I'll adjust the file

crates/build/src/docker.rs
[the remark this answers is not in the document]
SkymanOne: If the docker container fails to run, then the error is thrown at `docker.start_container(...)` which will propagate to the earlier error handling context.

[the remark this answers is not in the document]
ascjones: I think then that `run_build` should not return `Ok(())` when the build fails if possible? Or at least not call into `read_build_result`.

`The build probably failed!`

Ideally we would know whether the build failed for sure.

build-image/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+# gcc package version
+ARG GCC_VERSION=4:10.2.1-1
+# wget package version
+ARG WGET_VERSION=1.21-1+deb11u1
+# g++ package version
+ARG G_VERSION=4:10.2.1-1
+# Flag to indicate whether to install cargo-contract from git
+ARG USE_GIT=false
+# Url to the cargo-contract repository to install from
+ARG CARGO_CONTRACT_GIT
```
ascjones, line 18: Can we make these just a single var here? i.e. just `CARGO_CONTRACT_GIT` and check whether it has a value or not?

Also I see they are not set anywhere yet, I guess eventually we should be able to control whether it uses a `crates.io` release or the git dependency.

crates/build/src/docker.rs
```diff
@@ -253,6 +246,7 @@ fn update_metadata(
 async fn run_build(
     build_args: String,
     build_image: &ImageSummary,
+    file_path: &PathBuf,
```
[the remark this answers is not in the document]
SkymanOne: Good spot

SkymanOne:
@deuszx Docker Engine allows enabling Rosetta simulation for X86_64 images. After enabling it, the image build time was reduced to ~500s, and the build inside the container was <1min.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,535 @@
...
+    };
+    if verbosity.is_verbose() {
+        let spinner_style =
+            ProgressStyle::with_template(" {spinner:.cyan.bold} {wide_msg}")?
+                .tick_chars("⠁⠂⠄⡀⢀⠠⠐⠈ ");
+
+        let pb = ProgressBar::new(1000);
+        pb.enable_steady_tick(Duration::from_millis(100));
+        pb.set_style(spinner_style);
+        pb.set_message("Build is being executed...");
```
ascjones, line 386: I think it would be much better to capture the output of the build via the logs and pipe it to the console, in order that the build looks similar to a "local" build. Also will make it easier to troubleshoot what is going on inside the container.

SkymanOne: There is no output as the docker build is carried out with `--output-json` flag

ascjones: An idea would be to change all the progress output to `stderr` (as cargo does) and then we can still show output even for `--output-json`. Anyway this is a nice to have.

crates/build/src/lib.rs
```diff
@@ -808,6 +815,37 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
     })
 }
 
+/// Modifies permissions of all entries in the dir to be read_only = false
+#[allow(clippy::permissions_set_readonly_false)]
+fn change_permissions(p: &PathBuf) -> Result<()> {
```
ascjones, line 820: Maybe try the approach of running the container as the current user e.g. https://unix.stackexchange.com/a/627028

deuszx: I thought you had it figured out @SkymanOne (you said so) but it seems like it's still unfinished.

Give a try to [this approach](https://github.com/Cardinal-Cryptography/ink-wrapper/blob/main/Makefile#L77) - it should run the docker container as the current user.

SkymanOne: Wouldn't work on windows

SkymanOne: > I thought you had it figured out @SkymanOne but it seems like it's still causing some troubles. Not surprising, we all were there.
> 
> Give a try to [this approach](https://github.com/Cardinal-Cryptography/ink-wrapper/blob/main/Makefile#L77) - it should run the docker container as the current user.

The problem with that is that it requires rebuilding an image for every newly spawned container. While some intermediate layers may be cached, there is still a potential issue with determinism. Additionally, as I mentioned above, this method will not work on Windows machines.

So, the only solution I see currently is to set permission's read_only flag to false which would give access to the `target` folder to all users

ascjones: > The problem with that is that it requires rebuilding an image for every newly spawned container.

I'm displaying my lack of docker knowledge here, are you saying that specifying the user causes the image to rebuilt each time? I wasn't able to see a difference on this between setting the `--user` and leaving it out and defaulting to `root`

I've played with it a bit and was able to get this to work just using the raw docker commands:

image
image

> this method will not work on Windows machines

There must be another way though to control the user in windows? In any case the current solution of `@SkymanOne
[give permission to target to everyone](https://github.com/paritytech/cargo-contract/pull/1148/commits/217ab6a49338e35ba7ba19b4fb4ca6cb970db10c)` does not work on Linux for me at the moment either.

There must be a way we can get this working on all OSs.

SkymanOne: Managed to fix the file permission issue on linux. Now just need to figure out how to get the conditional compilation code working

ascjones: Permissions issue fixed now :+1:

crates/build/src/docker.rs
```diff
@@ -0,0 +1,563 @@
...
+        .await?;
+
+    maybe_println!(
+        verbosity,
+        " {} {}\n {}",
+        format!("{build_steps}").bold(),
+        "Started the build inside the container"
+            .bright_green()
+            .bold(),
+        "You can close this terminal session. The execution will be finished in the background"
```
ascjones, line 375: I find this confusing when it prints this, even though it is true. Why would we tell a user to do this, especially when we are still waiting for a success or failure of the build?

In my mind we should be attaching to the docker run anyway, so we can see any errors produced by the build. e.g. if there is a compilation error in the contract we just get:

image

crates/build/src/docker.rs
```diff
@@ -0,0 +1,563 @@
...
+    entrypoint.append(&mut cmds);
+
+    // in order to optimise the container usage
+    // we are hashing the inputted command
+    // in order to reuse the container for the same permutation of arguments
+    let mut s = DefaultHasher::new();
+    entrypoint.hash(&mut s);
+    let digest = s.finish();
+    // taking the first 5 digits to be a unique identifier
+    let digest_code: String = digest.to_string().chars().take(5).collect();
```
ascjones, line 289: This was causing some issues for me because if there already exists a container with the id derived from the entrypoint from a *different* image, it will just use that so the image is effectively ignored and it will just use that container.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,563 @@
...
+            let args = compose_build_args()?;
+
+            let client = Docker::connect_with_socket_defaults().map_err(|e| {
+                anyhow::anyhow!("{}\nDo you have the docker engine installed in path?", e)
+            })?;
+            let _ = client.ping().await.map_err(|e| {
+                anyhow::anyhow!("{}\nIs your docker engine up and running?", e)
+            })?;
+            let build_image =
+                get_image(client.clone(), image, &verbosity, &mut build_steps).await?;
```
ascjones, line 149: I don't quite understand the purpose of this step of manually fetching the image. Using raw docker commands e.g. `docker run` or `docker create`, you can just supply the image name and docker will take care of pulling it to the local machine?

crates/build/src/docker.rs
SkymanOne: Good spot. I will include the image name in the calculation of the digest.

crates/build/src/docker.rs
SkymanOne: * First, I am working with the docker API, so some of the operations need to be explicit. So, the API still requires us to "create" an image: https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.create_image 
* Second, if we just specify the image upon creating the container, then when inspecting the container, it does not show the image name, only the digest, i.e.:
```
...
"Image": "sha256:022e29e85a19c24a555744d2c44fbc95a15d3b70df5bb5d6fc04c0c79b182698",
...
```
* Lastly, since we need to explicitly "create" an image. It allows us to show a progress bar for the user of the download. If the image is large, then we want to show the user how much time it takes. The idea behind verifiable builds is that we assume that the user does not have any experience with the docker and image/container management. Therefore, providing this information is just a nice thing to have.

crates/build/src/docker.rs
SkymanOne: Yes, I will attach the error message if there is any. However, since the docker container runs the build with the `--output-json` there are any build logs produced other than errors

SkymanOne:
```
I find this confusing when it prints this, even though it is true. Why would we tell a user to do this, especially when we are still waiting for a success or failure of the build?
```
Because this is the behaviour, the container will continue to run if the user quits the session with `ctrl + c`. So, I am just giving the user information.

ascjones: I, as a user, found it confusing. Even though as you say it is technically correct.

Also, if the user were to close the terminal session as suggested, then the build result would not be rewritten and the resulting `.bundle` file would be incorrect?

Anyway, let's attach to the docker process and keep it as similar as possible to a regular build.

SkymanOne: Actually, the bundle will be correct since it doesn't contain any paths. All the build result shows is the target dirs of the artefacts and the code size. If we exit terminal early, the build result will just not be printed in the terminal. This is equivalent of just running the build in the background

ascjones: As I read the, once the build has succeeded it rewrites some paths and updates the image used in the metadata https://github.com/paritytech/cargo-contract/blob/5ec303bdc62e652dd84be3d4475d6b97db2781b7/crates/build/src/docker.rs#L162C29-L162C29.

So if you quit the terminal it will never do that step. True that the contract files will still be there but isn't the main reason one runs `--verifiable` to be able to compare the output of builds, and to know which image was used.

SkymanOne: All we can do in this case is not to display the "you can close the terminal" message

crates/build/src/docker.rs
ascjones: > First, I am working with the docker API, so some of the operations need to be explicit. So, the API still requires us to "create" an image: https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.create_image

Ah okay so the `docker` client has the logic which you are replicating here. 

I did test without explicitly fetching the image but it turns out I still had a local image from a previous run. I was missing the `--all` from `docker image prune` to clear all images.

> Second, if we just specify the image upon creating the container, then when inspecting the container, it does not show the image name, only the digest, i.e.:

I was able to get this to work, but I guess it is moot now:

image

crates/build/src/docker.rs
ascjones: Let's remove the message then :+1:

sourabhniyogi:
This is great!  We will use it in Polkaholic -- can you make the Docker image public 
```
# docker pull paritytech/contracts-verified
Using default tag: latest
Error response from daemon: pull access denied for paritytech/contracts-verified, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

crates/build/src/docker.rs
ascjones: Dumb question perhaps: why do the hashing at all and not just use the concatenated image + contract name, does it have to be a unique for different argument permutations? What would happen if they just use the same container?

crates/build/src/docker.rs
```diff
@@ -0,0 +1,607 @@
...
+                            let msg =
+                                msg.split('\r').filter(|m| !rex.is_match(m)).collect();
+                            Some(msg)
+                        } else {
+                            None
+                        }
+                    })
+                    .collect::<Vec<String>>()
+                    .join("");
+                anyhow::bail!("{}\n{}", e, err_string);
```
ascjones, line 438: This doesn't appear to work correctly, but at least we can see the error now which is better than before:

image

crates/build/src/docker.rs
```diff
@@ -0,0 +1,607 @@
...
+    let start_time = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .context("Invalid start time")?
+        .as_secs() as i64;
+
+    maybe_println!(
+        verbosity,
+        " {} {}",
+        format!("{build_steps}").bold(),
+        "Started the build inside the container"
```
ascjones, line 371: It might be useful to print out the container id here, it would be useful for e.g. attaching to it. Or could add it as an `info!` at the very least.

crates/build/src/docker.rs
SkymanOne: You can not use the same container, because upon creation, the entry point command becomes part of the container. So if you need to run a build with different set of args, you need to create a new container. So the idea is that we spawn a new container for a new permutation of args, and then reuse it. The reason for that was discussed before, because we cannot use `docker exec` since there is a limitation in RAM when running on Apple Silicon because of emulation overhead.

crates/build/src/docker.rs
```diff
@@ -0,0 +1,607 @@
...
+    build_image: &ImageSummary,
+    build_result_path: &PathBuf,
+    contract_name: &str,
+    host_folder: &Path,
+    verbosity: &Verbosity,
+    build_steps: &mut BuildSteps,
+) -> Result<()> {
+    let client = Docker::connect_with_socket_defaults()?;
+
+    let mut entrypoint = vec!["/bin/bash".to_string(), "-c".to_string()];
```
ascjones, line 271: Why do we need the `/bin/bash` and can't just execute `cargo contract build` as the entrypoint directly?

crates/build/src/docker.rs
```diff
@@ -0,0 +1,607 @@
...
+    let b_path_str = build_result_path
+        .as_os_str()
+        .to_str()
+        .context("Cannot convert Os String to String")?;
+    let mut cmds = vec![format!(
+        "mkdir -p target && cargo contract build {} --output-json > {}",
```
ascjones, line 277: Is the `mkdir -p target` strictly required?

crates/build/src/docker.rs
deuszx: > there is a limitation in RAM when running on Apple Silicon because of emulation overhead.

I haven't experience that problem. Can you post steps to reproduce @SkymanOne ?

crates/build/src/docker.rs
ascjones, line 277: Why do all other arguments get added as individual array entries and this one is its own long string? I guess it would be ugly with `vec!["cargo".to_string(), "contract".to_string()]` (like you already do above with `vec!["/bin/bash".to_string(), "-c".to_string()];`)

crates/build/src/docker.rs
SkymanOne: See: https://docs.docker.com/engine/reference/builder/#entrypoint

specifically: `ENTRYPOINT ["executable", "param1", "param2"]`

crates/build/src/docker.rs
SkymanOne: Ah, I need manually filter out logs, because everything is dumped into `CONSOLE` output

crates/build/src/docker.rs
SkymanOne: > > there is a limitation in RAM when running on Apple Silicon because of emulation overhead.
> 
> I haven't experience that problem. Can you post steps to reproduce @SkymanOne ?

See previous discussion: https://github.com/paritytech/cargo-contract/pull/1148#discussion_r1223379844

crates/build/src/docker.rs
SkymanOne: Yes, because that directory may not exist if the contract has never been built before, and ` ... > target/build_result.json` will error with `no directory exists`. So we optionally create the directory if it doesn't exist `that's what `-p` arg is for.

SkymanOne: The reason is because `>` and `&&` can't be represented as individual elements in the vec, so that's why I need to concat them together

crates/build/src/docker.rs
SkymanOne: Should be fixed by the recent commit

altaua:
> This is great! We will use it in Polkaholic -- can you make the Docker image public
> 
> ```
> # docker pull paritytech/contracts-verified
> Using default tag: latest
> Error response from daemon: pull access denied for paritytech/contracts-verified, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
> ```

The image is public, you just seem to have mistyped its name. It's "verifiable", not "verified": https://hub.docker.com/r/paritytech/contracts-verifiable :)

sleepyyj:
Does anyone know about this error?
image

deuszx:
> Does anyone know about this error? image

Try putting `--manifest-path` after `contract build`. It's an argument of `cargo-contract` and not `cargo` itself.

sleepyyj:
@deuszx Thanks. But a new error occurred.
image

ascjones (commented):
For me the `docker.rs` implementation feels convoluted. I realise that is not very actionable feedback, so I have had a play around and have identified the areas I can think can be simplified.

1. Sending the `build_result.json` via a file on the mounted volume should be replaced by consuming `stdout` directly by attaching to the container after starting it. 
2. Print all cargo contract status to `stderr` to enable the above, then we can pipe the attached `stderr` directly to our console and also allow the build progress to be shown, which allows removing of the status bar code for the build. 
3. Remove container digest and reusing container code: replace creating a new container for each build and removing when complete (like with `docker run --rm ,,,`). There would be a cost of fetching `crates.io` registry for a fresh container, but this can be solved by mounting a shared volume for `$CARGO_HOME` to persist the cache across builds.
4. Possibly remove the code which rewrites the paths of the `BuildResult`, it is a bit brittle. One alternative would be to use relative paths to the `Cargo.toml` there instead.
5. Remove the checking for whether a local image exists, just call `pull_image` on the image name. It will do that on its own, pull if it does not exist.
6. Possibly use the "built in" progress bar for showing image pull status, available via `summary.progress`. This will remove a bunch of code and give more of a "native" docker experience.

sourabhniyogi:
> > This is great! We will use it in Polkaholic -- can you make the Docker image public
> > ```
> > # docker pull paritytech/contracts-verified
> > Using default tag: latest
> > Error response from daemon: pull access denied for paritytech/contracts-verified, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
> > ```
> 
> The image is public, you just seem to have mistyped its name. It's "verifiable", not "verified": https://hub.docker.com/r/paritytech/contracts-verifiable :)

Thank you, not sure how I ended up there!   
```
# docker pull paritytech/contracts-verifiable
Using default tag: latest
latest: Pulling from paritytech/contracts-verifiable
f3cb9c152eac: Pull complete 
9f517e294b33: Pull complete 
22acd62a0369: Pull complete 
6ed49230c42f: Pull complete 
Digest: sha256:5ad8f4406e98486cfafeb220c7b8c7f0a5d3c5d1e963f045bb549441116a5746
Status: Downloaded newer image for paritytech/contracts-verifiable:latest
docker.io/paritytech/contracts-verifiable:latest
```

crates/build/src/docker.rs
```diff
@@ -0,0 +1,552 @@
...
+            filters,
+            ..Default::default()
+        }))
+        .await?;
+
+    let container_option = containers.first();
+
+    if container_option.is_some() {
+        return Ok(container_name)
+    }
```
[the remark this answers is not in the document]
SkymanOne: Yep. Next iteration. As long as we don't make breaking changes to the image, we should be fine

crates/build/src/docker.rs
```diff
@@ -0,0 +1,552 @@
...
+
+    let mut cmd = vec![
+        "build".to_string(),
+        "--release".to_string(),
+        "--output-json".to_string(),
+    ];
+
+    cmd.append(&mut build_args);
+
+    let digest_code = container_digest(entrypoint.clone(), build_image.to_string());
```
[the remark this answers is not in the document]
SkymanOne: Right. Good spot

crates/build/src/docker.rs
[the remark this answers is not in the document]
SkymanOne: The mounting thing will be the next feature, I plan also to detect path dependencies in `Cargo.toml`, manifest paths and mount respective dirs. So, we can also add the mounting of the index as part of this feature

sergejparity (approved):

SkymanOne:
The branch has been merged to master. Please use `paritytech/contracts-verifiable:master` for now until the crate is published.

sourabhniyogi:
Tested out `paritytech/contracts-verifiable:master` but must be doing it wrong:

```
# docker run -d --name ink-container --mount type=bind,source=".",target="/contract"  paritytech/contracts-verifiable:master
137b6f0a3aaa1de3d2ceb55314347b3d9596b39ceb3cef8ea889fa1bd807b4b6

# docker logs 137b6f0a3aaa1de3d2ceb55314347b3d9596b39ceb3cef8ea889fa1bd807b4b6
Utilities to develop Wasm smart contracts

Usage: cargo contract <COMMAND>

Commands:
  new          Setup and create a new smart contract project
  build        Compiles the contract, generates metadata, bundles both together in a `<name>.contract` file
  check        Check that the code builds as Wasm; does not output any `<name>.contract` artifact to the `target/` directory
  upload       Upload contract code
  instantiate  Instantiate a contract
  call         Call a contract
  encode       Encodes a contracts input calls and their arguments
  decode       Decodes a contracts input or output data (supplied in hex-encoding)
  remove       Remove contract code
  info         Display information about a contract
  help         Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
 ```
 Can you advise?

**Comment by SkymanOne** [comment]
> Tested out `paritytech/contracts-verifiable:master` but must be doing it wrong:
> 
> ```
> # docker run -d --name ink-container --mount type=bind,source=".",target="/contract"  paritytech/contracts-verifiable:master
> 137b6f0a3aaa1de3d2ceb55314347b3d9596b39ceb3cef8ea889fa1bd807b4b6
> 
> # docker logs 137b6f0a3aaa1de3d2ceb55314347b3d9596b39ceb3cef8ea889fa1bd807b4b6
> Utilities to develop Wasm smart contracts
> 
> Usage: cargo contract <COMMAND>
> 
> Commands:
>   new          Setup and create a new smart contract project
>   build        Compiles the contract, generates metadata, bundles both together in a `<name>.contract` file
>   check        Check that the code builds as Wasm; does not output any `<name>.contract` artifact to the `target/` directory
>   upload       Upload contract code
>   instantiate  Instantiate a contract
>   call         Call a contract
>   encode       Encodes a contracts input calls and their arguments
>   decode       Decodes a contracts input or output data (supplied in hex-encoding)
>   remove       Remove contract code
>   info         Display information about a contract
>   help         Print this message or the help of the given subcommand(s)
> 
> Options:
>   -h, --help     Print help
>   -V, --version  Print version
> ```
> 
> Can you advise?

The entry point now is `cargo contract`, you need to specify the command to run
https://github.com/paritytech/cargo-contract/blob/db242993af92a9860fd753207cf9583e5a95a23e/build-image/Dockerfile#L111

**Comment by sourabhniyogi** [comment]
I tried it out on 4 cases (where all 4 cases were generated from ChainIDE WASM Contract templates and compiled in ChainIDE), and flipper compiled but the other 3 did not.  Errors are included below:

1. [flipper](https://cdn.polkaholic.io/wasmcode/0x396418ff533172de8407004f53051b5409f6892564ddeba12f75cc855cb16fe0/flipper.zip) 

2. [my_psp22](https://cdn.polkaholic.io/wasmcode/0x5827c0552e527887b01ba78a1d854811f157dbc02353d9300c61462e9c06961d/package.zip)

```
# docker run -d --name ink-container --mount type=bind,source="./src",target="/contract" paritytech/contracts-verifiable:master build
# docker logs [...]
 [1/*] Building cargo project
error: failed to parse manifest at `/tmp/cargo-contract_RqgRr5/Cargo.toml`

Caused by:
  the target `my_psp22` is a binary and can't have any crate-types set (currentlERROR: command ["/usr/local/rustup/toolchains/1.69-x86_64-unknown-linux-gnu/bin/cargo", "build", "--color=always", "--target=wasm32-unknown-unknown", "-Zbuild-std=core,alloc", "--no-default-features", "--release", "--target-dir=/contract/target/ink", "--features", "ink/ink-debug"] exited with code 101
```

3. [shiden34](https://cdn.polkaholic.io/wasmcode/0xb33cafa42e6a8980d60dc20a32c307c610fa9e5a9d9992c6f8b46cab4c45944d/package.zip)

```
# docker run -d --name ink-container --mount type=bind,source="./src",target="/contract" paritytech/contracts-verifiable:master build
# docker logs
Some errors have detailed explanations: E0405, E0408, E0412, E0425, E0433, E0463, E0531.
For more information about an error, try `rustc --explain E0405`.
error: could not compile `proc-macro2` due to 451 prevERROR: command ["/usr/local/rustup/toolchains/1.69-x86_64-unknown-linux-gnu/bin/cargo", "build", "--color=always", "--target=wasm32-unknown-unknown", "-Zbuild-std=core,alloc", "--no-default-features", "--release", "--target-dir=/contract/target/ink", "--features", "ink/ink-debug"] exited with code 101
```

4. [factory_contract](https://cdn.polkaholic.io/wasmcode/0x86649b2d9fae6c57bf1c4aedacd4bbae324058cc5a2a0e96e9eb6c633a4d4bf6/package.zip)

```
# docker run -d --name ink-container --mount type=bind,source="./src",target="/contract" paritytech/contracts-verifiable:master build
# docker logs 605caabaf785400fb4d13750629ed92c22531e355f98c646d4f27a7ed3593042
ERROR: Error invoking `cargo metadata` for Cargo.toml

Caused by:
    `cargo metadata` exited with an error:     Updating crates.io index
        Updating git repository `https://github.com/727-Ventures/openbrush-contracts`
    error: failed to get `pair_contract` as a dependency of package `factory_contract v0.1.0 (/contract)`
    
    Caused by:
      failed to load source for dependency `pair_contract`
    
    Caused by:
      Unable to update /pair
    
    Caused by:
      failed to read `/pair/Cargo.toml`
    
    Caused by:
      No such file or directory (os error 2)
```      

We were happy that case 1 generated the same codehash as chainide but any insights to cases 2/3/4 would be appreciated!

SkymanOne:
> docker run -d --name ink-container --mount type=bind,source="./src",target="/contract" paritytech/contracts-verifiable:master build

I compiled the given contracts locally (in host environment), and got the same errors. The issue is unrelated to this PR. Please open respective issues in 727-Ventures/openbrush-contracts repo.

wufengtao1:
> > docker run -d --name ink-container --mount type=bind,source="./src",target="/contract" paritytech/contracts-verifiable:master build
> 
> I compiled the given contracts locally (in host environment), and got the same errors. The issue is unrelated to this PR. Please open respective issues in 727-Ventures/openbrush-contracts repo.

Hi. I am from Chainide. Based on my experience, the last three contracts require +nightly and rustc 1.67.1 to compile successfully (you can compile them successfully at https://staging-9589904a8a.chainide.com/s/dashboard/projects). The reason for this might be related to this: https://github.com/paritytech/cargo-contract/issues/1058. So, I believe either Openbrush should update their template or the Docker should support "cargo +nightly-2023-02-07 contract build". Either way, I think it requires communication among Openbrush, Paritytech, Chainide, and Polkaholic.

SkymanOne:
@wufengtao1 @sourabhniyogi can you please open a separate issue in cargo-contract and ping the relevant stakeholders? The new version of the cargo-contract has not yet been released. So, we still have some time to tweak the image to accommodate all the stakeholders that require this feature.

wufengtao1:
@SkymanOne Sure. Let me handle this.

PR #1162, opened 2023-06-14
Suggest valid message or constructor name, when misspelled

pgherveou:
Suggest a valid message or constructor name when the user misspell it. 
If there are no close match, it will display the full list of available messages / constructors. 
This can be helpful when messages are namespaced and the user input it with the correct case

Screenshot 2023-06-14 at 21 10 44


SkymanOne (approved):
Thanks!

PR #1181, opened 2023-06-27
feat: expose extrinsics operations as a library

### LINKED ISSUE #1140: [client library] decouple extrinsics related ability as library
## Info

**background**:
This is a proposal similar to #659, this time targeted extrinsics.

Since now we have both contract-transcode and contract-metadata as isolated libraries, any smart contract that is compatible with the metadata version can:
- use it to load the contract bundle
- use the transcoder to encode, decode the payload and event output.

Currently for live testing contracts against pallet-contract, most likely one will end up setup subxt and use the contract-metadata, contract-transcode to test it's logic and input and output. 
Operation around uploading/deploying/calling/reading is regularly involved when testing against live environment. If we can extract those extrinsics ops out, it might make writing testing framework or other alternative clients more easily.

**proposal**:
Expose most of `crates/cargo-contract/src/cmd/extrinsics` as library, so it can be reused or embedded in other projects.

https://github.com/hyperledger/solang/pull/990 is an in progress PR of creating tests against solang compiled solidity contracts using subxt, as you can see [here](https://github.com/hyperledger/solang/pull/990/files#diff-a1604e4277df2f527d3ec2b26f2012db902400d585b5e5c0b3c9b6279df976dd
), most of the setup code is achieving very similar goals like cargo-contract does in the `cmd` module.

The end result of this proposal will definitely let solang stay close to the testing environment cargo-contract currently have to iron up rough edges, and perhaps open up other limitation not needed by ink contracts but important to other smart contract runtime running on top of pallet-contracts. 
For example contract-metadata currently don't support I256, U256 type which solang supports, which the contract-transcode will refuse to encode input for solang if those types are summoned.

**effected use cases**:
the developer experience of the following use cases will most likely be improved due to less hand-rolled setup code.
- testing contract logic and state changes against a running chain
- testing against chain built-ins or extensions that through contract output
- incorporated into clients to have preliminary ability to upload/call/read contract, e.g solang CLI.

C.C @xermicus

tareknaser360:
## Description

This pull request introduces a library that exposes extrinsics operations as a reusable component. This makes it easier to integrate and reuse these operations in various projects.

Fixes #1140

## Notes to Reviewers:
- Most of the changes in this PR involve reorganizing existing code. No new code has been added.
- The `runtime_api` module has been moved to the new crate to avoid cyclic dependencies.
- The types `CodeHash`, `Balance`, and  `Client` have been defined in the new crate for the same reason.
- The documentation for the extrinsics has been moved to the readme of the crate itself.


crates/extrinsics/src/lib.rs
```diff
@@ -20,6 +20,7 @@ mod error;
 mod events;
 mod instantiate;
 mod remove;
+pub mod runtime_api;
```
ascjones, line 23: I'm not sure about exposing this entire module, there's a lot in there and it can easily cause breaking changes any time we update the metadata. We could

1. Move the code which fetches the `ContractInfo` into the `extrinsics` crate and export that.
2. Only export `ContractInfo` and the function used to fetch it.

What do you think?

crates/extrinsics/LICENSE
ascjones, line 1: This should be symbolic link to `../../LICENSE`

crates/extrinsics/src/call.rs
```diff
@@ -16,33 +16,27 @@
...
-            display_dry_run_result_warning,
-            events::DisplayEvents,
-            ErrorVariant,
-        },
-        runtime_api::api,
-        Balance,
-    },
-    DEFAULT_KEY_COL_WIDTH,
-};
+use crate::DEFAULT_KEY_COL_WIDTH;
```
ascjones, line 39: Could be moved inside `super` as with others.

crates/extrinsics/src/instantiate.rs
```diff
@@ -16,33 +16,27 @@
...
-            ErrorVariant,
-            TokenMetadata,
-        },
-        runtime_api::api,
-        Balance,
-        CodeHash,
-    },
-    DEFAULT_KEY_COL_WIDTH,
-};
+use crate::DEFAULT_KEY_COL_WIDTH;
```
ascjones, line 39: Same here, could go inside the `super`.

crates/extrinsics/src/lib.rs
tareknaser360: `fetch_contract_info` takes `InfoCommand` as a parameter. 
The `extrinsics` crate doesn't have access to it.
As far as I am concerned, we can't use it without having cyclic dependencies.

Is there another way around this?

ascjones: It only uses `&self.contract` from `InfoCommand` so could just pass that as a parameter instead.

For the return type could even hide the `ContractInfo` type and return `InfoToJson` so we control the API instead of the codegen.

tareknaser360: > It only uses &self.contract from InfoCommand so could just pass that as a parameter instead.

Yes that's fine

but for the return type. `InfoToJson` is defined in `cargo_contract` crate so it can't be the return type of the new function in the `extrinsics` crate.
and even if we have it, how can we cast `ContractInfo` to `InfoToJson`?

ascjones: Move `InfoToJson` to `extrinsics`, maybe just rename it to `ContractInfo`.

> and even if we have it, how can we cast ContractInfo to InfoToJson?

Just build the instance as we already do:

```rust
                    let convert_trie_id = hex::encode(info_result.trie_id.0);
                    let info_to_json = InfoToJson {
                        trie_id: convert_trie_id,
                        code_hash: info_result.code_hash,
                        storage_items: info_result.storage_items,
                        storage_item_deposit: info_result.storage_item_deposit,
                    };
```

tareknaser360: Thanks for the detailed answer!
Let me know if this is now resolved.

ascjones (approved):
LGTM. Thanks!

tareknaser360:
> Hey @tareknaser360, some of the tests e.g. for `transcoder` are failing, which is why the CI is red. Can you resolve those issues then we can merge this.

I looked into the failing tests in the CI environment and made the following updates:
- Updated the examples in the new crates' documentation to resolve the test failures. 
- Modified the examples to initialize `BalanceVariant` instances using strings, which is the most suitable method at the moment.

I also rebased on top of master.

PR #1184, opened 2023-06-28
Decode new `ink!` events with signature topic

ascjones:
Events are identified using the signature topic rather than the first byte.

Companion for https://github.com/paritytech/ink/pull/1827

**Changes to the `decode` command**: added the `--signature-topic` argument for decoding events, which is not required for `message` and `constructor`, so I split the decode types into their own subcommands e.g. `cargo contract decode event --signature-topic` and `cargo contract decode message`


SkymanOne (approved):
LGTM!

PR #1189, opened 2023-07-01
Enable Wasm sign_ext

athei:
Since Rustc 1.70 we can't disable signed extension anymore. Hence we have to allow it in order to keep functioning. Eventually, we want to remove the `parity-wasm` dependency but for now we can just enable the `sign_ext` feature.

Please note that those contracts will not deploy on `pallet-contracts`, yet. Support is coming soon.

### DIFF
--- a/crates/build/Cargo.toml
+++ b/crates/build/Cargo.toml
@@ -27,7 +27,7 @@ rustc_version = "0.4.0"
 scale = { package = "parity-scale-codec", version = "3.0.0", features = ["derive"] }
 toml = "0.7.5"
 tracing = "0.1.37"
-parity-wasm = "0.45.0"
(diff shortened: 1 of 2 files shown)


SkymanOne:
Let's keep the PR open until the support comes so we don't break the master

athei:
How would we break master? And which master? This PR blocks development of this repo unless you downgrade rustc.

athei:
Added a warning if a new rustc is used. I didn't want to scan for the instructions to not add additional dependencies on `parity-wasm`.

ascjones:
Reverted my CI change because the integration tests fail with `CodeRejected`. Can update the CI back to `stable` once `substrate-contracts-node` is updated to support `sign_ext`.

PR #1209, opened 2023-07-10
Configure tty output correctly

SkymanOne:
Previously `--output-json` and `--quiet` flag still produced some compilation status output due to misconfigured tty output. This PR provides a fix.

### DIFF
--- a/crates/build/src/util/mod.rs
+++ b/crates/build/src/util/mod.rs
@@ -104,9 +104,9 @@ pub fn cargo_tty_output(cmd: Expression) -> Expression {
         .map(|(width, _)| width.to_string())
         .unwrap_or_else(|| "100".to_string());
 
-    cmd.env("CARGO_TERM_COLOR", "always")
+    cmd.env("CARGO_TERM_COLOR", "auto")
         .env("CARGO_TERM_PROGRESS_WIDTH", term_size)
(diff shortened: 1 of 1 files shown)


crates/build/src/util/mod.rs
```diff
@@ -104,9 +104,9 @@ pub fn cargo_tty_output(cmd: Expression) -> Expression {
         .map(|(width, _)| width.to_string())
         .unwrap_or_else(|| "100".to_string());
 
-    cmd.env("CARGO_TERM_COLOR", "always")
+    cmd.env("CARGO_TERM_COLOR", "auto")
         .env("CARGO_TERM_PROGRESS_WIDTH", term_size)
-        .env("CARGO_TERM_PROGRESS_WHEN", "always")
+        .env("CARGO_TERM_PROGRESS_WHEN", "auto")
```
ascjones, line 109: Interesting that this actually works to still correctly detect and display the colour and progress bar, the docs say that the [default is `auto`](https://doc.rust-lang.org/cargo/reference/config.html#termprogresswhen) and yet it does not work when we explicitly set it. :shrug: 

We could go a step further here and pass in the verbosity to this function so that it only sets the `TERM_PROGRESS` vars when it is not `Verbosity::Quiet`

PR #1210, opened 2023-07-11
Suggestions for PR #1148: verifiable build docker interaction

ascjones:
Demonstrating some partially completed suggestions for https://github.com/paritytech/cargo-contract/pull/1148#pullrequestreview-1524627396.

In order to work you need to build a local image with e.g. `docker build . -t local --build-arg CARGO_CONTRACT_GIT=https://github.com/paritytech/cargo-contract --build-arg CARGO_CONTRACT_BRANCH=aj/verifiable-build`. This will install the version of `cargo-contract` from this branch, which is necessary so that the `stdout` and `stderr` are correctly piped.

## todo
- [ ] mount dir with `crates.io` index to avoid downloading index for newly spawned container.
- [ ] display pull image progress bar properly, need to clear previous cursor
- [ ] Sort out build steps.

### DIFF
--- a/crates/build/src/docker.rs
+++ b/crates/build/src/docker.rs
@@ -132,7 +132,14 @@ pub fn docker_build(args: ExecuteArgs) -> Result<BuildResult> {
                 }
             };
 
-            pull_image(client.clone(), &image, &verbosity, &mut build_steps).await?;
+            if let Err(err) = pull_image(client.clone(), &image, &verbosity, &mut build_steps).await {
+                // If the image could not be pulled, we will still attempt to use a local image of
+                // that name if it exists.
+                eprintln!(
+                    "{}",
+                    format!("Failed to pull the docker image {}: {}", image, err).yellow().bold(),
+                );
+            }
 
             let build_result = run_build(
                 args,
--- a/crates/build/src/docker.rs
+++ b/crates/build/src/docker.rs
@@ -42,7 +42,11 @@ use std::{
         Hash,
         Hasher,
     },
-    io::BufReader,
+    io::{
+        Read,
+        Write,
+        BufReader
+    },
     path::{
         Path,
         PathBuf,
@@ -56,6 +60,8 @@ use anyhow::{
 };
 use bollard::{
     container::{
+        AttachContainerResults,
+        AttachContainerOptions,
         Config,
         CreateContainerOptions,
         ListContainersOptions,
@@ -359,11 +365,6 @@ async fn run_build(
         .start_container::<String>(&container_id, None)
         .await?;
 
-    let start_time = std::time::SystemTime::now()
-        .duration_since(std::time::UNIX_EPOCH)
-        .context("Invalid start time")?
-        .as_secs() as i64;
-
     maybe_println!(
         verbosity,
         " {} {}",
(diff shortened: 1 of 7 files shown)


crates/build/src/docker.rs
```diff
@@ -172,130 +168,40 @@ pub fn docker_build(args: ExecuteArgs) -> Result<BuildResult> {
...
-            metadata_artifacts,
-            metadata,
-            build_steps,
-            verbosity,
-        )?;
-    }
-    Ok(())
+    let mut metadata = ContractMetadata::load(&metadata_artifacts.dest_bundle)?;
+
+    metadata.image = Some(build_image.to_owned());
```
SkymanOne, line 180: If the `latest` tag is used, you are not actually persisting the version of the image. We need to use only the specific tag

crates/build/src/docker.rs
```diff
@@ -146,23 +122,43 @@ pub fn docker_build(args: ExecuteArgs) -> Result<BuildResult> {
...
+                    format!("{}:{}", IMAGE, VERSION)
+                }
+            };
+
+            if let Err(err) =
+                pull_image(client.clone(), &image, &verbosity, &mut build_steps).await
+            {
+                // If the image could not be pulled, we will still attempt to use a local
+                // image of that name if it exists.
+                eprintln!(
```
SkymanOne, line 138: This will always error if the image has never been published. So, that's why I checked if there is a local copy first. Although it might be more code, the error would be more precise.

crates/build/src/docker.rs
```diff
@@ -146,23 +122,43 @@ pub fn docker_build(args: ExecuteArgs) -> Result<BuildResult> {
...
                 &crate_metadata.contract_artifact_name,
                 &host_dir,
                 &verbosity,
                 &mut build_steps,
             )
             .await?;
 
-            let build_result = read_build_result(&host_dir, &build_result_path)?;
+            let metadata_artifacts = MetadataArtifacts {
+                dest_metadata: crate_metadata.metadata_path(),
```
SkymanOne, line 157: This will give only the relative path. Although there might be an CI config that rely on the absolute paths

crates/build/src/docker.rs
```diff
@@ -345,122 +249,71 @@ async fn run_build(
...
-                anyhow::bail!("{}. Execution failed!", e.to_string())
+    // pipe docker attach output into stdout
+    let stderr = std::io::stderr();
+    let mut stderr = stderr.lock();
+    let mut build_result = None;
+    while let Some(Ok(output)) = output.next().await {
+        match output {
+            LogOutput::StdOut { message } => {
+                build_result = Some(
+                    serde_json::from_reader(BufReader::new(message.as_ref()))
```
SkymanOne, line 287: Agreed, that's the better approach than creating and reading a file.

crates/build/src/docker.rs
ascjones: Okay, best of both worlds would be then to optimistically create the container, and if that fails with "not found" then attempt to pull: https://github.com/paritytech/cargo-contract/pull/1210/commits/c7bc50609a15c3fd38f75ca62c3655133f8f18e5

ascjones:
Approved #1148 now, can resurrect some more ideas from here in follow-ups.

Issue #1195, opened 2023-07-05
Improved Integration Flexibility for the Extrinsics Crate


tareknaser:
# Description:
The current implementation of the `extrinsics` crate's commands presents limitations when integrating them into other projects. These commands are primarily designed to work with `ink!` contracts and lack the necessary flexibility to accommodate customization and usage in other projects. For example: 
 - The existing commands heavily rely on reading information from the `Cargo.toml` file to retrieve details about `ink!` contracts. This tight coupling makes it challenging to use the commands in projects that rely on alternative configuration mechanisms.
 - The command structs within the `extrinsics` crate have private fields and no constructors, making it impossible to initialize them directly from external code. 

# Proposed solutions
1- Public Fields: To enable easy customization, we can make the fields of the command structs public including `ExtrinsicOpts`.
2- Constructor Methods: Implementing constructors for the command structs would provide a convenient initialization method.
3- Moving the commands' logic from the structs into separate functions.

xermicus:
I think that adding builders to [these](https://github.com/paritytech/cargo-contract/blob/master/crates/extrinsics/src/call.rs#L61) might require minimal to no changes to cargo-contract itself, while allowing other consumers to have another interface than just parsing via clap.

ascjones:
I'm curious to understand the use case for this so I can properly think about how the API should be constructed.

xermicus:
Addressed by #1213

PR #1213, opened 2023-07-12
Enhancing Access to Extrinsic Operations through an API

tareknaser:
## Description

This pull request addresses limitations that hindered integration of the extrinsics into other projects. The existing commands were primarily designed for `ink!` contracts and lacked the necessary flexibility for customization in different project contexts. 
To overcome these challenges, this pull request introduces changes that enhance the crate's usability, and integration capabilities.

Related issue #1195

## Changes:
- Implementation of builders guarded by type states to enforce mandatory field requirements before creating objects.
- Implementation of `Default` for `ExtrinsicOpts` to simplify their usage throughout the codebase.


xermicus (commented):
To make this fully usable, we'd later also need to a way to get back the result (instead of just having it printed out). Since many of the `run` function already seem to return a result, currently the Ok variant is just a unit, this could be done in a follow up, right? @ascjones WDYT?

ascjones (commented):
As @xermicus alludes to, the issue is that the top level clap commands are tightly coupled to the invocation via a CLI (e.g. with user prompts). So I'm not sure whether it makes sense to make these commands the public API of this crate.

Another way to approach it might be to move the top level commands back to the `cargo-contract` crate, and make a finer grained API that does the transcoding and node interactions, and any CLI specific stuff (any `stdin`/`stdout`) IO be done

tareknaser360:
### What I did?
- separated the printing logic into a single `run` method within the extrinsics structs.
- modified all other methods to exclude printing.

### Why?
- By isolating printing within the `run` method, the extrinsics can be easily integrated into other projects not just CLIs.
-  Using builders to initialize an extrinsic object, developers can call only the necessary methods without triggering any terminal printing, making it an adaptable API for interacting with the deployed contracts.
- The `run` method serves as an excellent template for CLI invocation in other projects.
- The changes are non-invasive to cargo-contract. 
	- eg. all the code needed to upload a contract remains in `upload.rs`.

xermicus:
That looks better now, but I'm still missing a few things.

Let's take the `run` method of the `UploadCommand` for example.

This method creates the `code` and `signer` that is then passed to `upload_code`. Why are `code` and `signer` not crafted in the `UploadCommandBuilder::done()` method? If the idea of the API is that users just use the builders, and the builders know already about the extrinsics opts, why do I need to use them again when using the command after building it. The builder consumes the ExtrinsicOpts, so if I am not using run, I need to create them again or clone them. It creates a weird API where it's confusing as to why I supply `ExtrinsicOpts` but then need to use them again, potentially confusing `signer` and `code`.

Also, I do think the printing logic should be factored out further to go back into cargo contract. If I have my own tool like this:

```
use std::path::PathBuf;

use contract_extrinsics::{ExtrinsicOpts, UploadCommand};

fn main() {
    let extr = ExtrinsicOpts::new()
        .suri("//Alice".into())
        .file(PathBuf::from("Foo.contract"))
        .done();
    let upld = UploadCommand::new().extrinsic_opts(extr).done();
    upld.run().unwrap();
}
```

This prints:
```To submit the transaction and execute the call on chain, add -x/--execute flag to the command.```

But I might not have such a flag. So this is specific to `cargo-contract` and should not be here.

Maybe we can even make the the `UploadCommand` generic over dry run. The DryRun impl could implement a `dry_run` and `upload_code_rpc` methods, and the impl for executing would have `run` and `upload_code`? `run` could return `Result<CodeHashResult, ..>`, where as `dry_run` would return `Result<UploadDryRunResult, _>`. Both can then be printed accordingly in `cargo-contract` or whoever uses them.


Small nitpicks on the builder I noticed:
- Can for example set `.file()` multiple times, I don't know if we really want to allow this.
- `.suri()` would be nicer if it would take an `impl ToString` or something so that I don't have to do `"//Alice".to_string()`.
- The same for `.file()`; if this would take `Into<PathBuf>` as argument instead, user could just do something like `.file("Foo.contract")`

**Comment by tareknaser360** [comment]
Having `done` method of the builders craft the necessary artifacts for the extrinsics structs would introduce two problems:
1- These structs are built to be used from the command line and although we can skip them from the command line parsing or even have them as optional, having complex types in the structs would prevent `cargo contract` from implementing `Debug` all over the project.
2- `done()` is only invoked when all the required fields of the structs are specified and the object is ready to be created. This is the case when the struct is being built manually from the code but `done()` can't be called on objects created using user's command line arguments since `clap` handles the creation of the object. `done()` is implemented only for `UploadCommandBuilder<state::ExtrinsicOptions>`.

I propose having a method `preprocess` to handle getting the necessary information for other methods. 
This prevents the user from providing wrong data to the methods of the structs while also keeping the code organized and clean. 
I implemented this method for `upload`, `call` and `remove`. 
- for `upload`:
	- `preprocess` returns the code and the signer.
	- I further abstracted `upload_code_rpc` to return `UploadDryRunResult`. 
	- Now, `run` method contains only printing
- for `call`:
	- `preprocess` returns the transcoder, call data and the signer.
	- I couldn't modify `call_dry_run` to return `CallDryRunResult` since the value returned now is printed as it is in `cargo contract`
- for remove:
	- `preprocess` returns the transcoder, the signer and the final code hash.

**Comment by xermicus** [comment]
> we can define another struct UploadExec. we can then add whatever information needed to upload a contract 

This sounds good to me. So the general idea is that the `Command` can be parsed from clap or use the builders. This can be then turned into a  `Exec` struct that doesn't require `Debug` impls, can have struct fields that are not supposed to be parsed by clap etc.. This could even happen in a builder fashion (as if your "prepare" method would be a "done"), then users of the API don't even realize this right?

Additionally, regarding the printing. As we just discussed, it's probably better to move the `run()` methods that does mainly print things to the command line back into cargo contract. It would also be worthwhile to investigate whether these `Result` structs that have a `.print()` method could just implement `Debug` instead.

**Comment by tareknaser** [comment]
- The fields within the `Exec` structs are private. As there are no builders for these structs, they can now be initialized exclusively through the `preprocess` method of the commands. Access to the fields can be achieved via the getters.
- Certain structs, such as `UploadDryRunResult` and `CallDryRunResult`, are used solely for printing purposes. As a result, these structs were modified to have public fields to facilitate their role in printing.
- Printing operations have been centralized within the `cargo-contract` crate, which  integrates the `contract-extrinsics` crate. This setup serves as an example for leveraging the API.
- Builder methods have been adjusted to accept `Into<String>` rather than just `String`. Similar changes were applied to `file` and other applicable fields in the builders.
- Using the extrinsics now from code is simple. `crates/cargo-contract/src/main.rs` is a good example but we can even skip some printing. Here is a file representing uploading code from a local contract

```rust
use contract_extrinsics::{ExtrinsicOpts, UploadCommand};

#[tokio::main]
async fn main() {
    let opts = ExtrinsicOpts::new()
        .file("flipper.contract")
        .suri("//Alice")
        .done();
    let upload = UploadCommand::new().extrinsic_opts(opts).done().await;

    let result = upload.upload_code_rpc().await.unwrap().unwrap();
    println!(
        "Code uploaded successfully.\nCode hash: {:?}",
        result.code_hash
    );
}
```
or 
```rust
use contract_extrinsics::{ExtrinsicOpts, UploadCommand};

#[tokio::main]
async fn main() {
    let opts = ExtrinsicOpts::new()
        .file("flipper.contract")
        .suri("//Alice")
        .done();
    let upload = UploadCommand::new().extrinsic_opts(opts).done().await;

    let upload_result = upload.upload_code().await.unwrap();
    if let Some(code_stored) = upload_result.code_stored {
        println!(
            "Contract uploaded successfully.\nCode hash: {:?}",
            code_stored.code_hash
        );
    } else {
        println!("This contract has already been uploaded");
    }
}
```

crates/extrinsics/src/call.rs
```diff
@@ -90,12 +87,123 @@ pub struct CallCommand {
...
+            marker: PhantomData,
+        }
+    }
+
     pub fn is_json(&self) -> bool {
         self.output_json
     }
 
-    pub fn run(&self) -> Result<(), ErrorVariant> {
+    /// Helper method for preprocessing contract artifacts.
```
xermicus, line 205: Just by the method name it's not clear when and how this should be used. Can you please explain it better in the doc comment?

crates/extrinsics/src/call.rs
```diff
@@ -104,151 +212,90 @@ impl CallCommand {
...
+    value: BalanceVariant,
+    output_json: bool,
+    client: Client,
+    transcoder: ContractMessageTranscoder,
+    call_data: Vec<u8>,
+    signer: Keypair,
+}
+
+impl CallExec {
+    pub async fn call_dry_run(&self) -> Result<ContractExecResult<Balance, ()>> {
```
xermicus, line 250: Would be nice if this had a doc-comment

crates/extrinsics/src/call.rs
```diff
@@ -104,151 +212,90 @@ impl CallCommand {
...
     }
 
-    async fn call(
-        &self,
-        client: &Client,
-        data: Vec<u8>,
-        signer: &Keypair,
-        transcoder: &ContractMessageTranscoder,
-    ) -> Result<(), ErrorVariant> {
+    pub async fn call(&self, gas_limit: Weight) -> Result<DisplayEvents, ErrorVariant> {
```
xermicus, line 270: Would be nice if this had a doc-comment

crates/extrinsics/src/call.rs
```diff
@@ -104,151 +212,90 @@ impl CallCommand {
...
-        };
-        println!("{output}");
-
-        Ok(())
+        Ok(display_events)
     }
 
     /// Dry run the call before tx submission. Returns the gas required estimate.
-    async fn pre_submit_dry_run_gas_estimate(
+    pub async fn pre_submit_dry_run_gas_estimate(
```
xermicus, line 294: I'm wondering whether a simpler name can be used, e.g. something like `estimate_gas`?

crates/extrinsics/src/instantiate.rs
```diff
@@ -377,7 +455,10 @@ impl Exec {
     }
 
     /// Dry run the instantiation before tx submission. Returns the gas required estimate.
-    async fn pre_submit_dry_run_gas_estimate(&self) -> Result<Weight> {
+    pub async fn pre_submit_dry_run_gas_estimate(
```
xermicus, line 458: Is this method necessary to have in call and instantiate?

I'm wondering, it should be possible to just submit the dry run to get the estimated gas regardless of it's an instantiation or call right.

crates/extrinsics/src/instantiate.rs
```diff
@@ -377,7 +455,10 @@ impl Exec {
...
+        &self,
+        print_to_terminal: bool,
```
xermicus, line 460: Feels a bit odd for an API to have printing built in. Based on the code it seems possible to offload all printing to the caller.

crates/extrinsics/src/call.rs
```diff
@@ -90,12 +87,123 @@ pub struct CallCommand {
...
+impl<M, E> CallCommandBuilder<M, E> {
+    /// Sets the the address of the the contract to call.
+    pub fn contract(self, contract: <DefaultConfig as Config>::AccountId) -> Self {
+        let mut this = self;
+        this.opts.contract = contract;
+        this
+    }
+
+    /// Sets the arguments of the contract message to call.
+    pub fn args(self, args: Vec<String>) -> Self {
```
xermicus, line 137: A `Vec<T> where T: ToString` would be a more useful type for this argument

crates/extrinsics/src/instantiate.rs
```diff
@@ -317,48 +375,68 @@ impl Exec {
...
     }
 
-    async fn display_result(
+    pub async fn instantiate(
         &self,
-        result: &ExtrinsicEvents<DefaultConfig>,
-        code_hash: Option<CodeHash>,
-        contract_address: subxt::utils::AccountId32,
-        token_metadata: &TokenMetadata,
+        gas_limit: Weight,
```
xermicus, line 388: Cherry on top would be to have this as an `Option`, where this method would first issue a dry-run to estimate the gas for you. Same for the `call` version.

xermicus (commented):
That looks much better now, thanks! Already makes re-using it much easier.

I gave a first round of review on the updated version. @ascjones please have a look too once you are back and have the time for it.

crates/extrinsics/src/call.rs
xermicus: Also it would be nice to be able to supply arguments directly, not having to encode them as strings first. But for CLIs it doesn't matter, because strings is what we get anyways. I think this would be a nice addition in the future.

crates/extrinsics/src/instantiate.rs
tareknaser: Thanks for the feedback! I agree that offloading printing to the caller could be a cleaner approach in some cases. However, in this specific scenario, there are a few reasons for the current design:

Printing statements are distributed throughout various sections of the function.
consider this part for example
```rust
match instantiate_result.result {
  Ok(_) => {
      if !self.output_json && print_to_terminal {
          super::print_gas_required_success(instantiate_result.gas_required);
  }
```
- The `instantiate_result.gas_required` value is printed within the function itself but is not returned as part of the function's result. Trying to mimic this behavior by moving the printing outside of the function would require replicating significant portions of the function's code in `cargo-contract` crate.
- Moving the printing logic outside of the API would require a significant amount of code changes throughout the `cargo-contract` crate. This could lead to a less maintainable codebase, and in some cases, it may even require exposing private internals, which is not ideal.
for example, we would need to expose functions like `print_gas_required_success` and `print_dry_running_status` which are now private.

The current approach with the printing parameter offers flexibility and convenience to users, especially when dealing with CLI tools. However, I think this makes for a good future improvement that might involve separating the printing logic from the API.

tareknaser:
I've written two new tests for the updated API, which mirror existing tests focused on the entire lifecycle of cargo contract extrinsics. 
I also encountered an issue while attempting to run integration tests due to the current configurations for conditional compilation in the `integration_tests` file. I've run the tests locally to ensure they pass successfully.

crates/extrinsics/src/instantiate.rs
xermicus: >  moving the printing outside of the function would require replicating significant portions of the function's code in cargo-contract crate

So, this printing stuff is specific to cargo contract, and now this just shows why this code really doesn't belong into a generic contract extrinsic crate. Also the function is not very long, should be a rather small amount of work to move it into cargo contract. 

> Moving the printing logic outside of the API would require a significant amount of code changes throughout the cargo-contract crate

I don't see why a bunch of `print!` invocations requires significant code changes.

> we would need to expose functions like print_gas_required_success and print_dry_running_status

Can't this API here just expose a method to get the gas estimate and then anyone using it (for example cargo contract) can just build their own helpers which print stuff to the console around this if they want?

tareknaser: Agreed
I've created dedicated helper functions within `cargo-contract` crate to handle both gas estimation along with printing. The API now exclusively offers `estimate_gas` methods for gas estimation, leaving the printing functionality to `cargo contract`'s own helper methods.

crates/cargo-contract/src/main.rs
```diff
@@ -178,30 +199,256 @@ fn exec(cmd: Command) -> Result<()> {
...
+}
+
+fn handle_instantiate(
+    instantiate_command: &InstantiateCommand,
+) -> Result<(), ErrorVariant> {
+    Runtime::new()?.block_on(async {
+        let instantiate_exec = instantiate_command.preprocess().await?;
+
+        if !instantiate_exec.opts().execute() {
+            let result = instantiate_exec.instantiate_dry_run().await?;
```
ascjones, line 289: This is called twice...here and then inside `simulate_instantiation`. Should just be called once.

crates/extrinsics/src/instantiate.rs
```diff
@@ -190,75 +307,51 @@ pub struct Exec {
...
+impl InstantiateExec {
+    /// Simulates the instantiation of a contract without executing it on the blockchain.
+    ///
+    /// This function performs a dry-run simulation of the contract instantiation process
+    /// and returns an [`InstantiateDryRunResult`] object containing essential
+    /// information, including the contract address, gas consumption, and storage
+    /// deposit.
+    ///
+    /// It does not modify the state of the blockchain.
+    pub async fn simulate_instantiation(
```
ascjones, line 319: I know there is already the `dry_run_instantiate` fn, but we should not introduce a new term here `simulate`, even if it is a correct synonym for `dry_run`. Find a way to unify these methods or rename this one. Should be similar to pattern in `call`

crates/extrinsics/src/lib.rs
```diff
@@ -143,7 +164,127 @@ pub struct ExtrinsicOpts {
...
+    /// Type state for the Secret key URI.
+    pub struct Suri;
+    /// Type state for extrinsic options.
+    pub struct ExtrinsicOptions;
+    /// Type state for the name of the contract message to call.
+    pub struct Message;
+}
+
+/// A builder for extrinsic options.
+pub struct ExtrinsicOptsBuilder<Suri> {
```
ascjones, line 184: Please extract `ExtrinsicsOpts` and builders etc. to their own file.

crates/extrinsics/src/call.rs
```diff
@@ -90,12 +85,131 @@ pub struct CallCommand {
...
+    /// Preprocesses contract artifacts and options for subsequent contract calls.
+    ///
+    /// This function prepares the necessary data for making a contract call based on the
+    /// provided contract artifacts, message, arguments, and options. It ensures that the
+    /// required contract code and message data are available, sets up the client, signer,
+    /// and other relevant parameters, preparing for the contract call operation.
+    ///
+    /// Returns the `CallExec` containing the preprocessed data for the contract call,
+    /// or an error in case of failure.
+    pub async fn preprocess(&self) -> Result<CallExec> {
```
ascjones, line 212: Personally I would prefer `prepare` as a name

ascjones (commented):
There is still some CLI specific code in `extrinsics`. Is the end purpose of the libraray to be a pure "invoke contract extrinsics" crate? Or "invoke contract extrinsics" plus convenience CLI I/O methods for building your own CLI?

If we went full pure extrinsics invocation there could be even more separation, even moving the `clap` commands back to `cargo-contract`, which would involve conversion to extrinsics commands but would make clear the separation of concerns.

xermicus:
Regarding `"invoke contract extrinsics" plus convenience CLI I/O methods`: I'd say if these convenience methods are useful for many or all use cases (i.e. gas estimation), they can go in. So, for example the `clap` commands for cargo-contracts as well as anything that does print to the command line do not fall into this category.

ascjones:
> Regarding `"invoke contract extrinsics" plus convenience CLI I/O methods`: I'd say if these convenience methods are useful for many or all use cases (i.e. gas estimation), they can go in. The `clap` commands for cargo-contracts as well as anything that does print to the command line do not fall into this category.

The tradeoff there is that now those helpers become part of the API to the crate, so people will start depending on them and it is less easy to do refactoring or change the way they behave without the possibility of breaking somebody's downstream code. 

So possibly introducing a higher maintenance cost.

xermicus:
@tareknaser if you move the CLI structs parsed with `clap` back into cargo contracts, (you might want to store all required fields in the builder instead) and implement what `preprocess` does on the builder instead, I think not much cargo-contract specific code will be left. Also it should be easy to spot without the CLI structs.

crates/extrinsics/src/instantiate.rs
tareknaser: This function essentially serves as a decoder for the result obtained from the `dry_run_instantiate` method. I've renamed it to `decode_instantiate_dry_run` and adjusted its parameter list to accept the result generated by `dry_run_instantiate`.

xermicus (commented):
In the integration tests module, there is `#[cfg(any(feature = "integration-tests", feature = "test-ci-only"))]`, so you need to add these features to the `Cargo.toml` of the extrinsics library.

xermicus:
Also the CI tests did pass despite the integration test being broken with the missing feature flags. Maybe a missing `--workspace` argument to `cargo test` in the CI config?

crates/cargo-contract/src/cmd/instantiate.rs
```diff
@@ -0,0 +1,309 @@
...
+    if instantiate_exec.opts().skip_dry_run() {
+        return match (instantiate_exec.args().gas_limit(), instantiate_exec.args().proof_size()) {
+                (Some(ref_time), Some(proof_size)) => Ok(Weight::from_parts(ref_time, proof_size)),
+                _ => {
+                    Err(anyhow!(
+                        "Weight args `--gas` and `--proof-size` required if `--skip-dry-run` specified"
+                    ))
+                }
+            };
+    }
```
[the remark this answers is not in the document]
tareknaser: >  If it were an Option<Weight> instead, you can estimate the gas if it's `None`

This logic is already implemented in `call` and `instantiate` methods 
for example

```rust
pub async fn call(
    &self,
    gas_limit: Option<Weight>,
) -> Result<DisplayEvents, ErrorVariant> {
    // use user specified values where provided, otherwise estimate
    let gas_limit = match gas_limit {
        Some(gas_limit) => gas_limit,
        None => self.estimate_gas().await?,
    };
    // ...
    // Logic to call the contract 
```
I think a better approach is to modify `estimate_gas` methods to use what was provided in the builders directly
- if both `gas_limit` and `proof_size` were given, we estimate the gas using them
- otherwise, we do a dry run to estimate the gas

and as I mentioned before, `call` and `instantiate` methods have an optional parameter of type `Weight` so the user of the crate can actually ignore `estimate_gas` completely and pass the maximum gas directly to the methods.

Also, cargo contract crate doesn't use `estimate_gas` at all.
it implements its helper function `pre_submit_dry_run_gas_estimate_instantiate` that does the same logic but with checking for `skip_dry_run` flag and some printing.

xermicus (approved):
Looks much better now.

@ascjones I'll wait for your review before merging

crates/cargo-contract/src/cmd/call.rs
```diff
@@ -0,0 +1,249 @@
...
+    output_json: bool,
+}
+
+impl CallCommand {
+    /// Returns whether to export the call output in JSON format.
+    pub fn output_json(&self) -> bool {
+        self.output_json
+    }
+}
+pub async fn handle_call(call_command: &CallCommand) -> Result<(), ErrorVariant> {
```
ascjones, line 73: Not a huge thing but these `handle_` methods could be inherent methods on their respective commands. e.g. 

``rust
impl CallCommand {
  pub async fn handle(&self) { .. }
}
```

File: crates/extrinsics/src/call.rs [hunk 15]
```diff
@@ -16,278 +16,364 @@
...
+    ///
+    /// This function prepares the necessary data for making a contract call based on the
+    /// provided contract artifacts, message, arguments, and options. It ensures that the
+    /// required contract code and message data are available, sets up the client, signer,
+    /// and other relevant parameters, preparing for the contract call operation.
+    ///
+    /// Returns the `CallExec` containing the preprocessed data for the contract call,
+    /// or an error in case of failure.
+    pub async fn done(self) -> CallExec {
+        let artifacts = self.opts.extrinsic_opts.contract_artifacts().unwrap();
```
ascjones, line 173: This and other `done` methods should return `Result` instead of having all these `unwrap`s

ascjones (commented):
Okay, nearly there. I would like all those `unwrap`s gone from the `done` methods then we can merge this :+1:

crates/extrinsics/src/call.rs
```diff
@@ -16,278 +16,362 @@
...
+                        let object = ErrorVariant::from_dispatch_error(
+                            err,
+                            &self.client.metadata(),
+                        )?;
+                        Err(anyhow!("Pre-submission dry-run failed. Error: {}", object))
+                    }
                 }
             }
         }
     }
```
xermicus, line 319: This function is duplicate code in call and instantiate

xermicus:
@tareknaser CI failed

crates/cargo-contract/src/cmd/call.rs
ascjones, line 1: Missing file header

crates/cargo-contract/src/cmd/instantiate.rs
ascjones, line 1: Missing file header

crates/extrinsics/src/extrinsic_opts.rs
ascjones, line 1: Check all new files for missing headers

ascjones (approved):
LGTM!

Just missing some file headers on the new files.

crates/extrinsics/src/call.rs
xermicus: Well, the matching on the result could go into a helper just like that. Everything could go into a helper if `Exec` was a trait instead. But this might require a bit more careful design. Also it's not the end of the world for me . Let's not bike shed the PR forever. The PR is already a good move into the right direction, we can still improve from here.

xermicus:
Thank you @tareknaser, good stuff!

PR #1255, opened 2023-08-02
Add std to cargo template in generate-metadata

### LINKED ISSUE #1254: undefined reference to `__ink_generate_metadata' again
[This](https://github.com/paritytech/cargo-contract/issues/909) bug has reappeared due to a change to the template cargo.toml

Offending commit:
https://github.com/paritytech/cargo-contract/commit/94d96c5df60dc43d01f7329dbcef99c9454e9314

`std` needs to be enabled here:

https://github.com/paritytech/cargo-contract/blob/94d96c5df60dc43d01f7329dbcef99c9454e9314/crates/build/templates/generate-metadata/_Cargo.toml#L13

forgetso:
Fixes #1254

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,15 +22,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "addr2line"
-version = "0.20.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


crates/build/templates/generate-metadata/_Cargo.toml
```diff
@@ -10,7 +10,7 @@ name = "metadata-gen"
 path = "main.rs"
 
 [dependencies]
-contract = { path = "../.." }
+contract = { path = "../..", features= ["std"] }
```
ascjones, line 13: `std` should be already enabled via the `default` feature below?

ascjones (commented):
Indeed this is how it originally was but I moved this to the `default` feature (see below), so it should be enabled (we are not invoking `cargo run` with `--no-default-features`).

I am unable to reproduce the error, can you provide a reproduction so I can understand what is going on?

ascjones:
@dependabot appears to be drunk and merged this with 0 changes, not sure what happened there.

goastler:
I don't recall force-pushing to this, I have no idea what's happened here! I've been making some changes on our fork of cargo contract today, but these were very minor (added a file and ~5 lines changed in another) and had no involvement with the master branch on our end nor parity's.

Genuinely no idea what's happened here!

ggwpez:
The MR has +-0 and 0 commits since you force-pushed a master commit. It therefore breaks and is marked as "merged" although nothing actually was merged.

PR #1280, opened 2023-08-17
Add `SignextLowering` pass to `wasm-opt`

### LINKED ISSUE #1239: cargo-contract 3.1.0 emits sign_ext ops when using rustc 1.69.
Cargo-contract says "`Build using a version lower than 1.70.0 to make sure your contract will deploy`".

However, we have noticed that sign_ext operations appear in the wasm build with `cargo-contract-3.1.0+rustc-1.69` and there are no such problems when building with `cargo-contract-3.0.1+rustc-1.69`.

This might be due to #1189 enabling SignExt for wasm-opt, leading to the emission of sign_ext instructions even if Rustc didn't originally produce them.

[Here](https://github.com/Phala-Network/phat-bricks/tree/master/phat/contracts/action_offchain_rollup) is the contract.

ascjones:
Fixes #1239.

As suggested here: https://github.com/paritytech/cargo-contract/issues/1239#issuecomment-1673810089, this will allow a newer toolchain > `1.69` to be used to deploy to older versions of `pallet-contracts` which do not yet support `signext`.

It does so by adding an extra `wasm-opt` pass which "lowers" any `signext` instructions to mvp only features.

Tested that a contract built with Rust >= `1.70` can be uploaded to a node which does not support `signext`.

If it does work we can then consider either backporting this to a `3.2.0` release, or just going ahead and doing a `4.0.0` release.

### DIFF
--- a/crates/build/src/wasm_opt.rs
+++ b/crates/build/src/wasm_opt.rs
@@ -18,6 +18,7 @@ use anyhow::Result;
 use wasm_opt::{
     Feature,
     OptimizationOptions,
+    Pass,
 };
 
 use std::{
@@ -68,9 +69,14 @@ impl WasmOptHandler {
         );
 
         OptimizationOptions::from(self.optimization_level)
-            // Since rustc 1.70 `SignExt` can't be disabled anymore. Hence we have to allow it.
             .mvp_features_only()
+            // Since rustc 1.70 `SignExt` can't be disabled anymore. Hence we have to allow it,
+            // in order that the Wasm binary containing these instructions can be loaded.
             .enable_feature(Feature::SignExt)
+            // This pass will then remove any `signext` instructions in order that the resulting
+            // Wasm binary is compatible with older versions of `pallet-contracts` which do not
+            // support the `signext` instruction.
+            .add_pass(Pass::SignextLowering)
             // the memory in our module is imported, `wasm-opt` needs to be told that
             // the memory is initialized to zeroes, otherwise it won't run the
             // memory-packing pre-pass.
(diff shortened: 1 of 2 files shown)


crates/build/src/wasm_opt.rs
```diff
@@ -68,9 +69,14 @@ impl WasmOptHandler {
...
         OptimizationOptions::from(self.optimization_level)
-            // Since rustc 1.70 `SignExt` can't be disabled anymore. Hence we have to allow it.
             .mvp_features_only()
+            // Since rustc 1.70 `SignExt` can't be disabled anymore. Hence we have to allow it,
+            // in order that the Wasm binary containing these instructions can be loaded.
             .enable_feature(Feature::SignExt)
+            // This pass will then remove any `signext` instructions in order that the resulting
+            // Wasm binary is compatible with older versions of `pallet-contracts` which do not
+            // support the `signext` instruction.
+            .add_pass(Pass::SignextLowering)
```
agryaznov, line 79: Since https://github.com/paritytech/substrate/pull/14565 we do allow sign_extension in smart contracts on the pallet side. Shouldn't we make this pass optional for the cargo-contract so that the users of the newer versions of pallet-contracts could still opt in to have signext instructions in their contracts?

ascjones: I am aware, and I did consider making this optional. I decided against it because removing them by default means that both old and new versions of `pallet_contracts` will be supported with this, so will make users and our lives easier if the resulting wasm is compatible with as wide a range of target nodes as possible.

I assume the majority of node implementations currently are using an older version of `pallet_contracts` which does not support `signext`.

agryaznov: I see. Could adding smth like `--allow-signext` to the `cargo-contract contract build` so that `wasm_opt` runs without the pass, be an option then?

cmichi (approved):
Thanks!

PR #1290, opened 2023-08-21
Specify prod chain URL with names and check for the verifiable build upon upload

### LINKED ISSUE #1274: Warn if un-verifiable contract is getting deployed on a production chain
Feat #1065 
We should warn a user if they decide to deploy an unverifiable image (compiled without `--verifiable` flag) to a production chain.

SkymanOne:
Closes #1274

This PR introduces an additional flag `--chain` that allows you conveniently choose which production chain to deploy a contract to instead of specifying the endpoint URL. Currently, there are endpoints for Aleph Zero, Astar and Shiden according to the list of production chains on the contracts UI.

On the backend, I put up a macro that allows automatically generating required methods and trait implementations for new endpoints in the future.

Additionally, there the user is prompted with a warning if they try to deploy a contract which doesn't have an image for the verifiable build specified in the metadata file or doesn't have a metadata file at all.
If the user tries to specify the URL which appears to be a known endpoint, we recognise it as an attempt to make a call to a production chain


crates/extrinsics/src/prod_chains.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    };
+}
+
+define_chains! {
+    /// A list of all production chains where the contract can be deployed to.
+    pub enum ProductionChain {
+        AlephZero = "wss://ws.azero.dev:443",
+        Astar = "wss://rpc.astar.network:443",
+        Shiden = "wss://rpc.shiden.astar.network:443"
```
cmichi, line 83: Please include the other ones from https://use.ink/where-to-deploy.

crates/cargo-contract/src/cmd/mod.rs
```diff
@@ -310,6 +314,36 @@ where
...
+        "{} ({}): ",
+        "\nContinue?".bright_white().bold(),
+        "Y/n".bright_white().bold()
+    );
+
+    let mut buf = String::new();
+    io::stdout().flush()?;
+    io::stdin().read_line(&mut buf)?;
+    match buf.trim().to_lowercase().as_str() {
+        // default is 'y'
```
smiasojed, line 340: I am not sure if the default should be 'yes'; rather, we should recommend uploading only verifiable contracts to production chains. WDYT?

crates/cargo-contract/src/cmd/call.rs
```diff
@@ -143,6 +147,7 @@ impl CallCommand {
             .file(self.extrinsic_cli_opts.file.clone())
             .manifest_path(self.extrinsic_cli_opts.manifest_path.clone())
             .url(self.extrinsic_cli_opts.url.clone())
+            .chain(self.extrinsic_cli_opts.chain.clone())
```
smiasojed, line 150: Why do we need to pass chain to `extrinsics` crate? I think `extrinisics` crate should just get url as it gets it now. We could keep all the prod chain logic in `cargo-contracts` crate

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -74,6 +75,9 @@ pub struct InfoCommand {
         default_value = "ws://localhost:9944"
     )]
     url: url::Url,
+    /// A name of a production chain to upload or instantiate the contract on.
+    #[clap(name = "chain", long, conflicts_with = "url")]
```
smiasojed, line 79: Apart from the URL, the chain could also retain the chain configuration. So, when you specify 'alephzero', 'moonbeam', or whatever, it would also set the correct `config` (https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/config.rs) for the chain.

In this case chain would conflict with `url` and `config`

crates/cargo-contract/src/cmd/call.rs
SkymanOne: We also want to retain info about what kind of chain it is: `Prod` or `Custom`. So it is easier to encapsulate the logic this way.

crates/cargo-contract/src/cmd/info.rs
SkymanOne: This can be a follow up PR

crates/cargo-contract/src/cmd/mod.rs
SkymanOne: We do advise it, but we still want to give a user an option to upload an unverifiable contract (aka shoot their own leg)

crates/extrinsics/src/prod_chains.rs
SkymanOne: I took the list from the contrast-ui. Out of the list only Krest allows me to submit an extrinsic.

crates/cargo-contract/src/cmd/mod.rs
smiasojed: I was just thinking to make `n` as default

crates/cargo-contract/src/cmd/call.rs
smiasojed: Why do we need this information in `extrinsics` crate? The upload check is done in `cargo-contract` crate.
In `ExtrinsicOpts` we have url and chain which is kind of data duplication.

crates/cargo-contract/src/cmd/info.rs
smiasojed: Ok

crates/cargo-contract/src/cmd/call.rs
ascjones: I agree this is not belong in `extrinsics`

cmichi (approved):

cmichi:
A couple of changes are required, let's do them in follow-ups.

PR #1306, opened 2023-08-29
Add `verify` command

SkymanOne:
Supersedes #696

Adds a ` cargo contract verify <path to reference contract bundle>` command which allows to verify if the given contract code matches the reference bundle.

In addition to the previous PR, I have added the support to verify "verifiable" contract. If the contract bundle specify `image` in metadata, then we build the target contract code in the container of the specified image and then verify.

This PR also locks the version of `wasm-opt` crate which often is the cause of binary mismatch.


crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -0,0 +1,236 @@
...
+             However, the original contract was built using `{expected_rust_toolchain}`. Please\n\
+             install the correct toolchain (`rustup install {expected_rust_toolchain}`) and\n\
+             re-run the `verify` command.",);
+            anyhow::ensure!(rustc_matches, mismatched_rustc.bright_yellow());
+
+            let expected_cargo_contract_version = build_info.cargo_contract_version;
+            let cargo_contract_version = semver::Version::parse(VERSION)?;
+
+            // Note, assuming both versions of `cargo-contract` were installed with the
+            // same lockfile (e.g `--locked`) then the versions of `wasm-opt`
```
ascjones, line 120: This may prove to be a common stumbling block, where the same version of `cargo-contract` is using a different `wasm-opt` version.

We could add the `wasm-opt` version to the `build_info`, or pin `wasm-opt` to the specific version to guarantee the version consistency even if the user does not install with `--locked`

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -0,0 +1,236 @@
...
+            serde_json::from_reader(file).context(format!(
+                "Failed to deserialize contract bundle {}",
+                target_bundle.display()
+            ))?;
+
+        let target_code_hash = built_contract.source.hash;
+
+        if reference_code_hash != target_code_hash {
+            tracing::debug!(
+                "Expected Code Hash: '{:?}'\n\nGot Code Hash: `{:?}`",
```
ascjones, line 184: I'd like to see the actual code hashes here displayed on the console, instead of hidden in the logs.

crates/cargo-contract/src/cmd/verify.rs
SkymanOne: I think pinning one would be the best course of action here.

SkymanOne:
> One more thing: a lot of trust will be put into this command, therefore we should have integration tests here for both the positive and negative results of this command.

I doubt we will be able to simulate different OS environments, but minor changes in the code should work.

ascjones (approved):
LGTM.

Might be good as a next step to be able to verify from a contract account id that it is an instance of the verified contract. Should be straightforward just by looking up the code hash and then comparing that.

PR #1310, opened 2023-09-01
Detect `INK_STATIC_BUFFER_SIZE` env var

SkymanOne:
Feat https://github.com/paritytech/ink/pull/1869

`cargo-contract` now detects if the `INK_STATIC_BUFFER_SIZE` env var has been set and it is different from the value specified in the metadata. If so, we clean binaries and rebuild the contract.

If not, we simply skip the step.

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -17,6 +17,7 @@
 #![doc = include_str!("../README.md")]
 #![deny(unused_crate_dependencies)]
 
+use contract_metadata::ContractMetadata;
 use which as _;
 
 mod args;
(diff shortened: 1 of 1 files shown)


crates/build/src/lib.rs
```diff
@@ -370,6 +371,65 @@ fn exec_cargo_for_onchain_target(
...
+                .abi
+                // get `spec` field
+                .get("spec")
+                .context("spec field should be present in ABI.")?
+                // get `environment` field
+                .get("environment")
+                .context("environment field should be present in ABI.")?
+                // get `staticBufferSize` field
+                .get("staticBufferSize")
+                .context("`staticBufferSize` must be specified.")?
```
ascjones, line 405: If the value is not present we would possibly also want to do a clean still?

crates/build/src/lib.rs
```diff
@@ -370,6 +371,65 @@ fn exec_cargo_for_onchain_target(
...
+            .parse()
+            .context("`INK_STATIC_BUFFER_SIZE` must have an integer value.")?;
+
+        let cargo = util::cargo_cmd(
+            "clean",
+            Vec::<&str>::new(),
+            crate_metadata.manifest_path.directory(),
+            *verbosity,
+            vec![],
+        );
```
ascjones, line 392: Nit: I would move this closer to where it is used.

crates/build/src/lib.rs
SkymanOne: It's always present in the metadata

ascjones: Except when upgrading the version of `ink`, exceptional case I know but the build would fail in that case and require a manual clean of the artifacts? Might as well do it automatically?

ascjones (approved):

PR #1311, opened 2023-09-04
Add `--binary` flag for `info` command

smiasojed:
Closes https://github.com/paritytech/cargo-contract/issues/800

The combination --binary and --output-json returns JSON: 
```
{ 
  "wasm": "0x..."
}
```
I assumed that binary flag is a behaviour switch, and output-json just formatter.
The another option is to combine wasm hex code with the rest of info command default content:
```
{
  "trie_id": "f8ac805216a9318ab898d0777e24ac52583ca690a218556cfeb6b05744d191cf",
  "code_hash": "0x1a56980585ad4097aaea9daf1bee2332e2a9cf579d0c1a3b91a81900cb399a53",
  "storage_items": 1,
  "storage_item_deposit": 100000000000
  "wasm": "0x..."
}
```
I think that cleaner approach is the first one, where --binary is treated as behaviour switch


crates/extrinsics/src/lib.rs
```diff
@@ -363,6 +366,23 @@ impl ContractInfo {
     }
 }
 
+/// Fetch the contract wasm code from the storage using the provided client and code hash.
+pub async fn fetch_wasm_code(
+    hash: CodeHash,
+    client: &Client,
+) -> Result<Option<BoundedVec<u8>>> {
```
ascjones, line 373: Should convert the `BoundedVec` to a `Vec` for the return type: better not to expose types generated by subxt in the public api of this crate.

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -50,6 +54,9 @@ pub struct InfoCommand {
     /// Export the instantiate output in JSON format.
     #[clap(name = "output-json", long)]
     output_json: bool,
+    /// Display the contract's Wasm bytecode.
```
ascjones, line 57: Should we be more specific here and say "pristine" bytecode, i.e. non-instrumented? We may, for example add the possiblility to retrieve the instrumented byte code.

ascjones (commented):
Looks good, just the small nitpick about the return type and a question.

Also please update the description with "Closes https://github.com/paritytech/cargo-contract/issues/800", so we know which issue is being dealt with and it closes automatically when this is merged.

crates/cargo-contract/src/cmd/info.rs
smiasojed: It is named according to Michi comment https://github.com/paritytech/cargo-contract/pull/1013/files/234b258a2b11c309ab18ee09b35996ced1fb0c8e#r1143947838

But if you wish I can change it

crates/extrinsics/src/lib.rs
smiasojed: Done

crates/cargo-contract/src/cmd/info.rs
ascjones: No in that case we can keep it as it is.

ascjones (approved):
LGTM!

Issue #784, opened 2022-10-14
Add heuristic to determine contract origin from Wasm blob


cmichi:
Requires https://github.com/paritytech/cargo-contract/issues/783 to be implemented first.

The sub-command `cargo contract info --contract …` should return a heuristic info if the Wasm blob looks like an ink! 3.x one, or an ask! one or a Solidity one. Kind of like [`file(1)`](https://linux.die.net/man/1/file) works.

Block explorers will need this information and for dashboards it would be great as well. It will enable use-cases like displaying an overview how many contracts are deployed in ink! vs. other languages.

Before everybody starts building their own heuristic, we should have a central place to track this. Ideally we would put it into a library crate [`contract-info`](https://crates.io/crates/contract-info). Doesn't necessarily have to live in this repo either.

## Ideas for heuristics

Please comment if you have more ideas!

* In a discussion with @mfornos and @XY-Wang, Marc suggested to see if we can check for known values in Wasm sections to identify the contract origin.
* Another option would be function signatures that changed in between ink! versions, e.g. for our environmental functions. Though it's pretty unclear if these changes end up in the Wasm blob, after being run through the Rust compiler, LLVM, and `wasm-opt`.

One way of finding out if options are applicable is to check for breaking changes in [the ink! release notes](https://github.com/paritytech/ink/releases) or to investigate the output of `cargo contract build --release && wasm2wat target/ink/*.wasm`.

It's important that the heuristic should work for contracts build via `cargo contract build --release` and `cargo contract build` (without `--release`). This will allow to apply the command also to testnets, where people will deploy contracts with debug info.

## ToDos

- [ ] Port the Python script which was posted in a comment below to Rust.
- [ ] #1312

For examining the Wasm blob you can use the `parity-wasm` crate, we already use it in `cargo-contract`. See the function [`post_process_wasm`](https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/build/mod.rs#L515-L534) for a usage example.

mfornos:
We did a proof of concept with some **tentative heuristics** to determine the source language of the emitted contract binary.

You can find the code here:
https://gist.github.com/mfornos/7848c7d9d67eefe0e41418981a52212d

(it uses Pure Python Compiler Infrastructure to ease the handling of the byte code)

cmichi:
@mfornos This is very helpful, thank you!

cmichi:
[...]

PR #1329, opened 2023-09-12
Add heuristic to determine contract origin from Wasm blob

smiasojed:
## Summary
Closes https://github.com/paritytech/cargo-contract/issues/784

Add heuristic to determine contract origin from Wasm blob

## Description
Introduced contract-analyze crate with detects the programming language of a smart contract from its WebAssembly (Wasm) binary code.
New crate has been integrated with cargo contract info command, which generates now the output:
```
          TrieId c7fa1c1ef70d66023c5ecf4a517d39ec908c92d56e5e80af00b222c441dd0f40
       Code Hash 0x2c1388d32d59b87d0a2b6208e3562a11bed29a2d46cf1e631ee1c211a41004c1
   Storage Items 1
 Storage Deposit 999500000
 Source Language ink!
```
The downside of implementing this change is that it results in the need to download the WebAssembly (Wasm) code from the blockchain for every "info" call.


SkymanOne:
> The downside of implementing this change is that it results in the need to download the WebAssembly (Wasm) code from the blockchain for every "info" call.

Perhaps you can cache it?

smiasojed:
> 
I am not sure if it is real problem the issue could be visible when you iterate over all contracts addresses on chain to call info on them. In this case having cache<hash, wasm> could improve the performance assuming that the same contract is instantiated many times.

crates/analyze/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    let import_section_first = import_section
+        .ok_or(anyhow!("Missing required import section"))?
+        .entries()
+        .first()
+        .map(|e| e.field())
+        .ok_or(anyhow!("Missing required import section"))?;
+
+    if import_section_first != "memory"
+        && start_section.is_none()
+        && custom_sections.peek().is_none()
```
ascjones, line 48: I'm not 100% convinced by this heuristic, I know it is copied from that python script. 

It is because it is identifying the language based on the combination of a *lack* of things, which has the potential for false positives.

I would prefer to match on the presence of something, where we can be reasonably sure that it is an `ink` contract. We could add a custom section like the other languages (although that excludes contracts built with older versions). Alternatively we could identify the signature of a commonly used function somehow? e.g. a quick glance at a few contracts I can spot something that similar to:

incrementer.wat
```
(func (;8;) (type 5) (param i32)
    (local i32 i32 i32 i32)
    global.get 0
    i32.const 16
    i32.sub
    local.tee 1
    global.set 0
    local.get 1
    i64.const 16384
    i64.store offset=4 align=4
    local.get 1
    i32.const 65536
    i32.store
    local.get 1
    i32.const 0
    call 9
    block  ;; label = @1
```
flipper.wat
```
(func (;11;) (type 6) (param i32)
    (local i32 i32 i32 i32)
    global.get 0
    i32.const 32
    i32.sub
    local.tee 1
    global.set 0
    local.get 1
    i64.const 16384
    i64.store offset=20 align=4
    local.get 1
    i32.const 65536
    i32.store offset=16
    local.get 1
    i32.const 0
    i32.store offset=12
    local.get 1
    i32.const 16
    i32.add
    local.get 1
    i32.const 12
    i32.add
    i32.const 4
    call 6
    block  ;; label = @1
````
Although not sure which function that corresponds to, but maybe there is another ink core library function we can find which is guaranteed to be included in the generated Wasm.

Anyway I would prefer to be more strict and miss some contracts that are written in ink rather than falsely categorising some contracts as having been written in `ink`.

File: crates/analyze/Cargo.toml [hunk 2]
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "contract-analyze"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2021"
+
+[dependencies]
+contract-metadata = { version = "4.0.0-alpha", path = "../metadata" }
+parity-wasm = { version = "0.45.0", features = ["bulk"] }
```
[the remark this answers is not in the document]
smiasojed: While experimenting with the Ask contract on my local environment, I noticed that sometimes opcode 252 is present. That is way I added `bulk` feature, but as the pallet_contracts rejects such code maybe it does not make sense to have it there.

crates/analyze/src/lib.rs
smiasojed: I'm not a fan of this heuristic too. I only included it because the task was defined such way.
I can take a look to see if it could be done in a better way.

crates/analyze/Cargo.toml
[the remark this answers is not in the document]
smiasojed: sign_ext added

crates/analyze/src/lib.rs
smiasojed: I think that basing on function presence can be risky, because inlining optimization can affect our logic.

crates/analyze/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    let start_section = module.start_section();
+    if start_section.is_none()
+        && (is_ink_function_present(&module) || has_function_name(&module, "ink_env"))
+    {
+        return Ok(Language::Ink)
+    } else if start_section.is_none() && has_custom_section(&module, "producers") {
+        return Ok(Language::Solidity)
+    } else if start_section.is_some() && has_custom_section(&module, "sourceMappingURL") {
+        return Ok(Language::AssemblyScript)
+    }
```
ascjones, line 53: If we are reasonably certain that `ink` contracts will not have the custom sections of the `solidity` and `ask`, then we can change the order of this so that the `ink` `if` branch goes last.

This would reduce the possibility of a false positive `ink` identification, since it is still be possible for another type of contract to have a function matching the signature of `deny_payment` or `transferred_value` which has a call to `value_transferred`

ascjones (commented):
I think this is the right direction. We can consider adding a very small custom section to the Wasm for future `ink` versions, except obviously that will appear in every contract so that is the cost.

crates/analyze/src/lib.rs
smiasojed: Good point, I will change the order.

smiasojed: done

ascjones (approved):
LGTM! I think @cmichi should also have a quick look since he originally specified the issue in #784.

PR #1334, opened 2023-09-17
Add contract's ink! compatibility check

### LINKED ISSUE #754: Detect incompatible version of ink
The `2.0.0` release of `cargo-contract` will only be compatible with `4.x` releases of `ink`, so `3.x` releases and earlier will be incompatible.

We should add a check which reads the version of the `ink` dependency in the contract and if it is incompatible, provide the user with a suggestion to either

1. Upgrade their contract to `4.0`
2. Install an earlier `1.x` release of `cargo-contract`

Could also provide a link to docs with a table of compatible `cargo-contract` -> `ink` -> `pallet-contracts` combinations.

smiasojed:
## Add warning message when using incompatible contract's ink! version 
Closes #754

## Description
Checks the compatibility of the contract's ink version with the cargo-contract. If a version mismatch is detected, a warning message will be displayed:
warning: The cargo-contract is not compatible with the contract's ink! version. Please update the cargo-contract to version '1.5.0' or update the contract ink! to a version of '^4.0.0-alpha.3', '^4.0.0', '^5.0.0-alpha'


crates/metadata/src/lib.rs
```diff
@@ -643,6 +652,20 @@ impl ContractBuilder {
     }
 }
 
+/// Checks whether the contract's ink! version is compatible with the cargo-contract
+/// binary.
+pub fn check_contract_ink_compatibility(ink_version: &Version) -> Result<()> {
+    let ink_minimal_version =
+        Version::parse("4.0.0-alpha.1").expect("Parsing version failed");
```
smiasojed, line 659: here dependency is hardcoded,
optionally we could have compatibility.json file:
{
   "versions": [
        {
            "cargo-contract": ">=1.0.0,<2.0.0",
            "ink": ">=3.0.0,<4.0.0"
        },
        {
            "cargo-contract": ">=4.0.0",
            "ink": ">=4.0.0"
        }
    ]
}
 and include it to code (include_str)

crates/metadata/src/compatibility.rs
```diff
@@ -0,0 +1,177 @@
...
+    ink_req: VersionReq,
+}
+
+trait CustomMatch {
+    fn matches_including_pre_releases(&self, version: &Version) -> bool;
+}
+
+// The VersionReq matching mechanism behaves in such a way that 4.0.0-alpha.1 does not
+// satisfy the requirement ">2.0.0,<5.0.0". To modify this behavior, a custom matching
+// implementation has been provided.
```
ascjones, line 52: As discussed, I wonder if it is better just to deal in non pre-release version numbers. So for the configured specs we don't allow pre-release versions, and for the versions of `cargo-contract` and `ink` we just set the `pre` to `None`. WDYT?

smiasojed: The issue with this solution is that information loss can occur if compatibility changes have been introduced between certain pre-release versions, as seen in the example of Ink 4.0.0-alpha3. Therefore, if someone utilizes Ink 4.0.0-alpha1 in their contract, we may mistakenly assume it to be version 4.0.0 and therefore compatible with cargo-contract 2.0.0, when in fact, it is not.

crates/metadata/src/compatibility.rs
```diff
@@ -0,0 +1,177 @@
...
+
+#[derive(Debug, serde::Deserialize)]
+struct CompatibilityList {
+    versions: Vec<CompatibilityInfo>,
+}
+
+#[derive(Debug, serde::Deserialize)]
+struct CompatibilityInfo {
+    #[serde(rename = "cargo-contract")]
+    cargo_contract_req: VersionReq,
```
ascjones, line 41: I think instead of a `VersionReq` this should be a `Version` and we can explicitly enumerate all the `cargo-contract` releases, instead of a range. It forces us to do a compatibility check on this side for each release. This would also solve the pre-release issue for the `cargo-contract` version.

Could have a test which checks the current version has its own entry. You can still easily find the most recent compatible `cargo-contract` release for the warning message.

crates/metadata/src/compatibility.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[serde(rename = "ink")]
+    ink_req: VersionReq,
```
ascjones, line 43: How about making this a `Vec<VersionReq>` and making it a logical OR? Then we can provide additional requirements for prerelease ranges.

```json
{
    "cargo-contract": "2.0.0",
    "ink": [
        "4.0.0-alpha.3",
        "4.0.0"
    ]
}
```

crates/metadata/src/compatibility.rs
ascjones: If we apply the two other suggestions about the spec, we will no longer require the custom matching and can stay true to the semver ranges, which I think is important to avoid confusion.

smiasojed: I think it should work, but the downside is that it will result in higher maintenance costs because for each release, you'll need to add a new entry. Previously, you could skip it until a breaking change was introduced.

crates/metadata/src/compatibility.rs
smiasojed: Done

crates/metadata/src/compatibility.rs
smiasojed: Done

crates/metadata/src/compatibility.rs
```diff
@@ -0,0 +1,161 @@
...
+    // Check if the ink! version matches any of the requirement
+    if !ink_req.iter().any(|req| req.matches(ink_version)) {
+        // Get matching ink! versions
+        let ink_matches = ink_req
+            .iter()
+            .map(|req| format!("'{}'", req))
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        let ink_matches_message = if !ink_matches.is_empty() {
```
ascjones, line 69: If the `ink_req` is at all empty we should throw an error, that is invalid configuration.

crates/metadata/src/compatibility.rs
```diff
@@ -0,0 +1,161 @@
...
+            })
+            .max_by(|&a, &b| {
+                match (!a.pre.is_empty(), !b.pre.is_empty()) {
+                    (false, true) => std::cmp::Ordering::Greater,
+                    (true, false) => std::cmp::Ordering::Less,
+                    (_, _) => a.cmp(b),
+                }
+            })
+            .map(|ver| format!("update the cargo-contract to version '{}'", ver))
+            .unwrap_or_default();
```
ascjones, line 97:
```suggestion
        let best_cargo_contract_version = compatibility
            .cargo_contract_compatibility
            .iter()
            .filter_map(|(ver, reqs)| {
                if reqs
                    .ink_requirements
                    .iter()
                    .any(|req| req.matches(ink_version))
                {
                    return Some(ver)
                }
                None
            })
            .max_by(|&a, &b| {
                match (!a.pre.is_empty(), !b.pre.is_empty()) {
                    (false, true) => std::cmp::Ordering::Greater,
                    (true, false) => std::cmp::Ordering::Less,
                    (_, _) => a.cmp(b),
                }
            });
```

crates/metadata/src/compatibility.rs
```diff
@@ -0,0 +1,161 @@
...
+            return Err(anyhow!(
+                "The cargo-contract is not compatible with the contract's ink! version. Please {}",
+                 matches_message
+            ))
+        } else {
+            bail!(
+                "The cargo-contract is not compatible with the contract's ink! version, \
+                but a matching version could not be found"
+            )
+        }
```
ascjones, line 114: Then this code could be made clearer e.g. 

```rust
if let Some(cargo_contract_version) = best_cargo_contract_version {
   // build message
} else {
  bail!(
     "The cargo-contract is not compatible with the contract's ink! version, \
      but a matching version could not be found"
  )
}
```

Or even use `ok_or(bail!("The cargo-contract is not compatible ..."))?` on the end of the `best_cargo_contract_version` expression I suggest.

ascjones (commented):
Looks good, a few improvements can be made to the readability of the comparison logic :point_down:

crates/metadata/src/compatibility.rs
smiasojed: done

crates/metadata/src/compatibility.rs
smiasojed: done

crates/metadata/src/compatibility.rs
smiasojed: done

ascjones (approved):
LGTM

PR #1347, opened 2023-09-27
Add total contract deposit

smiasojed:
## Summary
Closes #_
- [y] y/n | Is it dependent on the specific version of `ink` or `pallet-contracts`?
Add total contract deposit

## Description
Added functionality to contract info command, which calculates and displays storage total deposit:
```
                TrieId 6ec258b9d34918154e689c37920ee3ff6d1afb0bb69938e67853092ce0fced18
             Code Hash 0x2de75ab18bfb8e24c968e8ff0a76d40e49ace69eb9e95d814c1cc4232be8c8eb
         Storage Items 1
  Storage Item Deposit 100000000000
 Storage Total Deposit 200770000000
 ```

### CODE REVIEW

File: crates/extrinsics/src/lib.rs [hunk 1]
```diff
@@ -303,6 +303,44 @@ pub fn parse_code_hash(input: &str) -> Result<<DefaultConfig as Config>::Hash> {
...
+) -> Result<Balance> {
+    let contract_pallet_version = fetch_contracts_pallet_version(client).await?;
+    let mut contract_deposit = storage_base_deposit
+        .saturating_add(storage_item_deposit)
+        .saturating_add(storage_byte_deposit);
+
+    // From contracts pallet version 10 deposit calculation has changed.
+    if contract_pallet_version >= 10 {
+        let existential_deposit_address =
+            api::constants().balances().existential_deposit();
```
smiasojed, line 321: This call introduces a dependency on the Balances pallet. The drawback is that we now have an additional dependency in the metadata, which has caused its size to grow.

crates/extrinsics/src/lib.rs
```diff
@@ -303,6 +303,44 @@ pub fn parse_code_hash(input: &str) -> Result<<DefaultConfig as Config>::Hash> {
...
+
+/// Fetch the contracts pallet version from the storage using the provided client.
+async fn fetch_contracts_pallet_version(client: &Client) -> Result<u16> {
+    let hash_pallet = hashing::twox_128(b"Contracts");
+    let hash_version = hashing::twox_128(b":__STORAGE_VERSION__:");
+    let key = [hash_pallet, hash_version].concat();
+
+    let version = client
+        .rpc()
+        .storage(key.as_slice(), None)
```
smiasojed, line 336: Raw access to the storage

crates/extrinsics/src/lib.rs
SkymanOne: I see no problem with that. We are not particularly concerned with the binary sized of `cargo-contract`

crates/extrinsics/src/lib.rs
```diff
@@ -303,6 +303,44 @@ pub fn parse_code_hash(input: &str) -> Result<<DefaultConfig as Config>::Hash> {
...
+    storage_byte_deposit: Balance,
+    client: &Client,
+) -> Result<Balance> {
+    let contract_pallet_version = fetch_contracts_pallet_version(client).await?;
+    let mut contract_deposit = storage_base_deposit
+        .saturating_add(storage_item_deposit)
+        .saturating_add(storage_byte_deposit);
+
+    // From contracts pallet version 10 deposit calculation has changed.
+    if contract_pallet_version >= 10 {
```
SkymanOne, line 319: Need to display an error to the user otherwise

crates/extrinsics/src/lib.rs
SkymanOne: Isn't there an RPC method to fetch a pallet version?

crates/extrinsics/src/lib.rs
smiasojed: For the pallet-contracts version >=8, it works. The issue may arise for pallets-contract versions <8, as it will result in an error during the scale decoding of ContractInfo struct, which has changed at this time. It's important to note that this limitation is not introduced by this commit.

crates/extrinsics/src/lib.rs
smiasojed: I have not found it

crates/cargo-contract/src/cmd/mod.rs
```diff
@@ -230,8 +230,13 @@ pub fn basic_display_format_extended_contract_info(info: &ExtendedContractInfo)
...
     );
     name_value_println!(
-        "Storage Deposit",
-        format!("{:?}", info.storage_item_deposit),
+        "Storage Items Deposit",
+        format!("{:?}", info.storage_items_deposit),
+        MAX_KEY_COL_WIDTH
+    );
+    name_value_println!(
+        "Storage Total Deposit",
```
ascjones, line 238: Could use the `STORAGE_DEPOSIT_KEY` here?

crates/extrinsics/src/lib.rs
```diff
@@ -347,29 +384,58 @@ pub async fn fetch_contract_info(
...
-            let convert_trie_id = hex::encode(info_result.trie_id.0);
-            Ok(Some(ContractInfo {
-                trie_id: convert_trie_id,
-                code_hash: info_result.code_hash,
-                storage_items: info_result.storage_items,
-                storage_item_deposit: info_result.storage_item_deposit,
-            }))
+    let total_balance: Balance = match contract_info.as_type::<DepositAccount>() {
+        Ok(account) => {
+            // StorageVersion >= 10 and < 15 contains deposit in separate account
```
ascjones, line 422: This comment could be expanded to explain a bit what is going on here, I took me a little bit to figure it out. I understand it as: newer versions of `pallet_contracts` have an additional field `deposit_account`, so we first attempt to decode that and get the balance for that account. If that fails we assume that it is the older version without that field. Then I am assuming that on line 431 this decoding will succeed for both cases because the `contract_info.as_type::<DepositAccount>()` will have already consumed the additional field if present?

Anyway it would be a nice present for future maintainers since there are no tests for this code.

ascjones: Maybe we could add a unit test for this part of the logic, if we extract a pure function e.g. `fn get_deposit_account_id(contract_info: DecodedValueThunk) -> Result<AccountId32>`

crates/extrinsics/src/lib.rs
```diff
@@ -347,29 +384,58 @@ pub async fn fetch_contract_info(
...
+            let deposit_account = AccountId32(account.deposit_account.0);
```
ascjones, line 423:
```suggestion
            let deposit_account = &account.deposit_account;
```
Don't need to construct a new `AccountId32`?

crates/extrinsics/src/lib.rs
smiasojed: > This comment could be expanded to explain a bit what is going on here, I took me a little bit to figure it out. I understand it as: newer versions of `pallet_contracts` have an additional field `deposit_account`, so we first attempt to decode that and get the balance for that account. If that fails we assume that it is the older version without that field. Then I am assuming that on line 431 this decoding will succeed for both cases because the `contract_info.as_type::<DepositAccount>()` will have already consumed the additional field if present?
> 
> Anyway it would be a nice present for future maintainers since there are no tests for this code.

Versions [>=10, <15] have additional field `deposit_account` in `ContractInfoOf` struct. If it exists the contract's deposit is there, otherwise contract's deposit is on the main contract's account.

Regarding decoding on the line 431: It succeeds for both case, because we decode only common part of data.

smiasojed: > DecodedValueThunk

I would like to do it, but I do not see the public API to create DecodedValueThunk object

ascjones: Hm I guess you have to construct it with https://github.com/paritytech/subxt/blob/43809ee5928a77eed5438f7d61f24fe0183f970a/subxt/src/dynamic.rs#L44. Anyway if it is too much trouble then no need to bother, but would be nice for this to be tested.

smiasojed: I will add it, thanks

crates/extrinsics/src/contract_info.rs
```diff
@@ -32,45 +32,128 @@ use scale::Decode;
...
+    };
+
+    let info = contract_info.as_type::<ContractInfoOf>()?;
+    Ok(ContractInfo {
+        trie_id: hex::encode(info.trie_id.0),
+        code_hash: info.code_hash,
+        storage_items: info.storage_items,
+        storage_items_deposit: info.storage_item_deposit,
+        storage_total_deposit: total_balance,
+    })
```
ascjones, line 147: I actually meant to extract and test this whole block, to see that it works for both versions of the metadata.

ascjones: Suggestion for this here: https://github.com/paritytech/cargo-contract/pull/1400/files

smiasojed: Done

crates/extrinsics/src/contract_info.rs
```diff
@@ -32,45 +32,135 @@ use scale::Decode;
...
-            let convert_trie_id = hex::encode(info_result.trie_id.0);
-            Ok(Some(ContractInfo {
-                trie_id: convert_trie_id,
-                code_hash: info_result.code_hash,
-                storage_items: info_result.storage_items,
-                storage_item_deposit: info_result.storage_item_deposit,
-            }))
+/// Enum representing different types of contract info, distinguishing between
+/// main account deposit and secondary account deposit.
+pub enum ContractInfoRaw {
```
ascjones, line 102:
```suggestion
enum ContractInfoRaw {
```
Only used internally here.

crates/extrinsics/src/contract_info.rs
```diff
@@ -32,45 +32,135 @@ use scale::Decode;
...
+    MainAccountDeposit(AccountId32, ContractInfoOf),
+    SecondaryAccountDeposit(AccountId32, ContractInfoOf),
```
ascjones, line 104: IMO since both variants have the same fields this could just be a struct with an extra field as an enum (or even just a bool) which then avoids having the matches to extract the fields. Not a masive deal since this is encapsulated here anyway.

crates/extrinsics/src/contract_info.rs
```diff
@@ -161,3 +256,207 @@ pub async fn fetch_all_contracts(
...
+
+/// A struct representing `Vec`` used in the storage reads.
+#[derive(Debug, DecodeAsType)]
+#[decode_as_type(crate_path = "subxt::ext::scale_decode")]
+struct BoundedVec<T>(pub ::std::vec::Vec<T>);
+
+/// A struct used in the storage reads to access contract info.
+#[derive(Debug, DecodeAsType)]
+#[decode_as_type(crate_path = "subxt::ext::scale_decode")]
+pub struct ContractInfoOf {
```
ascjones, line 283: I guess this should also not be `pub` (and `AccountData`)?

crates/extrinsics/src/contract_info.rs
smiasojed: done

ascjones (approved):

Issue #672, opened 2022-08-04
Add flag `--wait-for-finalization`. so that success/events are not displayed until the block is finalized


ascjones:
Extracted from #432

Currently the results of a transaction are displayed as soon as the block is produced containing that transaction. However for extra security the user may want to wait until the block is finalized, which takes a lot longer.

PR #1352, opened 2023-09-29
Upgrade `subxt` to `0.32.0`

ascjones:
## Summary

Update to latest subxt release https://github.com/paritytech/subxt/pull/1178

## Description

Main changes are to the storage api for fetching all contract infos, and using the legacy rpc client to fetch system properties for token metadata.

### best block vs finalized block

The latest release of subxt has changed the default of `at_latest` for reading storage changes to the latest finalized block, where previously it was the best block. Because `cargo-contract` extrinsics feature is intended primarily for use during development, and because we use `substrate-contracts-node` for testing (which does not have finalization), in this PR I have changed all storage reads to read for the latest best block to maintain the current behaviour.

In the future we probably want to add the ability to read from the latest finalized flag, either by default or by way of a flag, see also https://github.com/paritytech/cargo-contract/issues/672


crates/extrinsics/src/lib.rs
```diff
@@ -403,28 +455,28 @@ fn parse_contract_account_address(
...
 
-    let contracts = keys
-        .into_iter()
-        .map(|e| parse_contract_account_address(&e.0, key.len()))
-        .collect::<Result<_, _>>()?;
+    let best_block = get_best_block(rpc).await?;
+    let mut contracts = client.storage().at(best_block).iter(query).await?;
+
+    let mut contract_accounts = Vec::new();
+    while let Some(result) = contracts.next().await {
```
smiasojed, line 468: Is there a "bulk fetch" function, which we could use to speed up the execution?

crates/cargo-contract/src/cmd/info.rs
```diff
@@ -71,25 +72,14 @@ pub struct InfoCommand {
...
     pub async fn run(&self) -> Result<(), ErrorVariant> {
-        let client =
-            OnlineClient::<DefaultConfig>::from_url(url_to_string(&self.url)).await?;
+        let rpc = RpcClient::from_url(url_to_string(&self.url)).await?;
+        let client = OnlineClient::<DefaultConfig>::from_rpc_client(rpc.clone()).await?;
 
         // All flag applied
         if self.all {
             // 1000 is max allowed value
             const MAX_COUNT: u32 = 1000;
```
smiasojed, line 81: MAX_COUNT should be removed. It is not needed, as we do not use the response paging anymore.

crates/extrinsics/src/lib.rs
```diff
@@ -403,28 +455,28 @@ fn parse_contract_account_address(
 /// requested elements starting from an optional address
 pub async fn fetch_all_contracts(
```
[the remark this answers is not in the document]
ascjones: Yes I had changed it to `iter()` which fetches the full value as well as the key. I didn't realise `fetch_keys` had become `fetch_raw_keys` so I've changed it to that,

crates/extrinsics/src/lib.rs
```diff
@@ -403,28 +455,28 @@ fn parse_contract_account_address(
...
     client: &Client,
+    rpc: RpcClient,
     count: u32,
```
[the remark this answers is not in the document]
ascjones: Removed :+1:

crates/extrinsics/src/lib.rs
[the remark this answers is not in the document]
ascjones: Tested, it's back to 3s again

crates/extrinsics/src/lib.rs
ascjones: In the background it is doing pagination so every iteration does not result in a network request. https://github.com/paritytech/subxt/blob/cf7e2db1b7b57385d2e0bb0e49169bd6627f6811/subxt/src/backend/legacy/mod.rs#L342

Anyway since I switched it to `fetch_keys` it is now back down to 3s.

crates/cargo-contract/src/cmd/info.rs
ascjones: Done :+1:

crates/extrinsics/src/call.rs
```diff
@@ -204,10 +211,12 @@ pub struct CallExec {
     gas_limit: Option<u64>,
     proof_size: Option<u64>,
     value: BalanceVariant,
+    rpc: RpcClient,
```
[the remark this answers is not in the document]
ascjones: Yes that's much better. Done :+1:

crates/extrinsics/src/call.rs
```diff
@@ -221,23 +230,21 @@ impl CallExec {
...
             origin: account_id(&self.signer),
             dest: self.contract.clone(),
-            value: self.value.denominate_balance(&token_metadata)?,
+            value: self.value.denominate_balance(&self.token_metadata)?,
             gas_limit: None,
             storage_deposit_limit,
             input_data: self.call_data.clone(),
         };
-        state_call(&url, "ContractsApi_call", call_request).await
+        state_call(self.rpc.clone(), "ContractsApi_call", call_request).await
```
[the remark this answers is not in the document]
ascjones: Done :+1:

crates/extrinsics/src/lib.rs
```diff
@@ -310,17 +347,29 @@ pub fn parse_code_hash(input: &str) -> Result<<DefaultConfig as Config>::Hash> {
...
+    let rpc = LegacyRpcMethods::<DefaultConfig>::new(rpc);
+    rpc.chain_get_block_hash(None)
+        .await?
+        .ok_or(subxt::Error::Other("Best block not found".into()))
+}
+
 /// Fetch the contract info from the storage using the provided client.
 pub async fn fetch_contract_info(
     contract: &AccountId32,
+    rpc: RpcClient,
```
smiasojed, line 363: could be `rpc: &LegacyRpcMethods::<DefaultConfig>`

crates/extrinsics/src/lib.rs
```diff
@@ -310,17 +347,29 @@ pub fn parse_code_hash(input: &str) -> Result<<DefaultConfig as Config>::Hash> {
     Ok(arr.into())
 }
 
+/// Fetch the hash of the *best* block (included but not guaranteed to be finalized).
+async fn get_best_block(
+    rpc: RpcClient,
```
smiasojed, line 352: maybe we should align each function to use LegacyRpc `rpc: &LegacyRpcMethods::<DefaultConfig>`

crates/extrinsics/src/lib.rs
```diff
@@ -410,28 +460,28 @@ fn parse_contract_account_address(
 /// requested elements starting from an optional address
 pub async fn fetch_all_contracts(
     client: &Client,
-    count: u32,
-    count_from: Option<&AccountId32>,
+    rpc: RpcClient,
```
smiasojed, line 463: could be rpc: &LegacyRpcMethods::<DefaultConfig>

smiasojed (approved):

PR #1359, opened 2023-10-06
Use new `account_nonce` API from subxt `0.32.1`

ascjones:
Quick follow up to #1352 to use the patch release and the new `account_nonce` method to replace the copied impl.


SkymanOne (approved):

smiasojed (approved):
LGTM

PR #1377, opened 2023-10-20
Compare `Environment` types against the node

### LINKED ISSUE #1167: Before contract upload/instantiate: verify that node types match contract types
With https://github.com/paritytech/ink/issues/1695 the environmental types of a contract are now persisted to its metadata. Before a contract is uploaded/instantiated to a chain we should conduct a verification that the contract's types actually match the ones from the node metadata.

SkymanOne:
## Summary
Closes #1167 
- [y] y/n | Is it dependent on the specific version of `ink` or `pallet-contracts` - ink! `5.0.0-alpha` & `polkadot-1.2.0` ?

Introduces check for a type match between the contract environment and the target node.

## Description

This PR allows checking environmental type match before execution `call`, `upload` and `instantiate` including dry-running.


crates/transcode/src/lib.rs
```diff
@@ -273,6 +279,40 @@ impl ContractMessageTranscoder {
     fn messages(&self) -> impl Iterator<Item = &MessageSpec<PortableForm>> {
         self.metadata.spec().messages().iter()
     }
+    /// Compares the contract's environment type with a provided type definition.
+    pub fn compare_type(
```
ascjones, line 283: This method belongs better together with the rest in `env_check` rather than in the transcoder. We could just add a getter to access `self.metadata`.

crates/transcode/src/env_check.rs
ascjones, line 1: If we move `compare_type` here, we could move this entire file into the `extrinsics` crate from where it is used, and it doesn't need then to be a public API.

crates/transcode/src/env_check.rs
```diff
@@ -0,0 +1,89 @@
...
+) -> Result<TypeDef<PortableForm>> {
+    let tt = registry
+        .resolve(id)
+        .context("Type is not present in registry")?;
+    if tt.type_params.is_empty() {
+        if let TypeDef::Composite(comp) = &tt.type_def {
+            println!("Resolve type definition: {:#?}", tt);
+            let tt_id = comp
+                .fields
+                .get(0)
```
ascjones, line 53: We should check for invalid composites which have more than one field. Maybe a test for that?

crates/transcode/src/env_check.rs
```diff
@@ -0,0 +1,89 @@
...
+            .get(0)
+            .context("type param is not present")?
+            .ty
+            .context("concrete type is not present")?
+            .id;
+        resolve_type_definition(registry, param_id)
+    }
+}
+
+pub fn compare_node_env_with_contract(
```
ascjones, line 72: Would be good to have some unit tests for this.

crates/transcode/src/env_check.rs
SkymanOne: Shouldn't it be a public API?

ascjones: Not unless it needs to be

SkymanOne: If we move it to `extrinsics` then `ink_metadata` crate also needs to be imported there. I'd rather avoid duplicating dependencies. `transcode` is responsible for working with ink metadata, so that's my reasoning for `env_check` to stay there.

ascjones: The responsibility of `transcode` is to convert strings into SCALE encoded data, and vice versa, *using* the metadata.

`ink_metadata` is already an indirect dependency of `extrinsics` (via `transcode`), so I don't see a problem with making it a direct dependency.

crates/transcode/Cargo.toml
```diff
@@ -37,6 +37,7 @@ serde = { version = "1.0.189", default-features = false, features = ["derive"] }
 serde_json = "1.0.107"
 thiserror = "1.0.50"
 strsim = "0.10.0"
+subxt = "0.32.1"
```
ascjones, line 40: Can't see this used anywhere

crates/transcode/src/env_check.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+}
+
+impl std::error::Error for EnvCheckError {}
+
+impl From<anyhow::Error> for EnvCheckError {
+    fn from(value: anyhow::Error) -> Self {
+        Self::RegistryError(value.to_string())
+    }
+}
```
ascjones, line 45: Could get rid of all of this if we just used `anyhow::Error` at the top level. You are already using that anyway on the internal methods.

crates/transcode/src/env_check.rs
```diff
@@ -0,0 +1,89 @@
...
+    Result,
+};
+
+use crate::ContractMessageTranscoder;
+
+#[derive(Debug, Clone)]
+pub enum EnvCheckError {
+    CheckFailed(String),
+    RegistryError(String),
+}
```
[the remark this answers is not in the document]
ascjones: Hmm maybe best to use `anyhow::Error` directly, see other comment.

crates/transcode/src/env_check.rs
ascjones: I believe this code should belong in the `extrinsics` crate, from where it is called.

crates/transcode/src/env_check.rs
SkymanOne: I want to have a concrete error for the failed field check

ascjones: Why though? It just gets converted immediately back into an `anyhow::Error` in the only (non test) usage: https://github.com/paritytech/cargo-contract/blob/9ebfa62c7d0ddc6e252461cddee21f00fe7d5845/crates/extrinsics/src/lib.rs#L383.

And for all the other errors you will have lost all the magic from the context and error chaining by converting the `anyhow::Error` to a `String` in `Self::RegistryError(value.to_string())`.

crates/extrinsics/src/env_check.rs
```diff
@@ -0,0 +1,371 @@
...
+    Context,
+    Result,
+};
+
+fn get_node_env_fields(registry: &PortableRegistry) -> Result<Vec<Field<PortableForm>>> {
+    let env_type = registry
+        .types
+        .iter()
+        .find(|t| t.ty.path.segments.contains(&"Environment".to_string()))
+        .context("The node does not contain `Environment` type. Are you using correct `pallet-contracts` version?")?;
```
ascjones, line 19: I wonder if we should just print a warning in this case, in order to still support older versions of `pallet-contracts`?

ascjones (approved):
LGTM, Just a question of whether we should warn instead of error in the case above.

crates/extrinsics/src/env_check.rs
```diff
@@ -0,0 +1,371 @@
...
+use anyhow::{
+    Context,
+    Result,
+};
+
+fn get_node_env_fields(registry: &PortableRegistry) -> Result<Vec<Field<PortableForm>>> {
+    let env_type = registry
+        .types
+        .iter()
+        .find(|t| t.ty.path.segments.contains(&"Environment".to_string()))
```
ascjones, line 18: I had an idea to make this more accurate, and use the full namespace here and have it work in the tests, we could use the recently added feature: https://github.com/paritytech/scale-info/pull/197

SkymanOne: Good one

crates/extrinsics/src/env_check.rs
SkymanOne: Not sure how this would be implemented. We need to make the overall check optional then

ascjones: Move this check to `compare_node_env_with_contract`, print the warning then return `Ok(())`.

crates/extrinsics/src/env_check.rs
```diff
@@ -11,18 +11,24 @@ use anyhow::{
...
+        t.ty.path.segments[len - 2..] == ["pallet_contracts", "Environment"]
+    });
+
+    // if we can't find the type, then we use the old contract version.
+    if env_type_option.is_none() {
+        tracing::warn!("The node does not contain `Environment` type. Are you using correct `pallet-contracts` version?");
+        return Ok(None)
+    }
+
+    let env_type = env_type_option.unwrap();
```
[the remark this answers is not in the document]
SkymanOne: You can, but I want to avoid if nesting, makes easier to read code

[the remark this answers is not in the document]
ascjones: To clarify, what I meant was with `let else` you can do the early return inside the conditional.

```
if let Some(env_type) = env_type_option else {
  // if we can't find the type, then we use the old contract version.
        tracing::warn!("The node does not contain `Environment` type. Are you using correct `pallet-contracts` version?");
        return Ok(None)
}

// carry on as normal
``

File: crates/extrinsics/src/env_check.rs [hunk 10]
```diff
@@ -0,0 +1,384 @@
...
+
+fn get_node_env_fields(
+    registry: &PortableRegistry,
+) -> Result<Option<Vec<Field<PortableForm>>>> {
+    let Some(env_type) = registry.types.iter().find(|t| {
+        let len = t.ty.path.segments.len();
+        t.ty.path.segments[len - 2..] == ["pallet_contracts", "Environment"]
+    }) else {
+        // if we can't find the type, then we use the old contract version.
+        contract_build::verbose_eprintln!(Verbosity::Verbose,
```
ascjones, line 23: if you're not passing in the verbosity you might just as well use `eprintln` directly

```suggestion
        eprintln!(
```

smiasojed (approved):

PR #1380, opened 2023-10-25
Do not use `wasm-opt` default features

xermicus:
The most recent version of `wasm-opt` added a feature flag for supporting binaryens DWARF passes. I'm not aware that we need those, and because this will introduce namespace conflicts with LLVM, [if the crate is used together with other LLVM crates](https://github.com/brson/wasm-opt-rs/issues/154), we should consider deactivating them.

### DIFF
--- a/crates/build/Cargo.toml
+++ b/crates/build/Cargo.toml
@@ -35,7 +35,7 @@ serde_json = "1.0.107"
 tempfile = "3.8.0"
 term_size = "0.3.2"
 url = { version = "2.4.1", features = ["serde"] }
-wasm-opt = "=0.116.0"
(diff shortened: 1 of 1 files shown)


ascjones (approved):

tareknaser:
Hello @ascjones

I'm looking forward to incorporating the `contract-extrinsics` crate into my project. Currently, I'm using it as a git dependency. 
I was wondering if there are any plans to publish it in the near future?

Thank you!

ascjones:
Hi @tareknaser, it is already on crates.io here https://crates.io/crates/contract-extrinsics. Although it is a few months old now, we will release a new version by December.

Issue #101, opened 2020-11-09
Storage inspection command


ascjones:
Contract metadata describes the storage in the `layout` section. A `storage` command should be introduced which utilizes this to allow inspection of a contract's storage from the command line.

So for the `flipper`:

```
#[ink::contract]
pub mod flipper {
    #[ink(storage)]
    pub struct Flipper {
        value: bool,
    }
}
```
Calling `cargo contract storage --contract <contract_account>` would result in the following:

```
Flipper { value: false }
```
This would work by using the metadata to generate keys and query the `get_storage` rpc: https://github.com/paritytech/substrate/blob/30ecb9b259140a860b20c662fda01b5b2c33e5fa/frame/contracts/rpc/runtime-api/src/lib.rs#L54.

The metadata would then be used to decode the results into human readable form, building on the "SCON" work done in #79.

For an initial implementation I think we don't need to do everything in a single PR, we can define a basic implementation that is useful and then add features as follow up PRs.

### Lazy and Collections

We would need to consider how to handle the built-in lazy and collection types, they would probably require some custom display code.

### Keys

It might be useful to be able to augment the data with the actual storage keys, and also query individual storage keys either by the raw key or by name.

### Raw Storage

For debugging of corrupt storage it would be useful to be able to inspect the raw contract storage (without decoding).

Robbepop:
I added the `blocked` label since this feature depends on some changes made in ink!'s `layout` metadata section. The current metadata format cannot represent some more or less common scenarios and really needs some restructuring. When I tried some weeks ago I immediately hit a wall for recursive data structures.

We first have to define semantics, definitions and structures for a new and better `layout` metadata section before we should tackle this issue. We have to think about cases like:

- recursively defined data structures such as `enum Tree { Leaf(i32), Node(Box<Tree>, Box<Tree>) }` or in case of `ink_storage`: `ink_storage::Vec<ink_storage::Vec<i32>>`, packed elements where some elements actually have some further pointers `Pack<(i32, ink_storage::Box<i32>)` and more.

Maybe we have to restrict the final metadata to not support certain use cases because their metadata format would become too complex otherwise.

cmichi:
Whoever implements this should take a look at the CLI layout that Truffle uses. I have only this shaky photo from a talk:

![ima_b84ccbc](https://user-images.githubusercontent.com/241530/222223945-c2bf637e-f336-47c0-a986-0513d390774c.jpeg)

The layout is quite nice and clean.

[...]

PR #1395, opened 2023-11-10
Basic storage inspection command

ascjones:
This provides a foundation for https://github.com/paritytech/cargo-contract/issues/101. 

It queries all the storage entries for a contract, and then associates the entries with root keys from the contract metadata storage layout, printing the results as JSON for now:

image

Also can display the raw storage keys/values if specified or if no contract metadata available:

image

For now the structure of the cells is flat, and only the raw data for the cells is shown. To continue on from this we should:

- Refine and possible structure the output types (e.g. grouping per root key)
- Option for displaying non-json user friendly view of data (see Michi's picture in #101)
- Decode the raw data using metadata from the contract.


smiasojed:
I think we should update the CHANGELOG.md and the usage section of README.md file

ascjones:
> I think we should update the CHANGELOG.md and the usage section of README.md file

Done, but only a basic entry in the README, can be updated once the command is more fleshed out.

crates/extrinsics/src/contract_artifacts.rs
```diff
@@ -142,6 +143,20 @@ impl ContractArtifacts {
         })
     }
 
+    /// Get the deserialized [`InkProject`] metadata.
+    ///
+    /// ## Errors
+    /// - No contract metadata could be found.
+    /// - Invalid contract metadata.
+    pub fn ink_project_metadata(&self) -> Result<InkProject> {
```
smiasojed, line 151: I think we should consider updating the 'scheme' command to use 'contract_artifacts'. There is some code duplication.

crates/extrinsics/src/contract_storage.rs
```diff
@@ -0,0 +1,312 @@
...
+
+pub struct ContractStorage<C: Config = DefaultConfig> {
+    rpc: ContractStorageRpc<C>,
+}
+
+impl<C: Config> ContractStorage<C>
+where
+    C::AccountId: AsRef<[u8]> + Display + IntoVisitor,
+    DecodeError: From<<<C::AccountId as IntoVisitor>::Visitor as Visitor>::Error>,
+    BlockRef<sp_core::H256>: From<C::Hash>,
```
smiasojed, line 67: Maybe we should just return BlockRef<sp_core::H256> from get_best_block until we figure out how to make it cleaner.

crates/extrinsics/src/contract_storage.rs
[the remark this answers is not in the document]
ascjones: It's a good idea, let's do it as a follow up though.

crates/extrinsics/src/contract_artifacts.rs
ascjones: I hadn't even looked at the `schema` command, but just looked and you are absolutely right it could/should use `ContractArtifacts`. Should be a separate PR though.

crates/extrinsics/src/contract_storage.rs
SkymanOne: I don't see the problem with the current approach using generics

crates/extrinsics/src/contract_storage.rs
```diff
@@ -0,0 +1,312 @@
...
+    ) -> Result<ContractStorageLayout> {
+        let data = self.load_contract_storage_data(contract_account).await?;
+        let layout = ContractStorageLayout::new(data, metadata.layout());
+        Ok(layout)
+    }
+}
+
+/// Represents the raw key/value storage for the contract.
+#[derive(Serialize)]
+pub struct ContractStorageData(BTreeMap<Bytes, Bytes>);
```
SkymanOne, line 117: Why do you need a dedicated structure for this if it only contains a single field, and has no methods?

crates/extrinsics/src/contract_storage.rs
```diff
@@ -0,0 +1,312 @@
...
+                for (variant, struct_layout) in enum_layout.variants() {
+                    for field in struct_layout.fields() {
+                        let label =
+                            format!("{}::{}", enum_layout.name(), variant.value());
+                        Self::collect_root_keys(label, field.layout(), root_keys);
+                    }
+                }
+            }
+            Layout::Array(_) => {
+                todo!("Figure out what to do with an array layout")
```
SkymanOne, line 175: What's the context?

crates/extrinsics/src/contract_artifacts.rs
SkymanOne: Yh, it makes sense to reuse this method for schema operations

SkymanOne:
Also worth looking at https://docs.rs/csv/latest/csv/cookbook/index.html#writing-basic to display values in CSV format

crates/cargo-contract/src/cmd/storage.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[derive(Debug, clap::Args)]
+#[clap(name = "storage", about = "Inspect contract storage")]
+pub struct StorageCommand {
+    /// The address of the contract to inspect storage of.
+    #[clap(
+        name = "contract",
+        long,
+        env = "CONTRACT",
+        required_unless_present = "all"
```
smiasojed, line 40: I think, this line is not needed

crates/extrinsics/src/contract_storage.rs
ascjones: I've removed these bounds now, it was because I was still using the `pub type Client = OnlineClient<DefaultConfig>` alias. I have switched that to using the generic `OnlineClient` which has fixed it.

crates/cargo-contract/src/cmd/storage.rs
ascjones: Yes, not needed. Copypasta

crates/extrinsics/src/contract_storage.rs
```diff
@@ -0,0 +1,310 @@
...
+        (root_key, mapping_key)
+    }
+}
+
+#[derive(Serialize)]
+pub struct ContractStorageCell {
+    key: Bytes,
+    value: Bytes,
+    label: String,
+    root_key: u32,
```
smiasojed, line 211: Maybe it would be more useful to dispaly the `root_key` as hex?

ascjones: Possibly, yes - that is how they are serialized in the metadata.

Anyway I will leave that up to you if that is okay? I expect this data structure will not be the final form of how we display the structured storage output.

smiasojed (approved):
LGTM

PR #1397, opened 2023-11-13
Prevent submitting calls to messages marked as immutable

xermicus:
## Summary

Related issue: https://github.com/paritytech/ink/issues/1969

`ink!` can't properly prevent messages supposed to be immutable from changing state regardless. Instead, we rely on the client and tooling to honor the `mutates` field in the metadata.

## Description
Without this change, cargo contract will happily execute messages like the one below. This message receives `&self` which signals it should be immutable. Other contract languages, e.g. Solidity, would refuse to compile this. Instead, we rely on the client to prevent it. The rogue balance transfer is just an illustration; Things like that might come from a rogue snippet or sit inside a rogue library.

```rust
        #[ink(message)]
        pub fn get(&self) -> bool {
            // oops
            self.env()
                .transfer(self.env().caller(), self.env().balance())
                .unwrap();

            self.value
        }
```

### DIFF
--- a/crates/extrinsics/src/call.rs
+++ b/crates/extrinsics/src/call.rs
@@ -15,13 +15,31 @@
 // along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
 
 use super::{
-    account_id, events::DisplayEvents, runtime_api::api, state, state_call,
-    submit_extrinsic, AccountId32, Balance, BalanceVariant, Client,
-    ContractMessageTranscoder, DefaultConfig, ErrorVariant, Missing, TokenMetadata,
+    account_id,
+    events::DisplayEvents,
+    runtime_api::api,
+    state,
+    state_call,
+    submit_extrinsic,
+    AccountId32,
+    Balance,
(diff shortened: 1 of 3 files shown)


crates/extrinsics/src/call.rs
```diff
@@ -267,6 +267,23 @@ impl CallExec {
...
+            .metadata()
+            .spec()
+            .messages()
+            .iter()
+            .find(|msg| msg.label() == &self.message)
+            .expect("message exist after calling CallExec::done()")
+            .mutates()
+        {
+            let inner = anyhow!(
+                "Tried to execute a call to immutable contract message '{}'. Please do a dry-run instead.",
```
SkymanOne, line 281:
```suggestion
                "Tried to execute a call on the immutable contract message '{}'. Please do a dry-run instead.",
```

or 

```suggestion
                "Tried to submit a call to the immutable contract message '{}'. Please do a dry-run instead.",
```

SkymanOne (approved):

Issue #1168, opened 2023-06-21
Support multiple `pallet-contracts` versions to interact with


cmichi:
`cargo-contract` currently contains a fixture of the node's metadata ([the file is here](https://github.com/paritytech/cargo-contract/blob/master/crates/cargo-contract/src/cmd/runtime_api/contracts_runtime.scale)). This metadata file includes the `pallet-contracts` metadata. If users try to interact with a node that has a different metadata version from the one in their local `cargo-contract` installation, this error will be thrown:

```
ERROR: Metadata: Pallet Contracts Call Contracts has incompatible metadata
```

So currently `cargo-contract` can only target one version of `pallet-contracts` and we should fix this.

We reference the fixture here:
https://github.com/paritytech/cargo-contract/blob/d9950ca212a22d491fb61a022f550361280bbc79/crates/cargo-contract/src/cmd/runtime_api/mod.rs#L20

`subxt` by now supports downloading the node metadata automatically. The API is `#[subxt::subxt(runtime_metadata_url = "wss://rpc.polkadot.io:443")]`. But we don't know the endpoint at compile time, since the user can supply a custom `--url`. 

## The Solution

Modify the `subxt` setup to use dynamic values instead. This same issue has actually already been solved for the ink! end-to-end testing framework, [here is the commit](https://github.com/paritytech/ink/pull/1526/commits/2648f6acd9517a88a6ad7e8d0d45ffd6d3da90af). It should be possible to take that commit as a template and transfer the solution to `cargo-contract`.

ascjones:
Copying my response from https://github.com/paritytech/cargo-contract/pull/1186#issuecomment-1613340426 regarding using different environment types for `AccountId`:

====

Until now we have only targeted a single node configuration, the recent `substrate-contracts-node`, so both the `Environment` types (`AccountId`, `Balance` etc.) are fixed and the generated types and call API (from `subxt` metadata) are fixed against this target node.

What we want is to be able to target different node configurations. I see a couple of different approaches:

1. Make the `Environment` types generic everywhere they are used, maintain multiple versions of the `subxt` metadata for each type of node. At runtime detect which type of node we are targeting and invoke the command with the concrete `Environment` types configuration
  1a. Alternatively replace the subxt metadata with handcrafted calls with generic parameters for `Environment` types
2. Introduce a plugin model whereby we can support any number of configurations/versions of chains, you just need to install a plugin which conforms to the given API for the supported commands.

[...]

PR #1399, opened 2023-11-13
Add dynamic types support

smiasojed:
## Summary
Related #1168
Add dynamic types support

## Description
Remove the need for compile-time node metadata stored in extrinsics/src/runtime_api/metadata.scale, and update the `subxt` setup to utilize dynamic values instead of static.
Note that the metadata is retained solely for testing purposes.

- [y] My code follows the style guidelines of this project
- [y] I have added an entry to `CHANGELOG.md`
- [y] I have commented my code, particularly in hard-to-understand areas
- [n] I have added tests that prove my fix is effective or that my feature works
- [n] Any dependent changes have been merged and published in downstream modules


crates/extrinsics/src/lib.rs
```diff
@@ -122,6 +123,67 @@ pub fn account_id(keypair: &Keypair) -> AccountId32 {
...
+    Default,
+    scale::Encode,
+    scale::Decode,
+    scale::MaxEncodedLen,
+    scale_encode::EncodeAsType,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+#[encode_as_type(crate_path = "subxt::ext::scale_encode")]
+pub struct Weight {
```
smiasojed, line 142: Do we want to use this Weight in interaction with this crate, or is it better to use sp_weights and convert it internally?

crates/extrinsics/src/upload.rs
```diff
@@ -239,3 +244,42 @@ pub struct UploadResult {
...
+    /// # Note
+    ///
+    /// **Never** use this mode for on-chain execution.
+    Relaxed,
+}
+
+/// A raw call to `pallet-contracts`'s `upload`.
+#[derive(Debug, scale_encode::EncodeAsType)]
+#[encode_as_type(crate_path = "subxt::ext::scale_encode")]
+pub struct UploadCode {
```
smiasojed, line 281: I think that internal structs should be private, but here they are made public to keep it consistent with the rest of the code.

crates/extrinsics/src/call.rs
```diff
@@ -290,14 +292,20 @@ impl CallExec {
...
+        let call = subxt::tx::Payload::new(
+            "Contracts",
+            "call",
+            Call {
+                dest: self.contract.clone().into(),
+                value: self.value.denominate_balance(&self.token_metadata)?,
+                gas_limit: gas_limit.into(),
+                storage_deposit_limit,
+                data: self.call_data.clone(),
+            },
```
ascjones, line 308: I'd like to see all extrinsic constructors and types grouped together in a module.

crates/extrinsics/src/instantiate.rs
```diff
@@ -363,13 +368,17 @@ impl InstantiateExec {
...
+            "Contracts",
+            "instantiate_with_code",
+            InstantiateWithCode {
+                value: self.args.value,
+                gas_limit: gas_limit.into(),
+                storage_deposit_limit: self.args.storage_deposit_limit_compact(),
+                code,
+                data: self.args.data.clone(),
+                salt: self.args.salt.clone(),
+            },
```
ascjones, line 381: Using the metadata generated API gave us an error when attempting to interact with a version of `pallet_contracts` with incompatible metadata. This was annoying but useful.

We should consider adding some safety and UX back by at least checking that the constructed call we have is compatible with the target node. It will prevent either submitting an extrinsic which succeeds to decode into a different form (bad), or less worse the user just getting a not useful error because the extrinsic could not be decoded.

ascjones (commented):
Could we put all type definitions and constructors for extrinsics and storage keys etc. together in a single module? 

The raw subxt stuff is better encapsulated, and it will be easier once we need to have multiple definitions for different versions.

crates/extrinsics/src/runtime_api/metadata_v15.scale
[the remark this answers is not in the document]
smiasojed: done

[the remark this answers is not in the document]
ascjones: Also can remove them from https://github.com/paritytech/cargo-contract/blob/2b16a8f2a3e6e09e3276200e271d920b22d756b4/crates/extrinsics/Cargo.toml#L15

crates/build/Cargo.toml
```diff
@@ -35,7 +35,7 @@ serde_json = "1.0.108"
 tempfile = "3.8.1"
 term_size = "0.3.2"
 url = { version = "2.4.1", features = ["serde"] }
-wasm-opt = { version = "=0.116.0", deafult-features = false }
```
ascjones, line 38: I think we need this see https://github.com/paritytech/cargo-contract/pull/1380

crates/extrinsics/src/instantiate.rs
ascjones: Okay great I didn't know about that!

ascjones (approved):

PR #1410, opened 2023-11-23
Update README.md

SkymanOne:
Just updates README to include instructions for schema operations.


smiasojed:
As I see, we are also missing 'encode' and 'verify' commands. Maybe you could add them as well.

SkymanOne:
Done

smiasojed (approved):
LGTM

PR #1412, opened 2023-11-28
Integrate `dylint` lints to build

jubnzv:
## Summary
- [y] y/n | Is it dependent on the specific version of `ink` or `pallet-contracts`?

## Description
This PR follows up https://github.com/paritytech/ink/pull/2001 making `dylint`-based lints mandatory in the build process. This change allows us to introduce ink-specific compilation errors more flexibly through the dylint-based ink_linting module.


crates/build/src/lib.rs
```diff
@@ -451,59 +452,25 @@ fn check_buffer_size_invoke_cargo_clean(
...
-    }
-    Ok(())
 }
 
-/// Run linting steps which include `clippy` (mandatory) + `dylint` (optional).
+/// Run linting that involves two steps: `clippy` and `dylint`. Both are mandatory as
+/// they're part of the compilation process and implement security-critical features.
 fn lint(
-    dylint: bool,
+    extra_lints: bool,
```
jubnzv, line 470: I don't like that all the lints will be executed on each build, because it slows down the build process and may bother the user with extra noise generated by lints. Ideally, we should execute only mandatory lints needed to check compilation errors (like `no_main`) and skip all the others.

I think the most simple way to do this is to split our `ink_linting` in two `dylint` libraries, like `ink_linting` and `ink_linting_extra`. The first one will contain only the mandatory lints, and the second one will run the optional analyses when user requesting it.

There are other ways to suppress lints (like hacking `Cargo.toml` in contract projects or compilation flags for `rustc`), but AFAIK these will execute lints anyway, suppressing their output.

WDYT @ascjones?

crates/build/src/lib.rs
```diff
@@ -451,59 +452,25 @@ fn check_buffer_size_invoke_cargo_clean(
...
-fn invoke_cargo_and_scan_for_error(cargo: duct::Expression) -> Result<()> {
-    macro_rules! eprintln_red {
-        ($value:expr) => {{
-            use colored::Colorize as _;
-            ::std::eprintln!("{}", $value.bright_red().bold());
-        }};
+fn execute_cargo(cargo: duct::Expression) -> Result<()> {
+    match util::cargo_tty_output(cargo)
+        .unchecked()
+        .stderr_capture()
```
ascjones, line 458: I think we no longer want to capture `stderr`, we should just let it print direct to the console.

crates/build/src/lib.rs
```diff
@@ -451,59 +452,25 @@ fn check_buffer_size_invoke_cargo_clean(
...
 /// Executes the supplied cargo command, reading the output and scanning for known errors.
 /// Writes the captured stderr back to stderr and maintains the cargo tty progress bar.
-fn invoke_cargo_and_scan_for_error(cargo: duct::Expression) -> Result<()> {
-    macro_rules! eprintln_red {
-        ($value:expr) => {{
-            use colored::Colorize as _;
-            ::std::eprintln!("{}", $value.bright_red().bold());
-        }};
+fn execute_cargo(cargo: duct::Expression) -> Result<()> {
+    match util::cargo_tty_output(cargo)
```
ascjones, line 456: We can remove this if we are no longer redirecting `stderr`. We only needed to do so in order to capture the compiler error.

crates/build/src/lib.rs
ascjones: Yes we only want the mandatory lints to execute every time, extra lints must still be opt-in.

> I think the most simple way to do this is to split our ink_linting in two dylint libraries, like ink_linting and ink_linting_extra

Sounds good to me :+1:

crates/build/src/lib.rs
```diff
@@ -451,59 +452,25 @@ fn check_buffer_size_invoke_cargo_clean(
...
-            return Err(anyhow::anyhow!("missing `no_main` attribute"))
-        }
-        if bytes_read == 0 {
-            // flush the remaining buffered bytes
-            io::Write::write(&mut io::stderr(), err_buf.make_contiguous())?;
-            break
-        }
-        buffer = [0u8; 1];
-    }
-    Ok(())
```
ascjones, line 497: So happy we are removing this :relaxed:

jubnzv:
It is ready for review; however, we need https://github.com/paritytech/ink/pull/2032 to be merged first. This will allow us to install the `ink_linting_mandatory` library in GitHub Actions to fix the CI.

crates/build/src/lib.rs
ascjones: Merged https://github.com/paritytech/ink/pull/2032 now

jubnzv:
https://github.com/paritytech/ink/pull/2060 will enable us to fix the CI issues here.

ascjones (approved):
LGTM!

cmichi (approved):
Thank you, Georgiy!

jubnzv:
It passed the CI and could be merged. But I would like to draw attention to a few points:

1. This PR introduces `rust-toolchain.toml` to `cargo-contract`. The linter runs on a [specific version of the toolchain](https://github.com/paritytech/ink/blob/master/linting/rust-toolchain.toml). This requires users to have that particular toolchain version installed, along with the `clippy` and `dylint` components. To simplify the installation process for the end-user, I suggest using the same toolchain in `cargo-contract`, therefore enforcing it in the `rust-toolchain` file.
2. `cargo-dylint` and `dylint-link` are now mandatory dependencies. I have updated all the instructions and the Dockerfile for the build image accordingly. However, I want to emphasize this point in case there is anything else where we should mention these dependencies.

ascjones:
While I agree with idea of making it simple, I'm not sure about `1.`, because `cargo-contract` is no longer only used for compiling contracts. There are lots of features e.g. for interacting with contracts on the chain, so for another set of users who may not be compiling contracts at all we are making installation more difficult because they have to install that specific toolchain.

For `2.`, I guess you need to update [use.ink setup instructions:](https://github.com/paritytech/ink-docs/blob/master/docs/getting-started/setup.md) source: https://github.com/paritytech/ink-docs/blob/master/docs/getting-started/setup.md

I am beginning to wonder whether we should start to encourage using a docker workflow, especially for workshops because getting setup with `cargo-contract` is becoming more complicated and time consuming, so something we should think about. Just thinking out loud, not something to solve here: but we should definitely keep in mind trying to keep things reasonably accessible when starting from scratch.

cmichi:
I think we should avoid requiring a particular Rust toolchain for `cargo-contract`. As Andrew stated it will result in more complex user instructions. In the past we saw that even adding one step to the installation instructions results in significantly more users asking for help in channels.

It's only needed for `cargo build --lint`, which is optional. My suggestion is to have `cargo build --lint` check the local toolchain version and error out with a meaningful message, asking the user to install the compatible Rust toolchain.

jubnzv:
I have removed the fixed toolchain version. Instead, I provide users with warnings and instructions on how to install the required toolchain. This occurs only if the user compiles for the RiscV target or explicitly runs the build with the `--lint` option.

I suggest not adding any complicated logic to toolchain management in the next release. This approach will give us time to discuss it and ensure that nothing is broken.

ascjones (approved):

cmichi (approved):

PR #1454, opened 2024-01-09
Add storage tests

### LINKED ISSUE #101: Storage inspection command

smiasojed:
## Summary
Closes #101

Add storage tests

### DIFF
--- a/crates/extrinsics/src/contract_storage_tests.rs
+++ b/crates/extrinsics/src/contract_storage_tests.rs
@@ -64,7 +64,7 @@ fn contract_default_spec() -> ContractSpec {
         .constructors(vec![ConstructorSpec::from_label("new")
             .selector([94u8, 189u8, 136u8, 214u8])
             .payable(true)
-            .args(vec![])
+            .args(Vec::new())
             .returns(ReturnTypeSpec::new(TypeSpec::with_name_str::<
                 ConstructorResult<()>,
             >(
@@ -76,7 +76,7 @@ fn contract_default_spec() -> ContractSpec {
             .selector([231u8, 208u8, 89u8, 15u8])
             .mutates(true)
             .payable(true)
-            .args(vec![])
+            .args(Vec::new())
             .returns(ReturnTypeSpec::new(TypeSpec::with_name_str::<
                 MessageResult<()>,
             >(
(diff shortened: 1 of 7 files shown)


crates/extrinsics/src/contract_storage_tests.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[test]
+fn storage_decode_simple_type_works() {
+    let root_key_encoded = Encode::encode(&ROOT_KEY);
+    #[derive(scale_info::TypeInfo, StorageLayout, Storable)]
+    struct Data {
+        a: i32,
+    }
+
+    if let Struct(data_layout) = <Data as StorageLayout>::layout(&ROOT_KEY) {
```
ascjones, line 111: Could use [`let-else` ](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) to have the panic at the top and remove the indentation of the test body e.g.

```rust
let Struct(data_layout) = <Data as StorageLayout>::layout(&ROOT_KEY) else {
  panic!("Layout shall be created");
}

// use `data_layout` here
```

ascjones (approved):
LGTM

PR #1458, opened 2024-01-12
Add raw RPC calls

### LINKED ISSUE #987: Allow raw RPC calls via `cargo contract rpc`
```
cargo contract rpc [options] METHOD [PARAMS]
```

Users cannot make raw RPC calls to the node with the currently available commands.

Adding this capability would improve the debugging process by giving developers direct access to the node's RPC interface, making it easier to diagnose, fix, and experiment with different scenarios.

smiasojed:
## Summary
Closes #987

Add raw RPC calls

## Description
Invoke an RPC call to the node in the format:
`cargo contract rpc [Options] METHOD [PARAMS]`

e.g.

```bash
cargo contract rpc author_insertKey '"sr25"' '"//ALICE"' \
      5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
```

### DIFF
--- a/crates/cargo-contract/src/cmd/rpc.rs
+++ b/crates/cargo-contract/src/cmd/rpc.rs
@@ -1,0 +1,77 @@
+// Copyright 2018-2024 Parity Technologies (UK) Ltd.
+// This file is part of cargo-contract.
+//
+// cargo-contract is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// cargo-contract is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
+
+use contract_extrinsics::{
+    ErrorVariant,
+    Rpc,
+    RpcRequest,
+};
+use contract_transcode::scon::parse_value;
+use serde_json::value::RawValue;
+
+#[derive(Debug, clap::Args)]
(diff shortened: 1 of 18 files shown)


crates/cargo-contract/src/cmd/rpc.rs
```diff
@@ -0,0 +1,74 @@
+// Copyright 2018-2024 Parity Technologies (UK) Ltd.
```
SkymanOne, line 1:
```suggestion
// Copyright (C) Parity Technologies (UK) Ltd.
```

crates/extrinsics/src/rpc.rs
```diff
@@ -0,0 +1,208 @@
+// Copyright 2018-2024 Parity Technologies (UK) Ltd.
```
SkymanOne, line 1:
```suggestion
// Copyright (C) Parity Technologies (UK) Ltd.
```

docs/rpc.md
```diff
@@ -0,0 +1,38 @@
...
+ cargo contract rpc author_hasKey
+     5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY '"sr25"' \
+```
+ 
+account can be provided as ss58 address:
+`5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY`
+or in hex:
+`0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d`
+
+Some of the commands require SS58 address in the string format:
```
SkymanOne, line 21: Is there a way to automatically convert it to string to make it consistent? I see this as a source of user errors in future.

SkymanOne (approved):
LGTM, just few nits

docs/rpc.md
smiasojed: > Is there a way to automatically convert it to string to make it consistent? I see this as a source of user errors in future.

You would need to know for which command it is required. In this case, it means hardcoding the command name and matching agains it

SkymanOne: Sure, then I would create an issue for this, so we can pick it up later

PR #1463, opened 2024-01-19
Add generics to extrinsics crate

smiasojed:
## Summary
Related #1168

Add generics to extrinisics crate

## Description
The PR removes PolkadotConfig and DefaultEnvironment from the extrinsics crate and makes use of generics for Hash, Account and Balance.

- [ ] I have commented my code, particularly in hard-to-understand areas


crates/cargo-contract/src/cmd/mod.rs
```diff
@@ -266,3 +270,33 @@ pub fn display_all_contracts(contracts: &[<DefaultConfig as Config>::AccountId])
...
+        // with 0x prefix
+        assert!(parse_code_hash(
+            "0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"
+        )
+        .is_ok());
+        // without 0x prefix
+        assert!(parse_code_hash(
+            "d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"
+        )
+        .is_ok())
```
cmichi, line 300: Please also add a test for the failure case `.is_err()`.

crates/extrinsics/src/remove.rs
```diff
@@ -110,16 +125,16 @@ impl RemoveCommandBuilder<state::ExtrinsicOptions> {
...
         let transcoder = artifacts.contract_transcoder()?;
         let signer = self.opts.extrinsic_opts.signer()?;
 
         let artifacts_path = artifacts.artifact_path().to_path_buf();
 
         let final_code_hash = match (self.opts.code_hash.as_ref(), artifacts.code.as_ref()) {
-            (Some(code_h), _) => Ok(code_h.0),
-            (None, Some(_)) => artifacts.code_hash(),
+            (Some(code_h), _) => Ok(*code_h),
+            (None, Some(_)) => artifacts.code_hash().map(|h| h.into() ),
```
cmichi, line 137:
```suggestion
            (None, Some(_)) => artifacts.code_hash().map(Into::into),
```

cmichi (approved):
Thanks, Sebastian. Looks good!

ascjones (commented):
I'd like to see a bit more in the PR description to explain the change and how it is an intermediate step towards #1463.

crates/cargo-contract/src/cmd/call.rs
```diff
@@ -74,7 +79,7 @@ pub struct CallCommand {
     proof_size: Option<u64>,
     /// The value to be transferred as part of the call.
     #[clap(name = "value", long, default_value = "0")]
-    value: BalanceVariant,
+    value: BalanceVariant<<DefaultEnvironment as Environment>::Balance>,
```
ascjones, line 82: We may need to use a general concrete type here and for other command fields which use types like this, because clap is responsible for instantiating the commands, and we want to be able (eventually) to choose which `Environment` impl to use for invoking the actual commands. So there will need to be a runtime conversion from the parsed input type to the chosen env type.

smiasojed: Yes, This PR is just a first with focus on extrinisics crate.

crates/extrinsics/src/integration_tests.rs
```diff
@@ -485,13 +490,14 @@ async fn api_build_upload_instantiate_call() {
...
-    let instantiate_result = instantiate_result.unwrap();
+    let instantiate_result: InstantiateExecResult<DefaultConfig> =
+        instantiate_result.unwrap();
     let contract_account = instantiate_result.contract_address.to_string();
     assert_eq!(48, contract_account.len(), "{contract_account:?}");
 
     // call the contract
     // the value should be true
-    let call = CallCommandBuilder::default()
+    let call = CallCommandBuilder::<DefaultConfig, DefaultEnvironment, _, _>::default()
```
ascjones, line 500: It would be nice not to have to do this turbofish, perhaps the generic params could default to the default types here and then you can use a builder method `e.g. `.with_env::<MyConfig, MyEnvironment>`

ascjones (approved):

PR #1465, opened 2024-01-22
Add generics for info and storage commands

smiasojed:
## Summary
Related #1168

Add generics for the 'info' and 'storage' commands


crates/extrinsics/src/contract_storage.rs
```diff
@@ -75,21 +77,26 @@ use super::{
...
     ContractInfo,
-    DefaultConfig,
     TrieId,
 };
 
-pub struct ContractStorage<C: Config = DefaultConfig> {
-    rpc: ContractStorageRpc<C>,
+pub struct ContractStorage<C: Config, E: Environment> {
+    rpc: ContractStorageRpc<C, E>,
+    _phantom: PhantomData<fn() -> (C, E)>,
```
ascjones, line 85: Isn't this redundant because the type params are used in `ContractStorageRpc`?

ascjones (approved):
LGTM

crates/extrinsics/src/contract_storage.rs
smiasojed: I will remove it. Thanks

SkymanOne (approved):
LGTM. Please add changelog entry before you merge

smiasojed:
> LGTM. Please add changelog entry before you merge

I am not sure if I should add it into changelog, it does not affect an user, it is just internal change. WDYT?

SkymanOne:
> > LGTM. Please add changelog entry before you merge
> 
> I am not sure if I should add it into changelog, it does not affect an user, it is just internal change. WDYT?

Generally speaking, unless the change is <10 loc, it's worth adding an entry just for internal reporting reasons so we know what and when it was changed.

cmichi (approved):
I'm fine with not having a changelog entry, as the change doesn't change behavior of `cargo-contract` for users.

PR #1470, opened 2024-01-24
Replace legacy `pallet_contracts_primitives` with local types

ascjones:
Initially this PR began with updating the version of `subxt` to `0.34`, and the updating of the related substrate dependencies. 

However, the `pallet_contracts_primitives` crate is no longer published, so this PR creates local copies of the required types used mainly for the results of dry-run extrinsic RPC calls.

### DIFF
--- a/crates/extrinsics/Cargo.toml
+++ b/crates/extrinsics/Cargo.toml
@@ -32,10 +32,10 @@ serde_json = "1.0.111"
 url = { version = "2.5.0", features = ["serde"] }
 rust_decimal = "1.33"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
-sp-core = "22.0.0"
-sp-runtime = "25.0.0"
-sp-weights = "21.0.0"
-pallet-contracts-primitives = "25.0.0"
+sp-core = "28.0.0"
+sp-runtime = "31.0.0"
+sp-weights = "27.0.0"
+pallet-contracts-primitives = "29.0.0"
 scale-info = "2.10.0"
 subxt = "0.34.0"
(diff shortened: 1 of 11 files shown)


crates/extrinsics/Cargo.toml
```diff
@@ -32,13 +32,13 @@ serde_json = "1.0.111"
...
 rust_decimal = "1.33"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
-sp-core = "22.0.0"
-sp-runtime = "25.0.0"
-sp-weights = "21.0.0"
-pallet-contracts-primitives = "25.0.0"
+sp-core = "28.0.0"
+sp-runtime = "31.0.0"
+sp-weights = "27.0.0"
+pallet-contracts-uapi = "5.0.0"
```
smiasojed, line 38: Do we need these dependency? I am not sure if we are currently using any types that require it

crates/extrinsics/src/pallet_contracts_primitives.rs
```diff
@@ -0,0 +1,186 @@
...
+//
+// cargo-contract is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with cargo-contract.  If not, see <http://www.gnu.org/licenses/>.
+
+use pallet_contracts_uapi::ReturnFlags;
```
smiasojed, line 17: It seems that we only need `pallet-contracts-uapi` for one flag. Maybe we should just copy it?

smiasojed (approved):

PR #1483, opened 2024-01-31
Print type comparison warning only on `--verbose`

### LINKED ISSUE #1482: Print type comparison warning only on `--verbose`
This warning should only be printed when `-v` is used (`verbosity.is_verbose()`):
https://github.com/paritytech/cargo-contract/blob/master/crates/extrinsics/src/env_check.rs#L22-L26

The context is that currently most production chains don't support it yet and we should avoid users getting bombarded with this warning from the next `cargo-contract` release on.

Please change the text to

```
Warning: This chain does not yet support checking for compatibility of your contract types (https://use.ink/faq#type-comparison).
```

All on one line, so that it works well if someone scripts things. 

I think we might have used the color yellow for warnings in `cargo-contract` in the past, but I'm not sure. Best to check.

I'll add the entry for the FAQ with more details.

SkymanOne:
## Summary
Closes #1482

### DIFF
--- a/crates/build/src/args.rs
+++ b/crates/build/src/args.rs
@@ -45,7 +45,9 @@ impl TryFrom<&VerbosityFlags> for Verbosity {
 }
 
 /// Denotes if output should be printed to stdout.
-#[derive(Clone, Copy, Default, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
+#[derive(
+    Clone, Copy, Default, serde::Serialize, serde::Deserialize, Eq, PartialEq, Debug,
+)]
 pub enum Verbosity {
     /// Use default output
     #[default]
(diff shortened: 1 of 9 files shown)


ascjones (approved):

crates/build/src/args.rs
```diff
@@ -45,7 +45,9 @@ impl TryFrom<&VerbosityFlags> for Verbosity {
 }
 
 /// Denotes if output should be printed to stdout.
-#[derive(Clone, Copy, Default, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
+#[derive(
+    Clone, Copy, Default, serde::Serialize, serde::Deserialize, Eq, PartialEq, Debug,
+)]
```
cmichi, line 50:
```suggestion
#[derive(Clone, Copy, Default, serde::Serialize, serde::Deserialize, Eq, PartialEq)]
```

Probably a leftover from you developing.

PR #1496, opened 2024-02-07
Add contract build with lint option to CI

smiasojed:
## Summary
Closes #_

Add contract build with lint option to CI

## Description
Add contract build with lint option to CI, where nightly toolchain is used


crates/build/src/lib.rs
```diff
@@ -116,7 +116,7 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
 pub(crate) mod linting {
     /// Toolchain used to build ink_linting:
     /// https://github.com/paritytech/ink/blob/master/linting/rust-toolchain.toml
-    pub const TOOLCHAIN_VERSION: &str = "nightly-2023-12-28";
+    pub const TOOLCHAIN_VERSION: &str = "nightly-2024-02-07";
```
jubnzv, line 119: `TOOLCHAIN_VERSION` is used to check that the user has the required toolchain version before compiling and guide them to install the that version. Otherwise, the user will see somewhat confusing error messages from `cargo dylint`.

The version should be exactly the same [as specified in the linter](https://github.com/paritytech/ink/blob/master/linting/rust-toolchain.toml), because it depends on the unstable rustc API.

So, I understand the reason why these changes in the CI are needed, and it makes sense for me.
But can we left the value of `TOOLCHAIN_VERSION` as it is?

smiasojed: Done

rcny (approved):

.github/workflows/ci.yml
```diff
@@ -195,9 +195,13 @@ jobs:
...
       - name: Check Template
         run: >-
+          # The linter requires two crates
+          cargo install cargo-dylint dylint-link
           cargo -vV &&
           cargo run --profile debug-ci -- contract --version &&
           cargo run --profile debug-ci -- contract new --target-dir ${{ runner.temp }} foobar &&
           cargo run --profile debug-ci -- contract build --manifest-path=${{ runner.temp }}/foobar/Cargo.toml &&
+          # build with linting, nightly toolchain is used
+          cargo run --profile debug-ci -- contract build --lint --manifest-path=${{ runner.temp }}/foobar/Cargo.toml &&
```
jubnzv, line 205: It is enough to simply run `build --lint` here. The previous line (L202) could be safely removed.

smiasojed: Updated

PR #1499, opened 2024-02-09
:t-rex: fix clippy iterator suggestion

ascjones:


ascjones:
I guess it was already fixed on `master`

PR #1503, opened 2024-02-15
Add multi-config support

smiasojed:
## Summary
Related #1168
Add multi chain support

## Description
Add multi-chain config support. This solution includes an option --config for commands that depend on chain configuration. `Polkadot` serves as the default config, and if not provided, the default configuration is used.

For now, the chain configuration is selected using the `--config` option, but in the future, it may be discovered automatically.

Current limitations:
hash: `[u8;32]` - hardcoded in metadata crate
AccountId32 - transcoder trait implemented only for this type (crate transcoder)
Balance: `From<u128>/Into<u128>`

- [y] My code follows the style guidelines of this project
- [y] I have commented my code, particularly in hard-to-understand areas
- [y] Any dependent changes have been merged and published in downstream modules


crates/cargo-contract/src/main.rs
```diff
@@ -17,6 +17,7 @@
 #![deny(unused_crate_dependencies)]
 
 mod cmd;
+mod config;
```
ascjones, line 20: This could be moved inside the `cmd` module since it appears only to be used there

ascjones (commented):
Looking good, the main thing would be to attempt to reduce the duplication of the verbose trait bounds on all the `run` fns

crates/cargo-contract/src/main.rs
smiasojed: Done

crates/cargo-contract/src/cmd/instantiate.rs
```diff
@@ -28,7 +28,7 @@ use super::{
 use crate::{
     anyhow,
     call_with_config,
-    config::SignerConfig,
+    cmd::config::SignerConfig,
```
ascjones, line 31: Since we are already in `cmd` this could go in `super::config`

crates/cargo-contract/src/cmd/config.rs
```diff
@@ -0,0 +1,213 @@
...
+where
+    <Self as Config>::Signature: From<sp_core::ecdsa::Signature>,
+{
+    type Signer = SignerEcdsa<Self>;
+}
+
+/// A runtime configuration for the Polkadot based chain.
+/// This thing is not meant to be instantiated; it is just a collection of types.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Polkadot {}
```
cmichi, line 78:
```suggestion
pub enum Substrate {}
```

I'm leaning towards renaming it to Substrate, as that's more correct and includes also solochains like Aleph Zero.

cmichi:
I'm getting an unexpected error when trying to instantiate `flipper` on `substrate-contracts-node` with `Ecdsachain` ‒ so a mismatching config:

```
$ cargo contract instantiate --suri //Alice --args true -x -y --verbose --config Ecdsachain
 Dry-running new (skip with --skip-dry-run)
                Result ModuleError: Contracts::StorageDepositNotEnoughFunds: ["Origin doesn't have enough balance to pay the required storage deposits."]
          Gas Consumed Weight { ref_time: 0, proof_size: 0 }
          Gas Required Weight { ref_time: 0, proof_size: 0 }
 Storage Total Deposit StorageDeposit::Charge(0)
ERROR: Pre-submission dry-run failed. Use --skip-dry-run to skip this step.
```

The origin has enough balance. Do you have an idea why this misleading error is output?

Ideally the @SkymanOne implementation of https://github.com/paritytech/cargo-contract/pull/1377 would catch this, but I guess it doesn't check the `Signer` yet. @SkymanOne Can you check?

crates/cargo-contract/src/cmd/config.rs
ascjones: In `subxt` there is another config called [`Substrate`](https://github.com/paritytech/subxt/blob/fd0fe5700281ecd9a11d551f0aa6d4ee21ee6460/subxt/src/config/substrate.rs#L19), which matches the default kitchen sink node extrinsic params. So in `subxt` `Polkadot` -> relay chain and `node-template` config (used by most parachains and `substrate-contracts-node`), and `Substrate` -> substrate kitchen sink node with different extrinsic params.

We can give these different names possibly here though.

smiasojed:
@cmichi If you are using an ECDSA signer to generate keys and sign transactions, but the chain expects SR25519 keys, then the keys will differ. As a result, the addresses derived from these keys will also differ.
The @SkymanOne check will not catch it, because he is not checking node configs like used Signatures

crates/cargo-contract/src/cmd/call.rs
```diff
@@ -95,31 +110,51 @@ impl CallCommand {
...
-        let signer = create_signer(&self.extrinsic_cli_opts.suri)?;
+    async fn run<C: Config + Environment + SignerConfig<C>>(
+        &self,
+    ) -> Result<(), ErrorVariant>
+    where
+        <C as SignerConfig<C>>::Signer: subxt::tx::Signer<C> + Clone + FromStr,
+        <C as Config>::AccountId: IntoVisitor + FromStr + EncodeAsType,
+        <<C as Config>::AccountId as FromStr>::Err: Display,
+        C::Balance: From<u128> + Display + Default + FromStr + Serialize + Debug,
+        <C::ExtrinsicParams as ExtrinsicParams<C>>::OtherParams: Default,
```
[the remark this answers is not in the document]
smiasojed: I was trying to implement it in similar way to this https://github.com/rust-lang/rust/issues/54149,
but it does not reduce the code duplication. Do you know any pattern which could be used here?

crates/cargo-contract/src/cmd/config.rs
smiasojed: Added Substrate config

crates/cargo-contract/src/cmd/call.rs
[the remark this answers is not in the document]
ascjones: Fair enough that might be slightly tricky, fine to have the explicit bounds for now and we'll see if we can refactor in the future

smiasojed:
The `subxt` team is going to check if they can validate `subxt::Config` to see if it matches with node, before issuing request to it. I would suggest to wait for it before we decide to extend the pallet_contract API

ascjones (approved):

cmichi (approved):

PR #1511, opened 2024-02-27
Fix build --verifiable command

smiasojed:
## Summary
Closes #_
- [ ] y/n | Does it introduce breaking changes?
- [ ] y/n | Is it dependent on the specific version of `ink` or `pallet-contracts`?
Fix build --verifiable

## Description
Cargo contract version `4.0.0-rc.2`

`cargo contract build --verifiable`
causes an issue:
 ```
 [==] Checking clippy linting rules
error: 'cargo-clippy' is not installed for the toolchain 'stable-x86_64-unknown-linux-gnu'
To install, run `rustup component add clippy`
ERROR: 
ERROR: Failed to read build result from docker build
```
## Checklist before requesting a review
- [ ] My code follows the style guidelines of this project
- [ ] I have added an entry to `CHANGELOG.md`
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1513, opened 2024-02-28
Release `4.0.0-rc.3`

SkymanOne:
### Fixed
- Fix build `--verifiable` command [#1511](https://github.com/paritytech/cargo-contract/pull/1511)

### Note
Also bumps `ink` crates to `5.0.0-rc.2`


crates/metadata/compatibility_list.json
```diff
@@ -8,6 +8,9 @@
         },
         "4.0.0-rc.2": {
             "ink": ["5.0.0-alpha", "5.0.0"]
+        },
+        "4.0.0-rc.3" : {
+            "ink": ["5.0.0-rc.2", "5.0.0"]
```
smiasojed, line 13: we have overlapping range with previous entry. As I remember in 5.0.0-rc
 we have changed storage keys, so probably we should have such configuration:
        "4.0.0-rc.1": {
            "ink": [">=5.0.0-alpha, <5.0.0-rc.1"]
         "4.0.0-rc.3" : {
            "ink": ["5.0.0-rc.1", "5.0.0"]
           }

smiasojed: Now we do not have any c-c version compatible with `5.0.0-rc.1`

cmichi (approved):

PR #1516, opened 2024-03-02
Make the error message on incompatible versions clearer

DamianStraszak:
## Summary
- [] n | Does it introduce breaking changes?
- [] n | Is it dependent on the specific version of `ink` or `pallet-contracts`?

## Description
Previously the error message when using ver `4` of cargo contracts on contracts written in ink version `4.x` was to "update to cargo contract to 3.2" which is slightly confusing, since you actually have to downgrade then. The error message was adjusted.


DamianStraszak:
Yes, fair point. Adjusted now accordingly.

PR #1526, opened 2024-03-07
Fix parsing of docker STDOUT

### LINKED ISSUE #1455: Unable to build a verifiable ink! flipper contract
**Describe the bug**

When trying to execute a verifiable build locally for the [flipper contract](https://github.com/paritytech/ink-examples/tree/main/flipper) from [ink-examples](https://github.com/paritytech/ink-examples) project, I face the following error:

```bash
cargo contract build --verifiable
 [==] Started the build inside the container: ink-verified-flipper-63196
 [==] Checking clippy linting rules
error: 'cargo-clippy' is not installed for the toolchain 'stable-x86_64-unknown-linux-gnu'
To install, run `rustup component add clippy`
 [==] Building cargo project
    Building...
    Building [=========================> ] 86/87: flipper(bin)
    Finished release [optimized] target(s) in 2m 33s
 [==] Post processing code
 [==] Generating metadata
   Compiling...
    Finished release [optimized] target(s) in 5m 52s
    Running `target/ink/release/metadata-gen`
 [==] Generating bundle
ERROR: Error decoding BuildResult:
 }

Caused by:
    expected value at line 1 column 1
```

**To Reproduce**
1. Build the [flipper contract](https://github.com/paritytech/ink-examples/tree/main/flipper) with the --verifiable flag:

`cargo contract build --verifiable`

**Expected behavior**
The expected response should contain metadata with the `paritytech/contracts-verifiable` image according to the [documentation](https://use.ink/basics/verification/contract-verification).

**Environment (please complete the following information):**
 - OS: MacOS M1 darwin
 - rust: 1.72.1
 - rust toolchain: 1.72-aarch64-apple-darwin
 - `cargo-contract` version: 4.0.0-rc.1
 - ink! version: 4.2.0

**Additional context**

I don't think the clippy error cause this issue since I also tried overriding the `paritytech/contracts-verifiable` image with clippy installed.

SkymanOne:
## Summary
Closes #1455

It was previously common to see the following error at the end of a verifiable build, even when the build was successful:
image

The previous implementation expected the `BuildResult` to be sent in a stream as a whole chunk of bytes. However, there can be cases when the data is sent in multiple chunks, which results in incorrect parsing.

We simply collect the byte chunks and then attempt to decode them together.

### DIFF
--- a/crates/build/src/docker.rs
+++ b/crates/build/src/docker.rs
@@ -414,11 +414,11 @@ async fn run_build(
     let stderr = std::io::stderr();
     let mut stderr = stderr.lock();
     let mut build_result = None;
-    let mut build_bytes: Vec<u8> = vec![];
+    let mut message_bytes: Vec<u8> = vec![];
     while let Some(Ok(output)) = output.next().await {
         match output {
             LogOutput::StdOut { message } => {
-                build_bytes.extend(&message);
+                message_bytes.extend(&message);
             }
             LogOutput::StdErr { message } => {
                 stderr.write_all(message.as_ref())?;
@@ -431,12 +431,12 @@ async fn run_build(
         };
     }
 
-    if !build_bytes.is_empty() {
+    if !message_bytes.is_empty() {
         build_result = Some(
-            serde_json::from_reader(BufReader::new(build_bytes.as_slice())).context(
+            serde_json::from_reader(BufReader::new(message_bytes.as_slice())).context(
                 format!(
                     "Error decoding BuildResult:\n {}",
-                    std::str::from_utf8(&build_bytes).unwrap()
+                    std::str::from_utf8(&message_bytes).unwrap()
                 ),
             ),
         )


crates/build/src/docker.rs
```diff
@@ -414,17 +414,11 @@ async fn run_build(
     let stderr = std::io::stderr();
     let mut stderr = stderr.lock();
     let mut build_result = None;
+    let mut build_bytes: Vec<u8> = vec![];
```
smiasojed, line 417: I think `stdout_message` would be a better fit here.

ascjones: Maybe a better name would be something like `stdout_bytes/buffer`

ascjones (approved):

smiasojed:
I think it would be great to have a CI test, which could catch issues like this.

SkymanOne:
> I think it would be great to have a CI test, which could catch issues like this.

The problem is that we need to have a nested docker environment. Idk if it's possible to run docker container inside another container.

smiasojed:
> > I think it would be great to have a CI test, which could catch issues like this.
> 
> The problem is that we need to have a nested docker environment. Idk if it's possible to run docker container inside another container.

I do not see any docker image used for CI, it seems that we are building this directly on the runner

SkymanOne:
> > > I think it would be great to have a CI test, which could catch issues like this.
> > 
> > 
> > The problem is that we need to have a nested docker environment. Idk if it's possible to run docker container inside another container.
> 
> I do not see any docker image used for CI, it seems that we are building this directly on the runner

Sure, I'll do it in the follow up PR then

smiasojed (approved):

cmichi:
> Sure, I'll do it in the follow up PR then

@SkymanOne Please create a GitHub issue for it and assign it to yourself.

cmichi (approved):

PR #1551, opened 2024-03-20
Verify raw Wasm in `cargo contract verify`

### LINKED ISSUE #1539: Add `--wasm` to `cargo contract verify`
The `cargo contract verify` requires passing a full `*.contract` bundle. This is unhandy if one has the source code of a contract and wants to verify if a deployed contract matches that bundle.

We should add a argument `--wasm` that allows passing a binary obtained via e.g. `cargo contract info --contract … --binary > contract.wasm`.

SkymanOne:
## Summary
Closes #1539

Adds a raw verification of the contract against the reference wasm binary.

The command executes the code in the `Release` mode by default and then compares the wasm output with the reference contract.

### DIFF
--- a/crates/cargo-contract/src/cmd/verify.rs
+++ b/crates/cargo-contract/src/cmd/verify.rs
@@ -89,13 +89,7 @@ impl VerifyCommand {
         path: &PathBuf,
     ) -> Result<VerificationResult> {
         // 1. Read code hash binary from the path.
-        let file = File::open(path)
-            .context(format!("Failed to open contract binary {}", path.display()))?;
-
-        let mut ref_reader = BufReader::new(file);
-        let mut ref_buffer = Vec::new();
-        ref_reader
-            .read_to_end(&mut ref_buffer)
+       let ref_buffer = std::fs::read(path)
             .context(format!("Failed to read contract binary {}", path.display()))?;
 
         let reference_code_hash = code_hash(&ref_buffer);
@@ -126,16 +120,8 @@ impl VerifyCommand {
                 .bright_yellow())
         };
 
-        let file = File::open(&built_wasm_path).context(format!(
-            "Failed to open contract binary {}",
-            &built_wasm_path.display()
-        ))?;
-
-        let mut target_reader = BufReader::new(file);
-        let mut target_buffer = Vec::new();
-        target_reader
-            .read_to_end(&mut target_buffer)
-            .context(format!("Failed to read contract binary {}", path.display()))?;
+       let target_buffer = std::fs::read(&built_wasm_path)
+            .context(format!("Failed to read contract binary {}", built_wasm_path.display()))?;
 
         let output_code_hash = code_hash(&target_buffer);
 
--- a/crates/cargo-contract/src/cmd/verify.rs
+++ b/crates/cargo-contract/src/cmd/verify.rs
@@ -130,8 +130,8 @@ impl VerifyCommand {
             anyhow::bail!(format!(
                 "\nFailed to verify the authenticity of wasm binary at {} against the workspace \n\
                 found at {}.\n Expected {}, found {}",
-                format!("`{}`", built_wasm_path.display()).bright_white(),
                 format!("`{}`", path.display()).bright_white(),
+                format!("`{}`", built_wasm_path.display()).bright_white(),
                 format!("{}", reference_code_hash).bright_white(),
                 format!("{}", output_code_hash).bright_white())
             );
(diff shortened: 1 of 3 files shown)


crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -62,9 +73,100 @@ impl VerifyCommand {
...
+            &built_wasm_path.display()
+        ))?;
+
+        let mut reader = BufReader::new(file);
+        let mut buffer: [u8; 32] = [0; 32];
+        reader
+            .read_exact(&mut buffer)
+            .context(format!("Failed to read contract binary {}", path.display()))?;
+
+        let output_code_hash = CodeHash(buffer);
```
cmichi, line 141: Why are you only reading the first 32 bytes of the Wasm binary here and then assume it's the hash?

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -62,9 +73,100 @@ impl VerifyCommand {
...
+        let file = File::open(path)
+            .context(format!("Failed to open contract binary {}", path.display()))?;
+
+        let mut reader = BufReader::new(file);
+        let mut buffer: [u8; 32] = [0; 32];
+        reader
+            .read_exact(&mut buffer)
+            .context(format!("Failed to read contract binary {}", path.display()))?;
+
+        let reference_code_hash = CodeHash(buffer);
```
ascjones, line 102: This is just loading the first 32 bytes of the wasm file, it is not the code hash.

This could result in a false positive if the first 32 bytes of each file are the same.

Could just do `std::fs::read(path)?` to load the whole file.

crates/cargo-contract/src/cmd/verify.rs
ascjones, line 141: Again, just want to load the entire file.

ascjones (commented):
As commented, the whole files should be compared not just the first 32 bytes.

Also would be good to add to the existing [integration test](https://github.com/paritytech/cargo-contract/blob/b731f54b66d4b2ebab3deaaf23196bdbd4799d0c/crates/cargo-contract/tests/verify.rs) here.

In the issue it says:

> We should add a argument --wasm that allows passing a binary obtained via e.g. cargo contract info --contract … --binary > contract.wasm.

If that is the common use case then it would be great to fetch that automatically, so you can just specify the account id in the `verify` command. @cmichi

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -62,9 +72,100 @@ impl VerifyCommand {
...
+    ) -> Result<VerificationResult> {
+        // 1. Read code hash binary from the path.
+        let file = File::open(path)
+            .context(format!("Failed to open contract binary {}", path.display()))?;
+
+        let mut ref_reader = BufReader::new(file);
+        let mut ref_buffer = Vec::new();
+        ref_reader
+            .read_to_end(&mut ref_buffer)
+            .context(format!("Failed to read contract binary {}", path.display()))?;
```
ascjones, line 99: Could be shortened to:

```suggestion
       let ref_buffer = std::fs::read(path)
            .context(format!("Failed to read contract binary {}", path.display()))?;
```

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -62,9 +72,100 @@ impl VerifyCommand {
...
+        let file = File::open(&built_wasm_path).context(format!(
+            "Failed to open contract binary {}",
+            &built_wasm_path.display()
+        ))?;
 
+        let mut target_reader = BufReader::new(file);
+        let mut target_buffer = Vec::new();
+        target_reader
+            .read_to_end(&mut target_buffer)
+            .context(format!("Failed to read contract binary {}", path.display()))?;
```
ascjones, line 138:
```suggestion
       let target_buffer = std::fs::read(&built_wasm_path)
            .context(format!("Failed to read contract binary {}", built_wasm_path.display()))?;
```

crates/cargo-contract/src/cmd/verify.rs
```diff
@@ -62,9 +72,100 @@ impl VerifyCommand {
...
+        let output_code_hash = code_hash(&target_buffer);
+
+        if output_code_hash != reference_code_hash {
+            anyhow::bail!(format!(
+                "\nFailed to verify the authenticity of wasm binary at {} against the workspace \n\
+                found at {}.",
+                format!("`{}`", path.display()).bright_white(),
+                format!("{:?}", manifest_path.as_ref()).bright_white()).bright_red()
+            );
+        }
```
ascjones, line 149: If we are generating and comparing the hashes, instead of just the raw files, we should at least display them here.

PR #1584, opened 2024-04-10
Improve texts

cmichi:


crates/build/src/crate_metadata.rs
```diff
@@ -117,6 +117,7 @@ impl CrateMetadata {
             .packages
             .iter()
             .find_map(|package| {
+                eprintln!("name: {:?}", package.name);
```
smiasojed, line 120: Is it expected to be here?

cmichi: Argh. Thx.

cmichi, line 120:
```suggestion
```

smiasojed (approved):

PR #1594, opened 2024-04-12
Remove parity-wasm dependency from contract-build crate

### LINKED ISSUE #1141: Replace `parity-wasm` usage for post processing
`parity-wasm` has been [deprecated](https://github.com/paritytech/parity-wasm/pull/334) for a while, we should migrate to something else. Possibly [wasmparser](https://crates.io/crates/wasmparser) and [wasm-encoder](https://crates.io/crates/wasm-encoder) as suggested [here](https://github.com/filecoin-project/ref-fvm/issues/829).

Rel https://github.com/paritytech/cargo-contract/issues/1139

smiasojed:
## Summary
Closes #1141

Replace `parity-wasm` usage for post processing with `wasmparser` and `wasm-encoder`

## Description
Remove `parity-wasm` from `contract-build` crate
Implement the `WASM` post-processing logic with `wasmparser` and `wasm-encoder` crates


crates/build/src/lib.rs
```diff
@@ -689,93 +700,186 @@ fn check_dylint_requirements(_working_dir: Option<&Path>) -> Result<()> {
...
+
+        // Support for nested components and modules
+        match payload {
+            Payload::Version { encoding, .. } => {
+                output.extend_from_slice(match encoding {
+                    wasmparser::Encoding::Component => &wasm_encoder::Component::HEADER,
+                    wasmparser::Encoding::Module => &wasm_encoder::Module::HEADER,
+                });
+            }
+            Payload::ModuleSection { .. } | Payload::ComponentSection { .. } => {
```
smiasojed, line 835: @ascjones, I think we do not support components, and we expect only one module in WASM file.
Options are to:
- parse each of them and apply post-processing rules (some operations may not have sense here like shrinking exports)
- return error
What do you think?

crates/build/src/lib.rs
```diff
@@ -764,18 +778,79 @@ fn post_process_wasm(
...
     verbosity: &Verbosity,
-    max_memory_pages: u32,
+    max_memory_pages: u64,
 ) -> Result<()> {
     // Deserialize Wasm module from a file.
-    let mut module =
+    let module =
         load_module(optimized_code).context("Loading of optimized wasm failed")?;
+    let mut output = Vec::new();
+    post_process_module(
```
ascjones, line 787: Maybe `post_process_module` could return an allocated `Vec<u8>` since it looks like every usage just passes a new buffer instance.

ascjones (commented):
Have you tested this against some standard contracts to check that the sizes are the same (or better!)? Would be good to run the contract size check against the `integration-tests`.

Also I'd like to see this wasm post-processing refactored to a separate module, this `lib.rs` file is big enough already!

smiasojed:
> Have you tested this against some standard contracts to check that the sizes are the same (or better!)?
I tested it on some ink-examples, and the output was the same for both the old and new implementations.

ascjones (approved):

PR #1619, opened 2024-04-25
Update owners

cmichi:
Changes are in line with the CLA and IP transfer.

### DIFF
--- a/FILE_HEADER
+++ b/FILE_HEADER
@@ -1,4 +1,4 @@
-// Copyright {20\d{2}}-{20\d{2}} Parity Technologies (UK) Ltd.
+// Copyright {20\d{2}}-{20\d{2}} Use Ink (UK) Ltd.
 // This file is part of cargo-contract.
 //
 // cargo-contract is free software: you can redistribute it and/or modify
--- a/crates/analyze/Cargo.toml
+++ b/crates/analyze/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "contract-analyze"
 version = "4.1.1"
-authors = ["Parity Technologies <admin@parity.io>"]
+authors = ["Use Ink <ink@r0gue.io>"]
 edition = "2021"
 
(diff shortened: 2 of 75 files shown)


FILE_HEADER
```diff
@@ -1,4 +1,4 @@
-// Copyright {20\d{2}}-{20\d{2}} Parity Technologies (UK) Ltd.
+// Copyright {20\d{2}}-{20\d{2}} Use Ink (UK) Ltd.
```
smiasojed, line 1: Is `Use Ink` a company registered in the United Kingdom?

cmichi: Yes, cc @evilrobot-01 @peterwht.

evilrobot-01: Correct! 

USE INK LTD
Company number 15664482

smiasojed (approved):

crates/analyze/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "contract-analyze"
 version = "4.1.1"
-authors = ["Parity Technologies <admin@parity.io>"]
+authors = ["Use Ink <ink@r0gue.io>"]
```
evilrobot-01, line 4: Can we set something up to use use.ink instead?

cmichi: That's a good point. Is it possible for you set up `ink@use.ink` once the domains have been transferred?

evilrobot-01: Yeah, we can set it up as the existing one used. Please do let us know if you have any specific address in mind?

Would suggest that all other services currently being set up are then transitioned to this new address, but this can be done behind the scenes after the initial transition.

peterwht (approved):
Approved! To second Frank, it would be good to get an `ink@use.ink` or similar (e.g. `hi@use.ink`) for the authors.

al3mart (approved):

PR #1722, opened 2024-08-07
chore: bump `subxt` and `subxt-signer` versions

AlexD10S:
This PR upgrades the `subxt` and `subxt-signer` library from version 0.35.3 to 0.37.0.
This upgrade is necessary because version 0.37 of `subxt` includes support for the signed extension `CheckMetadataHash`.

This solves an issue when uploading a smart contract on certain chains.
How to Replicate the Issue:
```
// Generate a contracts parachain
pop new parachain contracts-parachain --template contracts
cd contracts-parachain
// Run it
pop up parachain -f ./network.toml
// Generate a new smart contract
cd .. & pop new contract
// Try to upload it
cargo contract upload --suri //Alice --url ws://127.0.0.1:57869 -x
```

This will throw the error:
```
ERROR: Extrinsic params error: The chain expects a signed extension with the name CheckMetadataHash, but we did not provide one
```
This issue is resolved with this upgrade.

**_Note_** 
 In addition to bumping the `subxt` version, some other changes were made:
- Remove the file `metadata_v11.scale `and its test. The test was failing [here](https://github.com/use-ink/cargo-contract/blob/master/crates/extrinsics/src/contract_info.rs#L412). Based on what I read in an issue opened in the subxt repo [link](https://github.com/paritytech/subxt/issues/1648), it seems the only way to fix it is to regenerated the metadata files used in the tests after the upgrade.
However, using the latest `subxt-cli `to regenerate the file for V11 was showing an error:
```
The node can only return version 14 metadata using the legacy API, but you've requested a different version.
```
The reason for this error is that `subxt` doesn't support anything before V14. Since `cargo-contract` relies heavily on `subxt`, the best approach seems to be supporting only from V14 onwards.

- The CI was failing during the clippy checks, so I had to add some code to fix that. To make it easier for the reviewer, all changes related to this issue can be seen in this commit: https://github.com/use-ink/cargo-contract/pull/1722/commits/37cfd9308bff3fb297527b03f62037a08a5daa17


crates/build/src/docker.rs
```diff
@@ -26,12 +26,11 @@
...
-//! 3. Calculate the digest of the command and use it
-//! to uniquely identify the container
+//! 3. Calculate the digest of the command and use it to uniquely identify the container
 //! 4. If the container exists, we just start the build, if not, we create it
-//! 5. After the build, the docker container produces metadata with
-//! paths relative to its internal storage structure, we parse the file
-//! and overwrite those paths relative to the host machine.
+//! 5. After the build, the docker container produces metadata with paths relative to its
+//!    internal storage structure, we parse the file and overwrite those paths relative to
+//!    the host machine.
```
Daanvdplas, line 33: This seams incorrect format

Daanvdplas (approved):
LGTM!

al3mart (commented):
`contract-build` seems to be depending on a yanked version of `zip` crate.
See: https://crates.io/crates/zip/versions

Maybe we could bump it here as well

crates/build/src/docker.rs
AlexD10S: The CI was failing because of this issue. You can run it locally with: 
```
cargo +nightly  clippy --message-format=json --profile debug-ci --all-features --all-targets -- -D warnings
```

al3mart (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Updated the toolchain version used by `ink_linting` - [#1616](https://github.com/paritytech/cargo-contract/pull/1616)
```
peterwht, line 10: Please add a "Removed" section. See [this for reference](https://github.com/use-ink/cargo-contract/blob/f99fc6cbea000ea5601e35888bf2cd9a8fb112f4/CHANGELOG.md#removed-1) for the removal of v11 metadata.

peterwht (commented):
LGTM. I think we just need a small change to the changelog. See below.

peterwht (approved):

smiasojed:
Maybe we should add some information to the README that c-c is compatible starting from version 14. Previously, we supported very old versions—I tested it starting from version 8.

smiasojed (approved):

PR #1851, opened 2024-12-03
Target `pallet-revive` instead of `pallet-contracts`

cmichi:
Based on work by @peterwht and @AlexD10S.

I'll remove the Wasm stuff from `cargo-contract` as a follow-up. I'll also remove the `revive` feature from `ink` as a follow-up.


crates/build/riscv32emac-unknown-none-polkavm.json
athei, line 1: It's no longer necessary to copy paste this file. It is embedded into the linker crate:
https://docs.rs/polkavm-linker/latest/polkavm_linker/fn.target_json_64_path.html

And also we are using 64bit now. 32bit contracts are no longer accepted in the current master.

PR #1910, opened 2025-01-23
Deprecate WebAssembly + Migrate to `pallet-revive`

cmichi:
wip


crates/build/riscv64emac-unknown-none-polkavm.json
athei, line 1: You can get rid of this file and just use:
https://docs.rs/polkavm-linker/latest/polkavm_linker/fn.target_json_64_path.html

cmichi: @athei Thanks for the note! I saw your comment on another PR of mine on this file a while ago :-). Didn't forget, just didn't get around to it yet.

PR #1930, opened 2025-02-03
Generate Solidity compatible metadata 

### LINKED ISSUE #1070: Eth ABI compatibility: emit eth ABI file
Emit an ethereum tooling compatible ABI file for `ink!` smart contracts. ([tracking issue](https://github.com/paritytech/polkadot-sdk/issues/111))

Questions:
- Can we take the existing metadata and just translate it directly into an eth ABI file?

TODO: flesh this out.

Related: https://github.com/paritytech/ink/issues/1748

davidsemakula:
## Summary
Closes #1070 
Also closes https://github.com/use-ink/ink-alliance/issues/13
- [y] y/n | Does it introduce breaking changes?

Adds option to generate Solidity compatible metadata files

## Description

- Adds a CLI arg `cargo contract build --metadata <ink|solidity>` for choosing the specification to use for metadata generation
- Adds ability to generate Solidity compatible [ABI](https://docs.soliditylang.org/en/latest/abi-spec.html#json) and [metadata](https://docs.soliditylang.org/en/latest/metadata.html) files (i.e. `<artifact_name>.abi` and `<artifact_name>.json`)

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -28,6 +28,7 @@ mod crate_metadata;
 mod docker;
 pub mod metadata;
 mod new;
+mod solidity_metadata;
 #[cfg(test)]
 mod tests;
 pub mod util;
@@ -42,6 +43,7 @@ pub use self::{
         BuildArtifacts,
         BuildMode,
         Features,
+        MetadataSpec,
         Network,
         OutputType,
         Target,
@@ -133,6 +135,7 @@ pub struct ExecuteArgs {
     pub output_type: OutputType,
     pub skip_clippy_and_linting: bool,
     pub image: ImageVariant,
+    pub metadata_spec: MetadataSpec,
 }
 
 /// Result of the build process.
@@ -204,8 +207,12 @@ impl BuildResult {
         );
         if let Some(metadata_result) = self.metadata_result.as_ref() {
             let bundle = format!(
-                "  - {} (code + metadata)\n",
-                util::base_name(&metadata_result.dest_bundle).bold()
+                "  - {} ({})\n",
+                util::base_name(&metadata_result.dest_bundle).bold(),
+                match metadata_result.spec {
+                    MetadataSpec::Ink => "code + metadata",
+                    MetadataSpec::Solidity => "Solidity compatible contract metadata",
+                }
             );
             out.push_str(&bundle);
         }
@@ -218,8 +225,12 @@ impl BuildResult {
         }
         if let Some(metadata_result) = self.metadata_result.as_ref() {
             let metadata = format!(
-                "  - {} (the contract's metadata)",
-                util::base_name(&metadata_result.dest_metadata).bold()
+                "  - {} (the contract's {})",
+                util::base_name(&metadata_result.dest_metadata).bold(),
+                match metadata_result.spec {
+                    MetadataSpec::Ink => "metadata",
+                    MetadataSpec::Solidity => "Solidity compatible ABI",
+                }
             );
             out.push_str(&metadata);
         }
@@ -680,6 +691,7 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
         unstable_flags,
         extra_lints,
         output_type,
+        metadata_spec,
         ..
     } = &args;
 
@@ -701,6 +713,8 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
     let clean_metadata = || {
         fs::remove_file(crate_metadata.metadata_path()).ok();
         fs::remove_file(crate_metadata.contract_bundle_path()).ok();
+        fs::remove_file(solidity_metadata::abi_path(&crate_metadata)).ok();
+        fs::remove_file(solidity_metadata::metadata_path(&crate_metadata)).ok();
     };
 
     let (opt_result, metadata_result, dest_binary) = match build_artifact {
@@ -722,17 +736,46 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
                     clean_metadata();
                 })?;
 
+            let (dest_metadata, dest_bundle) = match metadata_spec {
+                MetadataSpec::Ink => {
+                    (
+                        crate_metadata.metadata_path(),
+                        crate_metadata.contract_bundle_path(),
+                    )
+                }
+                MetadataSpec::Solidity => {
+                    (
+                        solidity_metadata::abi_path(&crate_metadata),
+                        solidity_metadata::metadata_path(&crate_metadata),
+                    )
+                }
+            };
             let metadata_result = MetadataArtifacts {
-                dest_metadata: crate_metadata.metadata_path(),
-                dest_bundle: crate_metadata.contract_bundle_path(),
+                spec: *metadata_spec,
+                dest_metadata,
+                dest_bundle,
             };
 
-            // skip metadata generation if contract unchanged and all metadata artifacts
-            // exist.
+            // skip metadata generation if contract is unchanged, metadata spec is
+            // unchanged, and all metadata artifacts exist.
+            let pre_metadata_spec =
+                fs::read_to_string(&crate_metadata.metadata_spec_path);
+            let is_unchanged_metadata_spec =
+                pre_metadata_spec.ok() == Some(metadata_spec.to_string());
             if opt_result.is_some()
+                || !is_unchanged_metadata_spec
                 || !metadata_result.dest_metadata.exists()
                 || !metadata_result.dest_bundle.exists()
             {
+                // Persists the current metadata spec used so we trigger regeneration
+                // when we switch
+                if !is_unchanged_metadata_spec {
+                    fs::write(
+                        &crate_metadata.metadata_spec_path,
+                        metadata_spec.to_string(),
+                    )?;
+                }
+
                 // if metadata build fails after a code build it might become stale
                 clean_metadata();
                 metadata::execute(
@@ -744,6 +787,7 @@ pub fn execute(args: ExecuteArgs) -> Result<BuildResult> {
                     *verbosity,
                     unstable_flags,
                     build_info,
+                    *metadata_spec,
                 )?;
             }
             (opt_result, Some(metadata_result), Some(dest_binary))
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -211,7 +211,7 @@ impl BuildResult {
                 util::base_name(&metadata_result.dest_bundle).bold(),
                 match metadata_result.spec {
                     MetadataSpec::Ink => "code + metadata",
-                    MetadataSpec::Solidity => "Solidity compatible contract metadata",
+                    MetadataSpec::Solidity => "Solidity compatible metadata",
                 }
             );
             out.push_str(&bundle);
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -1068,6 +1068,7 @@ mod unit_tests {
         let raw_result = r#"{
   "dest_binary": "/path/to/contract.polkavm",
   "metadata_result": {
+    "spec": "ink!",
     "dest_metadata": "/path/to/contract.json",
     "dest_bundle": "/path/to/contract.contract"
   },
@@ -1085,6 +1086,7 @@ mod unit_tests {
         let build_result = BuildResult {
             dest_binary: Some(PathBuf::from("/path/to/contract.polkavm")),
             metadata_result: Some(MetadataArtifacts {
+                spec: MetadataSpec::Ink,
                 dest_metadata: PathBuf::from("/path/to/contract.json"),
                 dest_bundle: PathBuf::from("/path/to/contract.contract"),
             }),
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -1075,7 +1075,7 @@ mod unit_tests {
         let raw_result = r#"{
   "dest_binary": "/path/to/contract.polkavm",
   "metadata_result": {
-    "ink": {
+    "Ink": {
       "dest_metadata": "/path/to/contract.json",
       "dest_bundle": "/path/to/contract.contract"
     },
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -1078,7 +1078,7 @@ mod unit_tests {
     "Ink": {
       "dest_metadata": "/path/to/contract.json",
       "dest_bundle": "/path/to/contract.contract"
-    },
+    }
   },
   "target_directory": "/path/to/target",
   "linker_size_result": {
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -36,7 +36,7 @@ mod validate_bytecode;
 mod workspace;
 
 #[deprecated(since = "2.0.2", note = "Use MetadataArtifacts instead")]
-pub use self::metadata::MetadataArtifacts as MetadataResult;
+pub use self::metadata::InkMetadataArtifacts as MetadataResult;
 
 pub use self::{
     args::{
@@ -55,9 +55,11 @@ pub use self::{
     crate_metadata::CrateMetadata,
     metadata::{
         BuildInfo,
+        InkMetadataArtifacts,
         MetadataArtifacts,
     },
     new::new_contract_project,
+    solidity_metadata::SolidityMetadataArtifacts,
     util::DEFAULT_KEY_COL_WIDTH,
     workspace::{
         Lto,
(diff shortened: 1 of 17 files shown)


davidsemakula:
@cmichi should be fixed - hadn't considered these types at all 🙂

`internal/lang-err/call-builder`should still choke on an `Option<T>` return type, but I guess that's expected behavior.

crates/build/src/solidity_metadata/abi.rs
```diff
@@ -0,0 +1,377 @@
...
+            }
+        }
+        // TODO: (@davidsemakula) should `Vec<u8>` be represented as `bytes[]`? (i.e.
+        // dynamic size arrays).
+        // Ref: <https://docs.soliditylang.org/en/latest/types.html#bytes-and-string-as-arrays>
+        TypeDef::Sequence(type_def_seq) => {
+            let elem_ty_id = type_def_seq.type_param.id;
+            let elem_ty = resolve_ty(elem_ty_id, registry, msg)?;
+            Ok(format!("{elem_ty}[]"))
+        }
```
davidsemakula, line 314: Should `Vec<u8>` be `bytes[]`?

crates/build/src/solidity_metadata/abi.rs
```diff
@@ -0,0 +1,377 @@
...
+        // TODO: (@davidsemakula) should `[u8; N]` arrays where `1 <= N <= 32` be
+        // represented as `bytes<N>`? (i.e. fixed-sized arrays).
+        // Ref: <https://docs.soliditylang.org/en/latest/types.html#fixed-size-byte-arrays>
+        TypeDef::Array(type_def_array) => {
+            let elem_ty_id = type_def_array.type_param.id;
+            let elem_ty = resolve_ty(elem_ty_id, registry, msg)?;
+            let len = type_def_array.len;
+            Ok(format!("{elem_ty}[{len}]"))
+        }
```
davidsemakula, line 323: Should `[u8; N]` where `1 <= N <= 32` be `bytes<N>`?

crates/build/src/solidity_metadata/abi.rs
```diff
@@ -0,0 +1,377 @@
...
+        .resolve(id)
+        .unwrap_or_else(|| panic!("Failed to resolve type `#{}` in {}", id, msg));
+    match &ty.type_def {
+        TypeDef::Composite(_) => {
+            let path_segments: Vec<_> =
+                ty.path.segments.iter().map(String::as_str).collect();
+            let ty = match path_segments.as_slice() {
+                // TODO: (@davidsemakula) should `primitive_types::H160` be bytes20?
+                ["ink_primitives", "types", "AccountId"]
+                | ["primitive_types", "H160"] => "address",
```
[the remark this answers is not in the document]
cmichi: You mean if `primitive_types::H160` should be mapped to bytes20 for the Solidity metadata? Is there no native "H160" type in Solidity, for contract addresses?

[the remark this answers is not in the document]
davidsemakula: > You mean if `primitive_types::H160` should be mapped to bytes20 for the Solidity metadata

Yes

> Is there no native "H160" type in Solidity, for contract addresses?

That would be `address` (the current implementation), the question is more about the ink! side. Is this a "safe" mapping? Does `primitive_types::H160` always represent an address or is it just a 20 bytes hash?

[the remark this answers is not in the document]
davidsemakula: On the Solidity side it's trivial to convert between `address` <-> `bytes20` <-> `uint160`. I'm just wondering if its "sound" to assume `primitive_types::H160` always represents an address on the ink! side.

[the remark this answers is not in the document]
cmichi: That's a good point. There's nothing from keeping people to just use the type as a general purpose `[u8; 20]` without any association to an address. 

Hmm, maybe we should after all introduce an `Environment::Address` type instead of using `H160` implicitly for that. That would give you more clarity on the semantic interpretation.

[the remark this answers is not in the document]
cmichi: Do you by chance have an idea if `H160` is used for anything but an address? Or is it an established idiom that H160 == address?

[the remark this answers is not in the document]
cmichi: Thinking more about this: If a Solidity contract function takes an address as an argument, as what type does this show up in metadata? `address` I guess?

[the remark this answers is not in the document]
cmichi: Gotcha.

How is the situation for `U256` and `H256`, are there Solidity types for `balance` or `hash` used for those?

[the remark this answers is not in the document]
cmichi: Okay, but no more semantic equivalent (as for `address` instead of `bytes20`/`uint160`).

So we only need to add `Environment::Address`, but not an equivalent for `U256`/`H256`.

crates/build/src/solidity_metadata.rs
```diff
@@ -0,0 +1,292 @@
...
+    #[serde(rename = "keccak256")]
+    hash: Option<CodeHash>,
+    /// Version of the compiler.
+    version: String,
+}
+
+/// Generated information about the contract.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Output {
+    /// ABI definition of the contract.
```
cmichi, line 106: I've looked for a JSON schema of the Solidity metadata online, but couldn't find one. Are you aware of any formal verification spec?

crates/build/src/solidity_metadata/abi.rs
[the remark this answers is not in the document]
davidsemakula: The other one that makes sense is a `Byte` type so we can semantically differentiate between `bytes` and `uint8[]` (and similar). This way mapping become:

ink! <-> Solidity
`u8` == `uint8`
`Byte` == `byte1`
`[u8; 2]` == `uint8[2]`
`[Byte; 2]` == `bytes2`
...
`[u8; 32]` == `uint8[32]`
`[Byte; 32]` == `bytes32`
`Vec<u8>` == `uint8[]`
`Vec<Byte>` == `bytes`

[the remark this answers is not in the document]
cmichi: Do I understand you right that you mean in the sense of us providing a `type ink_primitives::Byte = u8` for developers? 

So if e.g. someone wants to ensure the following gets mapped to `bytes2` in the Solidity metadata:

```
#[ink(message)]
fn get(&self) -> [Byte; 2]
```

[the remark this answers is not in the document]
davidsemakula: Exactly

crates/build/src/solidity_metadata/abi.rs
davidsemakula: See https://github.com/use-ink/cargo-contract/pull/1930#discussion_r1944798396

crates/build/src/solidity_metadata/abi.rs
davidsemakula: See https://github.com/use-ink/cargo-contract/pull/1930#discussion_r1944798396

crates/build/src/solidity_metadata/abi.rs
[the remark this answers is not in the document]
peterwht: For the existing `AccountId = [u8;32]`, would that stay the same? And then would the mapping to Solidity be something like: `AccountId` == `uint8[32]`.

Or would it become: `AccountId = [Byte; 32]` == `bytes32`?

One other thought: if an RLP encoded ink! message takes an `AccountId`(32) as a parameter, should we give any linting warnings or simply represent it as `bytes32` (or similar)?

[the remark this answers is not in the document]
davidsemakula: > Or would it become: AccountId = [Byte; 32] == bytes32?

Yes `ink_primitives::types::AccountId` should be `bytes32`. There's no need to convert it to `AccountId = [Byte; 32]` though, the top level wrapper type already makes it unambigious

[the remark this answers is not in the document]
davidsemakula: > if an RLP encoded ink! message takes an AccountId(32) as a parameter, should we give any linting warnings or simply represent it as bytes32 (or similar)?

I think we just represent it as `bytes32` with no warnings since Solidity has no equivalent elementary/primitive type and the user/dev can't do anything about it.

crates/build/src/lib.rs
```diff
@@ -204,8 +207,12 @@ impl BuildResult {
         );
         if let Some(metadata_result) = self.metadata_result.as_ref() {
             let bundle = format!(
-                "  - {} (code + metadata)\n",
-                util::base_name(&metadata_result.dest_bundle).bold()
+                "  - {} ({})\n",
+                util::base_name(&metadata_result.dest_bundle).bold(),
+                match metadata_result.spec {
+                    MetadataSpec::Ink => "code + metadata",
+                    MetadataSpec::Solidity => "Solidity compatible contract metadata",
```
cmichi, line 214:
```suggestion
                    MetadataSpec::Solidity => "Solidity compatible metadata",
```

For a consistent naming throughout.

crates/build/src/lib.rs
```diff
@@ -218,8 +225,12 @@ impl BuildResult {
         }
         if let Some(metadata_result) = self.metadata_result.as_ref() {
             let metadata = format!(
-                "  - {} (the contract's metadata)",
-                util::base_name(&metadata_result.dest_metadata).bold()
+                "  - {} (the contract's {})",
+                util::base_name(&metadata_result.dest_metadata).bold(),
+                match metadata_result.spec {
+                    MetadataSpec::Ink => "metadata",
+                    MetadataSpec::Solidity => "Solidity compatible ABI",
```
cmichi, line 232:
```suggestion
                    MetadataSpec::Solidity => "Solidity compatible metadata",
```

For a consistent naming throughout.

crates/build/src/solidity_metadata.rs
```diff
@@ -0,0 +1,292 @@
...
+pub struct SolidityContractMetadata {
+    /// Details about the compiler.
+    compiler: Compiler,
+    /// Source code language
+    language: String,
+    /// Generated information about the contract.
+    pub output: Output,
+    /// Compiler settings.
+    /// Required by the spec, but very Solidity/EVM specific.
+    // TODO: (@davidsemakula) include ink! compiler settings instead?
```
cmichi, line 80: The `cargo-contract` version that was used to compile would be fitting here.

crates/build/src/solidity_metadata/abi.rs
```diff
@@ -0,0 +1,377 @@
...
+
+/// Generates a Solidity-compatible ABI for the ink! smart contract (if possible).
+///
+/// Ref: <https://docs.soliditylang.org/en/latest/abi-spec.html#abi-json>
+pub fn generate_abi(ink_project: &InkProject) -> Result<JsonAbi> {
+    let registry = ink_project.registry();
+    let spec = ink_project.spec();
+
+    // Solidity allows only one constructor, so we arbitrarily choose the first one.
+    // FIXME: (@davidsemakula) can we make constructor selection a bit more deterministic?
```
cmichi, line 61: ink! supports the annotation of a default constructor with the syntax

```rust
#[ink(constructor, default)]
pub fn my_constructor() -> Self {}
```

[Code pointer](https://github.com/use-ink/ink/blob/7338425630ca122717e360573d1b6ad09992d559/crates/ink/ir/src/ir/attrs.rs#L301-L305)

Can you take that into consideration here? So that if a default one exists we use that one. 

The behavior of how the constructor is chosen should also be documented in the comments.

crates/build/src/solidity_metadata/abi.rs
```diff
@@ -0,0 +1,377 @@
...
+    let ctor = spec.constructors().first().ok_or_else(|| {
+        anyhow::anyhow!("Expected at least one constructor for in contract metadata")
```
cmichi, line 63:
```suggestion
        anyhow::anyhow!("Expected at least one constructor in contract metadata")
```

crates/build/src/solidity_metadata/natspec.rs
```diff
@@ -0,0 +1,262 @@
...
+    methods: HashMap<String, ItemDevDoc>,
+    /// Events developer documentation, keys are canonical event signatures.
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    events: HashMap<String, ItemDevDoc>,
+    /// Errors developer documentation, keys are canonical error signatures.
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    errors: HashMap<String, ItemDevDoc>,
+}
+
+/// NatSpec user documentation of the contract.
```
cmichi, line 69: Can you explain in a comment somewhere how developers can influence which text lands in `NatSpec` and which in `DevDoc`?

crates/build/src/solidity_metadata/abi.rs
davidsemakula: Makes sense, didn't think about the default arg for some reason 🙂

crates/build/src/solidity_metadata.rs
davidsemakula: Will do. 
What do you think of using the key `settings.ink` add including the equivalent of [`source.build_info` from ink!'s metadata](https://use.ink/basics/metadata#source) for this? 
We're using the `ink` as a form of crude namespacing since this is technically not part of the Solidity metadata spec.

crates/build/src/solidity_metadata/natspec.rs
davidsemakula: So for Solidity, NatSpec (https://docs.soliditylang.org/en/latest/natspec-format.html) has a concept of "tags" (e.g. `@dev` and `@notice`) that represent the target audience (e.g. developer or user) for specific lines in docs e.g.
``` Solidity
/// @notice An NFT
/// @dev Inherits OpenZepplin ERC721 implementation
```

ink!/Rust don't really have a "native" equivalent to this, and I'm not even sure we need one because explorers seem to show both AFAICT. So I think we just treat ink!/Rust comments as `DevDoc` for now, and ignore (leave empty) the `UserDoc` fields.

crates/build/src/lib.rs
davidsemakula: I'm not sure about this one, this line is for the `<artifact_name>.abi` file. See image below:
Screenshot 2025-02-07 at 15 43 27

With this suggested change, this becomes
Screenshot 2025-02-07 at 15 52 36

crates/build/src/solidity_metadata.rs
davidsemakula: This is done, but also added the code hash (for the binary) and the docker image (if any). See implementation.

crates/build/src/solidity_metadata/abi.rs
davidsemakula: Done

crates/build/src/solidity_metadata/abi.rs
```diff
@@ -0,0 +1,391 @@
...
+        .map(|param| {
+            let param_name = param.label();
+            let ty_id = param.ty().ty().id;
+            let sol_ty = resolve_ty(
+                ty_id,
+                registry,
+                &format!("arg `{param_name}` for event `{name}`"),
+            );
+            // TODO: (@davidsemakula) should we simply omit events with Solidity ABI
+            // incompatible types instead of bailing?
```
davidsemakula, line 203: Does this make sense?

crates/build/src/solidity_metadata/abi.rs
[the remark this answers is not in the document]
davidsemakula: Should be resolved by https://github.com/use-ink/ink/pull/2397 and https://github.com/use-ink/cargo-contract/pull/1930/commits/f4c2ea00f2d03a0619ad209a18c4b78e188e7e6c

crates/build/src/solidity_metadata/abi.rs
cmichi: I would bail, as not being able to persist all events in the metadata breaks the user expectations. If need comes up we could add a flag to overwrite this behavior.

cmichi (approved):
LGTM! This is great, David.

Have you tried putting e.g. `flipper.abi` into a Solidity block explorer to see if it would be accepted?

cmichi:
I'll add the type `Address` to ink!, instead of the `H160`'s there. We can then create a follow-up to this PR here so that `address` is written into the metadata.

davidsemakula:
> I'll add the type `Address` to ink!, instead of the `H160`'s there. We can then create a follow-up to this PR here so that `address` is written into the Solidity metadata.

@cmichi See this PR https://github.com/use-ink/ink/pull/2397

crates/build/src/solidity_metadata/abi.rs
davidsemakula: Cool ... this is the current behavior.

davidsemakula:
> Have you tried putting e.g. `flipper.abi` into a Solidity block explorer to see if it would be accepted?

I've tried `flipper.abi` in [ethers.js](https://docs.ethers.org) and [web3.js](https://web3js.org/)

PR #2015, opened 2025-04-21
feat: remove `skip_clippy_and_linting` flag from `build` command + add `lint` command

chungquantin:
## Summary
Closes https://github.com/use-ink/cargo-contract/issues/1989

## Description

- Move the linting logic from `build/lib.rs` to a separate `lint.rs` file.
- Remove `skip_clippy_and_linting` flag from the `cargo contract build` command. Adjust the `lint` flag to requires the `--generate=check-only` if the flag is provided. 
- Add a new lint command to only run linting the contract.
- Add `lint_code_only` test.

```bash
➜ cargo contract lint --help
Lint a contract

Usage: cargo contract lint [OPTIONS]

Options:
      --manifest-path <MANIFEST_PATH>
          Path to the `Cargo.toml` of the contract to build

      --quiet
          No output printed to stdout

      --verbose
          Use verbose output

      --lint
          Performs extra linting checks for ink! specific issues during the build process.

          Basic clippy lints are deemed important and run anyway.

  -h, --help
          Print help (see a summary with '-h')
```


crates/build/src/lint.rs
```diff
@@ -0,0 +1,259 @@
...
+use colored::Colorize;
+use std::{
+    path::Path,
+    process::Command,
+};
+
+/// Configuration of the linting module.
+///
+/// Ensure it is kept up-to-date when updating `cargo-contract`.
+pub(crate) mod linting {
```
al3mart, line 36: It feel like this module is a bit redundant now within `lint.rs`. I don't know if the refactor is big enough to not make it worth it to change, though.

crates/cargo-contract/src/cmd/lint.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[clap(long, value_parser)]
+    manifest_path: Option<PathBuf>,
+    #[clap(flatten)]
+    verbosity: VerbosityFlags,
+    /// Performs extra linting checks for ink! specific issues during the
+    /// build process.
+    ///
+    /// Basic clippy lints are deemed important and run anyway.
+    #[clap(long)]
+    lint: bool,
```
al3mart, line 40: It seems that this is only used to run extra lints via `dylint` aside from `clippy` lints. Might be worth changing the name of the option to `extra-lints` or `dylint`.

al3mart (commented):
Did a quick review at it and it looks good :)

Have left a couple of comments. One regarding the change of the name for the option `--lint` within the `lint` command so it reads less repetitive and the name a bit more meaningful given this new context.

The other comment is a nit about a potential refactor of module `linting` which I am not sure might be worth the effort, haven't either look for usages of that module across the codebase.

crates/build/src/lint.rs
chungquantin: Great suggestion, I have refactored to remove the redundant linting module.

crates/cargo-contract/src/cmd/lint.rs
chungquantin: Great! I have renamed it to `dylint` and updated the comment above

CHANGELOG.md
```diff
@@ -119,20 +119,41 @@ of bugs in PolkaVM that prohibit us from using e.g. LTO. The contract sizes will
...
 ```rust
-let args = ExecuteArgs {
-  skip_clippy_and_linting: true
-  ...
-};
-let res = contract_build::execute(args);
+let crate_metadata = CrateMetadata::collect(manifest_path)?;
+let verbosity = TryFrom::<&VerbosityFlags>::try_from(&self.verbosity)?;
+
+contract_build::lint(extra_lint, &crate_metadata, &verbosity);
```
al3mart, line 153: `extra_lint` is used here. I don't know if that was done in purpose or it should be `dylint`.

al3mart (approved):
Looks good to me!

One consideration would be adding in some deprecation messages for the options we are removing. But I think we should be mostly OK if we decide not doing so.

CHANGELOG.md
chungquantin: Great catch, this was missing. I have updated the code to get rid of all `extra_lint` and replaced with `dylint`

crates/cargo-contract/src/cmd/lint.rs
davidsemakula: @chungquantin @al3mart `dylint` is an implementation detail, so `extra-lints` is a better user-facing name if we're renaming this.

chungquantin: @davidsemakula I see, that makes sense and more user friendly. I reverted back to extra-lints.

davidsemakula (commented):
Looks good! 
Except for one high-level comment 🙂

crates/cargo-contract/src/cmd/build.rs
```diff
@@ -57,11 +57,11 @@ pub struct BuildCommand {
...
     build_offline: bool,
-    /// Performs extra linting checks for ink! specific issues during the build process.
-    ///
-    /// Basic clippy lints are deemed important and run anyway.
+    /// Performs extra linting checks during the build process. Basic clippy
+    /// lints are deemed important and run anyway. This flag is only effective when using
+    /// with `--generate=check-only`.
     #[clap(long)]
-    lint: bool,
+    extra_lints: bool,
```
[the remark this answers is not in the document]
chungquantin: @davidsemakula Removed completely here https://github.com/use-ink/cargo-contract/pull/2015/commits/59593b4381dc85a0e280ad6707976bc764bf768b. At first I want to keep that because there is a `generate=check-only` that can be used with. But I guess it complicates things unnecessary. Removed the flag to keep the build command cleaner

davidsemakula (approved):
Thanks @chungquantin

PR #2017, opened 2025-04-23
refactor: contract_address to have account id in parameters

AlexD10S:
Refactored `contract_address` to take an `AccountId` instead of a `Signer`. This makes the function easier to reuse, especially in cases where we only have the user's address and not a full signer, is useful for example in `pop-cli` where we sign the deployment of the contract using a wallet UI.


AlexD10S:
Closing this as it not needed anymore

PR #2162, opened 2025-11-05
fix: event decoding

### LINKED ISSUE #2150: fix: decode event values
**Describe the bug**
For a contract event, cargo contract is not properly decoding the value

**To Reproduce**
Use a contract with events, or I just have added a simple event in Flipper contract:
```rs
#[ink(event)]
    pub struct Test {
        value: bool,
    }
```
And emit it once calling the `flip` message:
```rs
#[ink(message)]
        pub fn flip(&mut self) {
            self.value = !self.value;
            Self::env().emit_event(Test {
                value: self.value,
            });
        }
```

The output from cargo-contract is:
```
Event Revive ➜ ContractEmitted
         contract: 0x5801b439a678d9d3a68b8019da6a4abfa507de11
         data: 0x040004882e8db2d778ca9655ea3223def2503a484e55d21aaec3485fef87b8fbadbf29
         topics: 0x040004882e8db2d778ca9655ea3223def2503a484e55d21aaec3485fef87b8fbadbf29
```

The data should be decoded into the event.

Daanvdplas:
Closes #2150

### DIFF
--- a/crates/extrinsics/src/events.rs
+++ b/crates/extrinsics/src/events.rs
@@ -193,6 +193,7 @@ impl DisplayEvents {
             };
 
             let event_data = &mut event.field_bytes();
+            tracing::debug!("event data: {:?}", hex::encode(&event_data));
             let mut unnamed_field_name = 0;
             for field_metadata in event_fields {
                 if let Some(ref ce) = contract_emitted {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 [Unreleased]
 
+### Fixed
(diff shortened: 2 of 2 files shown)


crates/extrinsics/src/events.rs
```diff
@@ -181,25 +181,66 @@ impl DisplayEvents {
...
-            let event_sig_topic = event.topics().iter().next();
             let mut unnamed_field_name = 0;
             for field_metadata in event_fields {
-                if <ContractEmitted as StaticEvent>::is_event(
-                    event.pallet_name(),
-                    event.variant_name(),
-                ) && (field_metadata.name == Some("data".to_string())
-                    || field_metadata.name == Some("topics".to_string()))
-                {
-                    tracing::debug!("event data: {:?}", hex::encode(&event_data));
```
cmichi: Can you retain the
```
tracing::debug!("event data: {:?}", hex::encode(&event_data));
```
? Just put it right below `let event_data`.

PR #2167, opened 2025-11-14
Update installation instructions

Lohann:
## Description
- [x] Correctly map docker volume when the local directory contain spaces.
- [x] Use `CARGO_CONTRACT_VERSION=6.0.0-beta.1` when building locally.

The following command fails when the current path `$(pwd)` contains spaces, and believe there's a lot of devs who doesn't know how to fix it, so I've updated the instructions. 
```shell
docker run --rm -it -v $(pwd):/sources
```

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/build-image/README.md
+++ b/build-image/README.md
@@ -55,7 +55,7 @@ _Settings_ -> _Features in development_ tab in Docker Desktop App.
 Build docker image:
 
 ```bash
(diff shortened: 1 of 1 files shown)


davidsemakula (approved):
Thanks!

